Issue #99, opened 2017-11-13
Hide comment form on unvetted proposals


sndurkin:
Admins can't comment on unvetted proposals anyway, the API is returning a "proposal not found" error.

PR #105, opened 2017-11-16
Hide comment form on unvetted proposals

tiagoalvesdulce:
Fixes https://github.com/decred/politeiagui/issues/99


src/components/snew/CommentArea.js
```diff
@@ -2,14 +2,14 @@ import React from "react";
...
-const CommentArea = ({ comments, loggedIn, ...props }) => (
-  <CommentAreaBase {...{
-    ...props,
-    num_comments: comments.length,
-    locked: !loggedIn,
-    currentSort: "new",
-    sortOptions: ["new"]
-  }} />
+const CommentArea = ({ comments, loggedIn, proposal, ...props }) => (
+  !loggedIn || proposal.status === 2 || proposal.status === 3 ? null :
```
go1dfish: Use the constants from https://github.com/decred/politeiagui/blob/master/src/constants.js

Otherwise this looks good.

go1dfish: Actually one more thing, we don’t want to hide the comment area entirely if a user is logged out, switch that backed to using the locked prop on CommentAreaBase and only return null for unreviewed or censored proposals.

mocks/api/v1/proposals/6284c5f8fba5665373b8e6651ebc8747b289fed242d2f880f64a284496bb4ca9/GET.json
```diff
@@ -0,0 +1,22 @@
+{
+  "proposal": {
+    "name": "This is an example proposal",
```
go1dfish: Let’s also make it clear in the name that this is an unreviewed proposal example.

Issue #103, opened 2017-11-15
Proposal submission success page


go1dfish:
https://github.com/decred/politeiagui/issues/81

Currently pgui redirects user to their proposal’s permalink after successful submission.

API no longer returns a users unvetted proposals so we will need to add a page that replicates proposal detail with the data from the successfully submitted proposal (should be in the proposal request’s payload and response in state)

tiagoalvesdulce:
I am working on that.

PR #110, opened 2017-11-19
Create new proposal success page

tiagoalvesdulce:
Resolves https://github.com/decred/politeiagui/issues/103


src/actions/api.js
```diff
@@ -221,12 +223,26 @@ export const onFetchProposalComments = (token) =>
...
+      .then(response => {
+        const proposal = {
+          censorshiprecord: { token: response.token },
+          files: response.files,
+          name,
+          description,
+          timestamp: Date.now() / 1000,
+          status: PROPOSAL_STATUS_UNREVIEWED
+        };
+        console.log("proposal: ", proposal);
```
go1dfish: You left a log here

src/selectors/api.js
```diff
@@ -110,7 +110,10 @@ export const unreviewedProposals = filtered(PROPOSAL_STATUS_UNREVIEWED);
 export const censoredProposals = filtered(PROPOSAL_STATUS_CENSORED);
 export const unvettedProposalsIsRequesting = or(isApiRequestingInit, isApiRequestingUnvetted);
 export const unvettedProposalsError = or(apiInitError, apiUnvettedError);
-export const proposal = or(compose(get("proposal"), apiProposalResponse), constant({}));
+export const proposal = or(compose(get("proposal"), apiProposalResponse), (state) => {
```
go1dfish: I don't like like having access to the app state in the api selectors.

We should think of the api selectors as lower level and do more complex logic in the app selectors.

Let's revert proposal as it was, and do an or in selectors/app that incorporates it and the new logic.

It can be exported as proposal and no other code will change

tiagoalvesdulce:
In order to move submittedProposals selector logic to the app file, I had to move the "files" selectors as well. I really think they fit better on the app file now with new proposal selector.

Issue #100, opened 2017-11-13
Need access to policy inside getHumanReadableError helper


sndurkin:
We need access to the policy object fetched from `/api/v1/policy` within `getHumanReadableError()` inside `src/helpers.js`.

fernandoabolafio:
Im working on it

fernandoabolafio:
@sndurkin  As we discussed on slack I guess we can close this issue. The policy won't be added within the `getHumandReadableError()`. We can provide better error messages doing policy validations whereas it is need. Take a look on my PR:
https://github.com/decred/politeiagui/pull/113

PR #113, opened 2017-11-21
Better Proposal Files Policy Validation 

fernandoabolafio:
I've added a validation for the files whenever a user try to add new files. The files which doesn't fit the policy validation won't be added to the input's files and error messages will be displayed to the user. 
Right now there are two main policy validations being processed: 

- Max number of file: make sure to respect the max number of images allowed by the policy and provide an global error message. 
- Max size of each file: make sure that every file added to the input's value will respect the max size allowed by the policy and provide an error message per file.
- Mime types is validated by the component 'react-file-reader' which won't allow you to select mime type that aren't specified in the prop `fileTypes`.

The FE result will be like this: 
screen shot 2017-11-21 at 3 49 59 pm

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
     "react": "^16.0.0",
     "react-body": "^0.2.0",
     "react-dom": "^16.0.0",
-    "react-file-reader": "^1.1.2",
+    "react-file-reader": "^1.1.3",
     "react-markmirror": "^0.4.16",
     "react-redux": "^5.0.6",
     "react-router-dom": "^4.2.2",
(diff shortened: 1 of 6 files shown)


fernandoabolafio:
Fixes #100

fernandoabolafio:
Sure!

fernandoabolafio:
updated:
screen shot 2017-11-21 at 4 13 46 pm

go1dfish:
Code looks good and really appreciate the screenshots 👍 

Should be able to test later today.

fernandoabolafio:
Fixed. Should be fine now.

PR #130, opened 2017-12-07
Fix remaining issues with loading icons

### LINKED ISSUE #127: Need <PageLoadingIcon> component
Currently we're using `<LoadingIcon>` for page-to-page transitions, but that's not working well on the proposal detail page. We should create a `<PageLoadingIcon>` component which is used for page-to-page transitions, and is consistently displayed maybe 200 - 300px from the top of the page. This should be separate from `<LoadingIcon>` which should be easily embeddable into other components. Similarly, `<ButtonWithLoadingIcon>` should be refactored to use `<LoadingIcon>`.

sndurkin:
Fixes #127.


tiagoalvesdulce:
Looks really good!

PR #140, opened 2017-12-08
Replace markdown parser snuownd by showdown

### LINKED ISSUE #137: Inconsistency between the markdown editor and the preview
On the proposal submit page, if you select a multiline block of code in the markdown editor and then click the code button, it will wrap the highlighted text with triple backticks, but they aren't displayed correctly in the markdown preview component:

image

The markdown preview component does correctly display code indented by 4 spaces, though.

fernandoabolafio:
I've replaced the markdown parser for both: Proposal details and Proposal Submit page. 
I think `showdown` parser (which is the same used in `react-mde`) works better than `snuownd`.
Besides it is more consistent with our current editor.
fixes #137

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,9 +26,8 @@
     "redux-form": "^7.1.1",
     "redux-logger": "^3.0.6",
     "redux-thunk": "^2.2.0",
-    "showdown": "^1.8.3",
+    "showdown": "^1.8.4",
     "snew-classic-ui": "https://github.com/decred/snew-classic-ui",
-    "snuownd": "https://github.com/JordanMilne/snuownd",
     "timeago-react": "^1.2.2",
     "tweetnacl": "^1.0.0",
     "tweetnacl-util": "^0.15.0"
--- a/src/components/MarkdownEditor/MarkdownPreview.js
+++ b/src/components/MarkdownEditor/MarkdownPreview.js
@@ -1,11 +1,11 @@
-import React from 'react';
-import PropTypes from 'prop-types';
-import MarkdownHelp from './MarkdownHelp';
-import Markdown from '../snew/Markdown';
+import React from "react";
+import PropTypes from "prop-types";
+import MarkdownHelp from "./MarkdownHelp";
+import Markdown from "../snew/Markdown";
 
-const MarkdownPreview = ({body}) => (
+const MarkdownPreview = ({body, showdownOptions}) => (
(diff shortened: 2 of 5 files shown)


fernandoabolafio:
fixes #137

go1dfish:
My thinking with snudownd is it is based on reddit's markdown parser which does not allow hot linking of any external content like images.

For privacy and security reasons I want to avoid any markdown parser that supports arbitrary html markup without escaping, and I don't want our markdown parser to support embedding images via urls as this can potentially compromise privacy of users.

I'm not stuck on snudownd for the long term.  My thinking was it would be easiest to achieve the image embeds we need while satisfying our security requirements by using something like:

https://github.com/rexxars/react-markdown

This would also allow us to do more customization of our markdown rendering as necessary including doing things like rendering links to youtube videos as embeds.

Again I'm not stuck on snudown what I am stuck on are:

- No arbitrary markup support (or we need to run through something to pull out html ourselves)
- No external image/media reference support without us being explicit about allowing it

But this depends on the security model we want to take.  @marcopeereboom may want to get in on this thread as well.

If we want to go with a looser markdown parser and potentially allow external references, then we need to disable markdown rendering by default for admins reviewing proposals.

I generally think we're better to stick with more restrictive markdown renderer and be very explicit about any externally allowed media.  Things like youtube etc... that are widely used might be allowable, but we shouldn't support arbitrary external references IMO.

fernandoabolafio:
I agree with @go1dfish concerns about security issues on markdown rendering. I read this [post](https://github.com/showdownjs/showdown/wiki/Markdown's-XSS-Vulnerability-(and-how-to-mitigate-it)) from the creator of the showdown parser (the one I've inserted on this PR). So unless we use some sort of [XSS filtering](https://github.com/yahoo/xss-filters) on server side I believe we should stick with the old one or migrate to `react-markdown`as @go1dfish have suggested. 
I tested `react-markdown` and it handles well all XSS I tested (test cases from the post I've mentioned above).

go1dfish:
Very good article.

Make a new PR that switches us to react-markdown and I’ll try it out.

XSS is an important concern but not the only one.  We want to prevent external image embedding and try to prevent browsers from auto following any embedded links for privacy concerns.

fernandoabolafio:
Ok, I will do another PR for `react-markdown`. Closing this one.

Issue #143, opened 2017-12-11
Incorrect state reached when creating and approving proposal


sndurkin:
When creating and approving a proposal with the same user without refreshing the browser, the client enters an incorrect state where it does not properly update the proposal status on the proposal detail page. It shows the proposal in the vetted list, but when you open the proposal details, it shows the Approve/Spam buttons when it should hide those and show the comment form.

PR #144, opened 2017-12-11
Fix new proposal status change

tiagoalvesdulce:
Fixes https://github.com/decred/politeiagui/issues/143


dajohi:
Fails to build with node 6, which is ok since production runs 8.  See the dockerfile.
If 6 is no longer supported, please remove the line from .travis.yaml.

PR #145, opened 2017-12-12
improve markdown 

### LINKED ISSUE #137: Inconsistency between the markdown editor and the preview

fernandoabolafio:
Now I've switched to react-markdown as we discussed on #140. 
This markdown cover lots of security potencial issues but we need still to figure out which approach to take for embedded images/links. 
@go1dfish looks like this can make our life easier handling links: 
https://github.com/rexxars/react-markdown/issues/124
Are we going to block any sort of external links?

 fixes #137

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,6 @@
     "redux-thunk": "^2.2.0",
     "showdown": "^1.8.3",
     "snew-classic-ui": "https://github.com/decred/snew-classic-ui",
-    "snuownd": "https://github.com/JordanMilne/snuownd",
     "timeago-react": "^1.2.2",
     "tweetnacl": "^1.0.0",
     "tweetnacl-util": "^0.15.0"
--- a/package.json
+++ b/package.json
@@ -18,6 +18,7 @@
     "react-body": "^0.2.0",
     "react-dom": "^16.0.0",
     "react-file-reader": "^1.1.3",
+    "react-markdown": "^3.1.1",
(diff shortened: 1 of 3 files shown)


fernandoabolafio:
One thing that might work is to trigger some alert like this when a user clicks some link:
```
 ______________________________
|          Alert               |
|      Follow this link        |
|      at your own risk        |
|      -> link                 |
|______________________________|
```

dajohi:
This fails to build on node 8, which is required.

go1dfish:
I don't think we need to block all external links, but we should mark them as nofollow.

All external embedded images need to be blocked or just rendered as a text link to the image.

We have no solution for handling attached image embeds yet, but I expect what we will do is use a filename reference in the image markdown and render it with an img tag using the base64 file data.  

I wouldn't worry about attached image embeds yet as the current renderer doesn't handle these either.

fernandoabolafio:
`react-markdown`actually can render embedded images, see:
screen shot 2017-12-12 at 2 33 55 pm
But they also provide an api method for transforming images, from their docs:

- transformImageUri - function|null Function that gets called for each encountered image with a single argument - uri. The returned value is used in place of the original.

PR #154, opened 2017-12-14
Markdown 'on the fly' help

### LINKED ISSUE #150: Users should be informed that 2 newlines are needed to create new paragraphs
A user on Slack reported that newlines on comments are broken; they weren't aware you need 2 newline characters to create a new paragraph.

I know this mistake is made fairly often with new users on Reddit, and I think the same will happen on the proposal system as a lot of people won't be familiar with Markdown formatting. I think the UI can be a little smarter here and inform users if they're using single newline characters in their proposal or comment that it will not show up as separate paragraphs.

fernandoabolafio:
I've introduced the component `MarkdownLiveHelper` it basically allows you to specify "mistakefinders" which take as input the current editor value and return if there is a potencial mistake and a helper message to be displayed for the user. We can use many of them but for now we have only detection for single newlines between chars.
In the overall it looks like this:
screen shot 2017-12-14 at 12 23 40 pm
fixes #150

### DIFF
--- a/src/components/MarkdownEditor/helpers.js
+++ b/src/components/MarkdownEditor/helpers.js
@@ -1,0 +1,68 @@
+const CHAR_CODES = {
+    newline: 10
+};
+
+const finderResult = (mistake, message = '') => ({mistake, message});
+
+function iterateOverStringCharsCode(str, callback) {
+    for (let i = 0, len = str.length; i < len; i++) {
+        callback(str.charCodeAt(i), i);
+    }
+};
+
+function tryingParagraphWithOneNewline(value) {
+    const { newline } = CHAR_CODES;
+    let thereAreASingleNewline = false;
+    iterateOverStringCharsCode(value, (charCode, index) => {
+        const nextCharCode = value.charCodeAt(index + 1);
+        const prevCharCode = value.charCodeAt(index - 1);
+        if (charCode === newline && nextCharCode && nextCharCode !== newline && prevCharCode !== newline ) {
+            thereAreASingleNewline = true;    
+        }
+    })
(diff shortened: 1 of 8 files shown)


src/components/MarkdownEditor/helpers.js
```diff
@@ -0,0 +1,37 @@
...
+        const nextCharCode = value.charCodeAt(index + 1);
+        const prevCharCode = value.charCodeAt(index - 1);
+        if (charCode === newline && nextCharCode && nextCharCode !== newline && prevCharCode !== newline ) {
+            thereAreASingleNewline = true;    
+        }
+    })
+    return finderResult(thereAreASingleNewline, '2 newlines are required to create a paragraph');
+};
+
+const mistakeFInders = [
```
sndurkin: Minor typo, should be `mistakeFinders`

tiagoalvesdulce:
It also fails for Headings and Lists:

screen shot 2017-12-14 at 13 18 40
screen shot 2017-12-14 at 13 19 33

src/components/snew/Markdown/helpers.js
```diff
@@ -0,0 +1,62 @@
...
+    const size = currentString.length;
+
+    if(currentString.charCodeAt(0) === 10) {
+        childrens.push(React.createElement('br'));
+    }
+
+    chunks.forEach((chunk, idx) => {
+        childrens.push(chunk);
+        if((idx === chunks.length - 1 && currentString.charCodeAt(size - 1) === 10) || idx !== chunks.length - 1  ) {
+            childrens.push(React.createElement('br'));
```
tiagoalvesdulce: Pass a unique key prop here to clear the console warning.

Issue #155, opened 2017-12-15
Move the text preview next to the editor on proposal submit page


sndurkin:
It would be nice to have the preview area side-by-side to the editor on the proposal submit page, and maybe (if it makes sense) have the editor grow in height with the preview area.

Issue #147, opened 2017-12-12
Test failing sporadically


fernandoabolafio:
The test below is failing sporadically when running continues integration from travis:
```
✖ resetPassword

   - Reset password (5.019s)
   Timed out while waiting for element <.message-ct.message-error> to be present for 3000 milliseconds.  - expected "visible" but got: "not found"
```
Need to investigate what is causing it to occur to decide for rewriting the test or removing it.

vctt94:
I believe that because all tests are starting at the same time, even with a bigger time to get visible, they are failing because is using a lot of processing. We could use a queue or something similar so tests run each at a time.

fernandoabolafio:
Yes, is definetly a processing overload issue. When I ran the test on a single session at my computer it works perfectly. If I try to run the tests into 2 or more sessions simultaneosly some of the test cases starts to fail. So this is probably due the capability of the instance being used by Travis.

PR #157, opened 2017-12-16
[WIP] Move text preview next to the editor on proposal submit page

tiagoalvesdulce:
Closes https://github.com/decred/politeiagui/issues/155

### DIFF
--- a/src/components/MarkdownEditor/MarkdownEditor.js
+++ b/src/components/MarkdownEditor/MarkdownEditor.js
@@ -32,14 +32,23 @@ class MarkdownEditor extends React.Component {
   }
 
   handleValueChange = (value) => {
-      const {onChange} = this.props;
-      onChange(value.text);
+    this.textArea.style.height = "auto";
+    this.textArea.style.height = (this.textArea.scrollHeight) + "px";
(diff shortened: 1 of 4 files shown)


tiagoalvesdulce:
Tests failed becasue of https://github.com/decred/politeiagui/issues/147

tiagoalvesdulce:
It can be done @sndurkin. But if the user uses lots of 2 newlines to create paragraphs, the `textarea` will increase its height way faster than the preview and the user might scroll up to see the preview of what he is currently typing and it'd be a bad UX too.

sndurkin:
I agree, but I'm not sure there's much we can do about that particular issue, because it's not a WYSIWYG editor; the preview text will always be larger than the editor text for any headers, which will cause the text to get out of sync as well. Maybe it's better to eliminate the scrollbar for the entire page and just have independent scrollbars for the editor and preview?

tiagoalvesdulce:
I don't think all these scrolls look good. Also, our screen doesn't fit on only one view height so we need the page scrollbar. I tested both approaches and the `textarea` resize works better in my opinion.

Issue #159, opened 2017-12-16
Loading indicator causes too much changes on the screen


xaur:
As I browse through proposals, my screen flashes back and forth between content and huge rotating loading indicator. I think it would be better to replace rotating logo with covering the dynamic area with semi transparent white cover.

tiagoalvesdulce:
Good catch @xaur. I'll improve our loading indicator to prevent this flashing between it and the content.

xaur:
Besides transparent white blanked, another common non-intrusive approach is thin (1-2px) line on top, like here on GitHub.

PR #160, opened 2017-12-18
Wait 1 second to show loading icon on page transitions

tiagoalvesdulce:
As pointed out in https://github.com/decred/politeiagui/issues/159, the loading icon is causing some flickering when the page doesn't take long to load. As a solution, I added a 1 sec delay to show the loading icon on page transitions. The other approaches would require drastic UI changes and would be more appropriate after a redesign from the design team.

Fixes https://github.com/decred/politeiagui/issues/159

### DIFF
--- a/src/components/snew/PageLoadingIcon.js
+++ b/src/components/snew/PageLoadingIcon.js
@@ -1,11 +1,36 @@
 import React from "react";
 import LoadingIcon from "./LoadingIcon";
 
-const PageLoadingIcon = ({ hidden }) => (
-  <LoadingIcon
-    hidden={hidden}
-    width={200}
-    style={{ margin: "300px auto 0 auto" }} />
-);
-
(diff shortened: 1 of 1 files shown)


src/components/snew/PageLoadingIcon.js
```diff
@@ -1,11 +1,36 @@
...
+  }
+  componentWillUnmount() {
+    if (this.loadingIconDelay) {
+      clearTimeout(this.loadingIconDelay);
+      this.loadingIconDelay = 0;
+    }
+  }
+  render () {
+    const { show } = this.state;
+    console.log(show);
```
sndurkin: Could you remove this log?

tiagoalvesdulce: Oh, sorry, forgot about it.

PR #161, opened 2017-12-19
Provide some extra information for email verification messages.

### LINKED ISSUE #151: Email messages should provide a little more information
I think the "Please check your inbox" message could be a little more informative. Because we don't check to see if the email address is actually registered in the system, users with multiple email addresses may use the wrong one to recover an account and not realize it doesn't exist. We should probably display the email address they entered in bold text (to help users realize typos after-the-fact) and also clarify that if they don't receive an email, the account may not exist in the system and we won't disclose that fact for privacy reasons.

sndurkin:
Fixes #151


fernandoabolafio:
Tested.  Code looks good, messages are way better.

Issue #164, opened 2017-12-21
Create a faucet button for paywall


vctt94:
When on testnet, the UI should provide a "Pay with testnet faucet" button to allow users who complete new user registration to easily pay. It should POST to the form on the Decred Testnet Faucet page with the paywall address and the amount set to 0.1 DCR.

On success, it should probably set some variable in local storage that indicates the user has already paid but needs to wait for confirmation, and navigate to some page that explains the transaction must reach 2 confirmations before it allows submission.

Link for faucet: https://faucet.decred.org/

PR #165, opened 2017-12-27
Paywall

### LINKED ISSUE #106: add paywall component
Need to port the bootstrap+jquery widget from dcrpayments.

vctt94:
fixes #106 

Still need some styling and a faucet button for automatic testing, as mentioned at #164 

Right now is needed to manually change the paywall address for testing.


dajohi:
@vctt94 fix the test failures?

vctt94:
Tests are failing because of https://github.com/decred/politeiagui/issues/147

src/selectors/api.js
```diff
@@ -138,3 +138,8 @@ export const setStatusProposalIsRequesting = isApiRequestingSetStatusProposal;
 export const setStatusProposalToken = compose(get("token"), apiSetStatusProposalPayload);
 export const setStatusProposalError = apiSetStatusProposalError;
 export const redirectedFrom = get(["api", "login", "redirectedFrom"]);
+export const paywallAddress = compose(get("paywalladdress"), apiNewUserResponse);
+export const paywallAmount = compose(get("paywallamount"), apiNewUserResponse);
+export const VerificationToken = compose(get("verificationToken"), apiNewUserResponse);
```
sndurkin: The variable name here should use camelCase, and the string that's sent from Politeia will be all lowercase (same issue with the 2 lines below this).

src/connectors/paywall.js
```diff
@@ -0,0 +1,17 @@
+import { connect } from "react-redux";
+import { bindActionCreators } from "redux";
+import * as sel from "../selectors";
+import * as act from "../actions";
+
+export default connect(
+  sel.selectorMap({
+    paywallAddress: sel.paywallAddress,
+    paywallAmount: sel.paywallAmount,
+    VerificationToken: sel.VerificationToken,
```
sndurkin: We should use camelCase here.

vctt94:
Thanks for reviewing my PR.

I believe we can use in-line instead of a modal, should I remove the modal from the code? The way it is now, I believe it can be reusable for some modal we may have someday. 

I can change the polling for another place where it is always called, like the navbar. But I agree with you, it is better to wait for the server to inform it is paid.

But are we going to wait for the server to say it is paid and check for the mined blocks on the gui? Or will it be all on the server side and the gui only asks if it good to access the submit proposal area?

sndurkin:
> But are we going to wait for the server to say it is paid and check for the mined blocks on the gui? Or will it be all on the server side and the gui only asks if it good to access the submit proposal area?

I think the server will reply with `hasPaid` on `/user/login` and `/user/me`, and if it's false, the GUI will poll for mined blocks, because it's possible that the user verifies his email, logs in, and then the transaction is confirmed on the blockchain he clicks around on the site. In that case, we will want to enable commenting and proposal submissions without the user having to press F5 to refresh the browser. Does that make sense?

vctt94:
I removed the modal and made it a normal component, I created the hadPaid, which is get from login and me, and added the poll on the header if it hasPaid === false. But we have a problem, I may not have the paywall address and I am not sure if it makes sense to send it together with `/user/login` and `/user/me` route. What do you think?

mocks/api/v1/user/me/GET.json
```diff
@@ -1,4 +1,5 @@
 {
-  "email": "foo@foo.com",
-  "isadmin": true
-}
+    "hasPaid": true,
```
sndurkin: This should be all lowercase.

src/lib/external_api.js
```diff
@@ -0,0 +1,14 @@
+const dcrdataURL = (network, address) => `https://${network}.dcrdata.org/api/address/${address}/raw`;
+const getRawAddresses = (network, address) => fetch(dcrdataURL(network, address)).then(r => {
+  // work around when transactions are not paid and dcrdata api returns Unprocessable Entity
+  if(r.statusText === "Unprocessable Entity") {
+    return null;
+  }
+  return r.json();
+});
+
+export const getPaymentsByAddress = address => {
```
jolan: Makes more sense to just check the first letter of the payment address. i.e. D = mainnet, T = testnet

src/components/Paywall/Page.js
```diff
@@ -0,0 +1,27 @@
...
+  <div className="modal-wrapper">
+    <div className="paywall-header">
+      <h2 className="paywall-header-title">Paywall</h2>
+    </div>
+    <div className="paywall-content">
+      <p>Please send exaclty {props.paywallAmount}DCR to <br/>{props.paywallAddress}</p>
+    </div>
+    <div>
+      Status: {
+        grantAccess ?  <p>Approved</p> : <p>Waiting for blocks to get mined</p>
```
sndurkin: Let's call this "Payment received" and "Waiting for payment"

src/components/Paywall/Page.js
```diff
@@ -0,0 +1,27 @@
...
+  onSubmit,
+  grantAccess,
+  ...props
+}) => (
+  <div className="modal-wrapper">
+    <div className="paywall-header">
+      <h2 className="paywall-header-title">Paywall</h2>
+    </div>
+    <div className="paywall-content">
+      <p>Please send exaclty {props.paywallAmount}DCR to <br/>{props.paywallAddress}</p>
```
sndurkin: Typo, should be "exactly".

src/components/Paywall/Page.js
```diff
@@ -0,0 +1,27 @@
...
+    </div>
+    <div>
+      Status: {
+        grantAccess ?  <p>Approved</p> : <p>Waiting for blocks to get mined</p>
+      }
+    </div>
+    <div className="paywall-footer">
+      <a className="btn">Pay!</a>
```
sndurkin: I don't think a button is necessary here, because the user will pay with Decrediton, right?

src/components/Paywall/index.js
```diff
@@ -0,0 +1,54 @@
...
+import PaywallPage from "./Page";
+import paywallConnector from "../../connectors/paywall";
+
+class Paywall extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = this.getInitialState();
+  }
+
+  getInitialState() {
```
tiagoalvesdulce: getInitialState should not be used with ES6 classes, you can initialize the state in the constructor.

src/components/Paywall/index.js
```diff
@@ -0,0 +1,54 @@
...
+        }}
+      />
+    );
+  }
+
+  resetState() {
+    this.setState(this.getInitialState());
+  }
+
+  onSubmit() {
```
tiagoalvesdulce: onSubmit function is empty here.

vctt94:
I fixed your comments and also I added a warningPaywallNotPaid component for not showing the submit proposal button and save button on comments.

src/components/snew/CommentForm.js
```diff
@@ -77,18 +79,23 @@ const CommentForm = ({
...
-                className="cancel"
-                onClick={() => onSetReplyParent()}
-                type="button"
-              >
-              cancel
-              </button>)) || null}
-            </div>
+            {
+              !grantAccess ? <WarningPaywallNotPaid
+                message="You can not Comment on proposals before paying the paywall" /> : (
```
sndurkin: 2 comments:

* Could you include the Save button but just disable it and the text area? I think it will make it a little clearer that this is the comment form if the button is present too.
* I'd like to refrain from calling it a paywall, can you rephrase it to: "You must pay the registration fee before you can comment on proposals"

src/components/snew/SubmitTextSidebox.js
```diff
@@ -1,26 +1,30 @@
...
+            Submit a new Proposal
+          </Link>
+          <div className="nub" />
+        </div>
       </div>
     </div>
-  </div>
-) : null;
+  ) : <WarningPaywallNotPaid
+    message="You can not send Proposals before paying The Paywall" />
```
sndurkin: Same comments here:

* Could we show the "Submit a new proposal" button but disable it?
* Can you rephrase the message to: "You must pay the registration fee before you can submit a proposal."

src/actions/external_api.js
```diff
@@ -0,0 +1,68 @@
...
+  return txid;
+};
+
+const checkTransaction = (transaction, addressToMatch, amount) => {
+  let addressSeen = false;
+  let addressValue = 0;
+  for (let voutData of transaction["vout"]) {
+    const addresses = voutData["scriptPubKey"]["addresses"];
+    if (addresses) {
+      addresses.forEach(address => {
```
tiagoalvesdulce: This line is throwing a [no-loop-func](https://eslint.org/docs/rules/no-loop-func) warning. You should basically create a function outside the `for (let voutData ... )` loop and call it inside the `addresses.forEach`.

vctt94:
Made the sean changes and I changed forEach, because it needs a callback function.

mocks/api/v1/user/me/GET.json
```diff
@@ -1,5 +1,4 @@
 {
-    "email": "foo@foo.com",
```
tiagoalvesdulce: Removing this messes up with `yarn uidev` and Travis

vctt94:
Ok, I am redirecting now and I returned the email to the mock api.

I believe we should not have tests depending on the mock API. If the test need to be logged in to pass, it needs to login. Maybe we should have an issue for that.

Issue #82, opened 2017-10-30
receive and verify bundle after proposal submission


behindtext:
the censorship token amounts to a signature from a particular pi instance confirming that a proposal submission has been received. instead of having users lookup their proposal by id or censorship token to check its status, the user should be prompted to download a "proposal bundle", which would contain all the censorship data, along with the file(s) that make up the proposal.

beyond prompting the user to download a proposal bundle, there needs to be a simple tool that users can run to verify a censorship token is valid. this is important not only for the user being censored, but also other users who wish to verify the proposal in question was indeed censored.

sndurkin:
> instead of having users lookup their proposal by id or censorship token to check its status

Users will still need to access Politeia to check on a proposal's status.


> the user should be prompted to download a "proposal bundle"

@go1dfish, is this something the UI can do without any server involvement, or will the bundle have to be hosted by the server for a period of time?

tiagoalvesdulce:
Working on it

PR #167, opened 2017-12-29
Download Proposal Bundle

tiagoalvesdulce:
Closes https://github.com/decred/politeiagui/issues/82


src/selectors/api.js
```diff
@@ -105,6 +105,8 @@ export const isAdmin = bool(or(
   compose(get("isadmin"), apiMeResponse),
   compose(get("isadmin"), apiLoginResponse)
 ));
+
+export const serverIdentity = state => state.api.init.response && state.api.init.response.identity;
```
sndurkin: The API docs are apparently wrong, it's not called `identity` but rather `pubkey`: https://github.com/decred/politeia/blob/master/politeiawww/api/v1/v1.go#L249

I think the code was changed and the docs weren't updated, I'll submit a PR to fix that.

PR #173, opened 2018-01-08
Manage keys

### LINKED ISSUE #168: User needs a way to generate a new keypair
We need a new page or a section on the user profile page that allows the user to generate a new keypair and submit it to the server, to accommodate the case where the user loses his private key. The server will deactivate his old keypair and activate the new one.

This issue depends on decred/politeia#168.

fernandoabolafio:
Fixes #168 
The key can be update on `localhost:3000/user/profile/` clicking on 'Update key'
It will generate a verification token that can be grabbed from the logs.
With the verificationtoken in hand go to
`localhost:3000/user/key/verify/?verificationtoken=<verififcationtoken>`


src/components/VerifyKey/index.js
```diff
@@ -0,0 +1,51 @@
...
+import PageLoadingIcon from "../snew/PageLoadingIcon";
+import verifyKeyConnector from "../../connectors/verifyKey";
+import qs from "query-string";
+import Message from "../Message";
+
+class VerifyKey extends Component {
+  componentWillMount() {
+    if (isEmpty(this.props.location.search)) {
+      this.props.history.push("/user/login");
+      return;
```
brunobraga95: you dont really need this return, not that it makes a different but it is good to avoid verbose.

src/Routes.js
```diff
@@ -41,6 +42,7 @@ class Routes extends Component {
         <Route path="/user/verify/success" component={VerifySuccess} />
         <Route path="/user/verify/failure" component={VerifyFailure} />
         <Route path="/user/verify" component={Verify} exact={true} />
```
brunobraga95: you dont need to explicit the value `true`,  `<Route path="/user/verify" component={Verify} exact />` should be enough

src/Routes.js
```diff
@@ -41,6 +42,7 @@ class Routes extends Component {
...
+        <Route path="/user/key/verify" component={VerifyKey} exact={true} />
```
brunobraga95: you dont need to explicit the value `true`,  `<Route path="/user/verify" component={Verify} exact />` should be enough

src/components/VerifyKey/index.js
```diff
@@ -0,0 +1,51 @@
+import React, { Component } from "react";
```
brunobraga95: could import `prop-types` and use it for this component? I know we are not going to refactor the whole project right now, and it makes sense. But it is good to start using it at least on new components

src/components/UserProfile.js
```diff
@@ -2,14 +2,40 @@ import React from "react";
 import connector from "../connectors/userProfile";
 import PasswordChange from "./PasswordChange";
 import PrivateKeyIdentityManager from "./PrivateKeyIdentityManager";
+import Message from './Message';
+
+const UpdatedKeyMessage = ({email}) => (
+  <span>
+    Successfull updated your key! Please check your email at
```
sndurkin: Minor typo, should be "Successfully"

fernandoabolafio:
@brunobraga95  @sndurkin  I've updated the code and rebased, should be go to go. Thanks for reviewing.

sndurkin:
Thanks for re-reviewing Bruno.

Issue #163, opened 2017-12-21
Landing Page 


fernandoabolafio:
See https://github.com/decred/politeia/issues/59 and #106

fernandoabolafio:
Depends on https://github.com/decred/politeia/issues/194

tiagoalvesdulce:
Workin on it based on https://github.com/decred/politeia/pull/200

PR #177, opened 2018-01-15
Landing Page with user proposals

tiagoalvesdulce:
Created a new route called `user/:userid/proposals` that will be the new landing page where the user is redirected to after the login. I am not sure if we want the `userid` on the URL, so LMK your thoughts on this.

Closes https://github.com/decred/politeiagui/issues/163.


sndurkin:
I was thinking that the proposal list could be on the main user page, so the url would be `/user/:userid`, but I'm not sure what we would do with the Reset Password or the identity management area.

@go1dfish what do you think?

tiagoalvesdulce:
Yes, I think the headers are a good idea. Also, I'll be adding a Dropdown (https://github.com/decred/politeiagui/issues/193) on my next PR that will add a way to easily access `user/proposals` from the UI.

src/connectors/proposals.js
```diff
@@ -9,7 +9,8 @@ export default connect(
     isAdmin: sel.isAdmin,
     proposals: sel.vettedProposals,
     error: sel.vettedProposalsError,
-    isLoading: sel.vettedProposalsIsRequesting
+    isLoading: sel.vettedProposalsIsRequesting,
+    header: () => "Vetted Proposals"
```
sndurkin: Let's call this "Active Proposals" for now.

src/components/snew/Content.js
```diff
@@ -11,8 +11,10 @@ export const CustomContent = ({
   proposals,
   isLoading,
   error,
+  header,
   ...props
 }) => {
+  console.log(header);
```
sndurkin: I think you forgot to remove this.

PR #184, opened 2018-01-19
[REVIEW-169]: handling onLogin key mismatch

### LINKED ISSUE #169: Need to display a message when there is a key mismatch between client and server
When the user logs in, the server sends the active public key that it has in the DB for that user. If this public key doesn't match the browser's local key (or there isn't a local key at all), the user should be alerted that he won't be able to perform any actions until it's resolved. The user should be directed to the user profile section for managing his keys so that he can either import an existing private key or generate a new one.

The message should probably be displayed on all pages and, additionally, actions like submitting proposals and comments should be disabled.

This issue sort of depends on #168; the message can be implemented before #168 is closed but will probably have to be slightly amended after it's in.

brunobraga95:
fix #169 
to remove the modal, just follow the instruction at #173 
image


fernandoabolafio:
Something has broken on the login page layout. See how it is on `upstream/master` branch:
screen shot 2018-01-29 at 12 17 53 pm
See how it looks on your branch:
screen shot 2018-01-29 at 12 18 37 pm

Also the proposal containers are overlapping the login form. Making it impossible to be selected. See:
screen shot 2018-01-29 at 12 18 46 pm

src/App.js
```diff
@@ -14,7 +15,12 @@ class Loader extends Component {
   }
 
   render() {
-    return this.props.children;
+    return (
+      <div>
+        { this.props.pubkeyMismatch && this.props.loggedIn && <TopModal label='Key mismatch, please solve this on the user profile screen to proceed with actions' /> }
```
fernandoabolafio: Please, wrap this with the Message component (type = error).
Also link the "user profile" to the user profile route.

brunobraga95: The problem related to the layout will probably be fixed with a simple rebase. About the ErrorWrapper, it is not that straightforward because message component has a `relative` position. After trying it I realized I would need to make a few changes in the CSS, and based on the fact that we will go through a redesign I believe such a thing can be delayed, what you think?

Also, the top modal will always appear as long as the key in the local storage is different than the one received from the api, which is the issue description. I would suggest opening a new issue that is just about this change, not really sure.

About the click to redirect to user profile screen, yeah, I forgot about it, good catch.

src/App.js
```diff
@@ -25,11 +55,9 @@ export default class App extends Component {
     return (
```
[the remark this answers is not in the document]
brunobraga95: I will check this out.

fernandoabolafio:
There still a few problems here:
The overlapping issue is solved but the white wrapper has disappear now, see:
screen shot 2018-02-05 at 8 05 23 pm

About the success screen for key verification, you can detect the key has successfylly update by lookin at the action `API_RECEIVE_VERIFIED_KEY`. It (if occur successfully will update the value `api.verifyUserKey.response` to `{success: true}`. Using that you can hide the modal withou need of refreshing the page.

fernandoabolafio:
LGTM 
Only check why the tests aren't passing and I thinkg it can go in.
@sndurkin can you please review this too?

sndurkin:
Implemented in #222.

Issue #115, opened 2017-11-22
need to display error context on password errors


jolan:
Currently there is no indication to the user what is actually wrong with the password.

thi4go:
Looking into it !

sndurkin:
@thi4go, for this issue we just have to check whether the UI is validating that the password is at least 8 characters, and informing the user if it isn't. It should be pulling the number from `passwordminchars` from [the `/v1/policy` call](https://github.com/decred/politeia/blob/master/politeiawww/api/v1/api.md#policy).

Note that this validation should happen both for registering a new user and for changing the password (on the user profile page).

thi4go:
Thank you for the guidelines @sndurkin ! Finishing this issue

PR #187, opened 2018-01-19
Password length validation for sign up and pwd change

thi4go:
Passed password change validator module to validators folder

Closes https://github.com/decred/politeiagui/issues/115

### DIFF
--- a/src/components/PasswordChange/index.js
+++ b/src/components/PasswordChange/index.js
@@ -4,11 +4,18 @@ import { SubmissionError } from "redux-form";
 import ChangePasswordForm from "./Form";
 import changePasswordConnector from "../../connectors/changePassword";
 import Message from "../Message";
-import validate from "./Validator";
+import validate from "../../validators/password-change";
 
 class ChangePassword extends Component {
+
+  componentDidMount() {
+    this.props.policy || this.props.onFetchData();
+  }
+
   onChangePassword(props) {
-    validate(props);
+    const policy = this.props.policy;
+    
+    validate(props, policy);
 
     return this
       .props
@@ -19,6 +26,7 @@ class ChangePassword extends Component {
         });
       });
   }
+
   render() {
     return (
       <div>
--- a/src/validators/password-change.js
+++ b/src/validators/password-change.js
@@ -1,13 +1,17 @@
 import { SubmissionError } from "redux-form";
-import { isRequiredValidator, passwordVerifyValidator } from "../../validators";
+import { isRequiredValidator, passwordVerifyValidator, passwordLengthValidator } from "./util";
 
-const validate = values => {
+const validate = (values, policy) => {
   if (!isRequiredValidator(values.existingPassword) || !isRequiredValidator(values.password) || !isRequiredValidator(values.password_verify)) {
     throw new SubmissionError({ _error: "All fields are required" });
   }
 
   if (!passwordVerifyValidator(values.password, values.password_verify)) {
     throw new SubmissionError({ _error: "Passwords do not match" });
+  }
+
+  if (!passwordLengthValidator(values.password, policy.minpasswordlength)){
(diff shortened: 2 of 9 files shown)


src/validators/util.js
```diff
@@ -17,11 +17,14 @@ export function proposalNameValidator(name, supportedCharacters) {
   return matches.length === name.length;
 }
 
+export function passwordLengthValidator(password, minLength) {
+  return password.length > minLength
```
sndurkin: This should be `>=`.

src/validators/password-change.js
```diff
@@ -1,14 +1,18 @@
...
   if (!isRequiredValidator(values.existingPassword) || !isRequiredValidator(values.password) || !isRequiredValidator(values.password_verify)) {
     throw new SubmissionError({ _error: "All fields are required" });
   }
 
   if (!passwordVerifyValidator(values.password, values.password_verify)) {
     throw new SubmissionError({ _error: "Passwords do not match" });
   }
+
+  if (!passwordLengthValidator(values.password, policy.passwordminchars)){
+    throw new SubmissionError({ _error: "Password must be at least "+policy.passwordminchars+" digits." });
```
sndurkin: Could you reword this to "Your password must be at least <policy.passwordminchars> characters."?

src/validators/signup.js
```diff
@@ -1,11 +1,15 @@
...
+import { emailValidator, isRequiredValidator, passwordVerifyValidator, passwordLengthValidator } from "./util";
 
-const validate = values => {
+const validate = (policy, values) => {
   if (!isRequiredValidator(values.email) || !isRequiredValidator(values.password) || !isRequiredValidator(values.password_verify)) {
     throw new SubmissionError({ _error: "All fields are required" });
   }
 
+  if (!passwordLengthValidator(values.password, policy.passwordminchars)){
+    throw new SubmissionError({ _error: "Password must be at least "+policy.passwordminchars+" digits." });
```
sndurkin: Same as above.

thi4go:
@sndurkin I implemented the changes, however I don't know why TravisCI is failing at that test... any ideas?

src/validators/util.js
```diff
@@ -17,11 +17,14 @@ export function proposalNameValidator(name, supportedCharacters) {
   return matches.length === name.length;
 }
 
+export function passwordLengthValidator(password, minLength) {
+  return password.length >= minLength
```
oktapodia: Can you add the semicolon, please?

src/components/PasswordChange/index.js
```diff
@@ -4,11 +4,18 @@ import { SubmissionError } from "redux-form";
...
+
+  componentDidMount() {
+    this.props.policy || this.props.onFetchData();
+  }
+
   onChangePassword(props) {
-    validate(props);
+    const policy = this.props.policy;
+    
+    validate(props, policy);
```
oktapodia: Because you already have the `policy` into the `props`, it should be great to keep the same function signature as before, please

src/validators/util.js
thi4go: Added ! sry about that

src/components/PasswordChange/index.js
thi4go: Actually, the props in this function is the argument of onChangePassword(props), which contains only the information of the Form. I could maybe add policy to props.policy, or something like that?

oktapodia: oh! Sorry did not check that!

thi4go:
Fixed Travis CI ! Thanks @oktapodia

oktapodia:
You just forgot to refer the test [here](https://github.com/decred/politeiagui/blob/master/test/e2e/main/signup.js#L19) and we will be ready to go

thi4go:
Done @oktapodia  !! Thanks for the review

PR #194, opened 2018-01-30
add publickey field for verification

AlanL1:
related to https://github.com/decred/politeia/issues/186


AlanL1:
Actually, the publickey field is already there in the New comment method. Maybe the original author planned that all along. I construct the Go code so that it fits nicely into the existing JavaScript code.

Issue #193, opened 2018-01-30
Create Dropdown


tiagoalvesdulce:
Create a Dropdown to improve navigation through politeia's routes (User Proposals, Update Key Pair, Change Password, Settings, etc).

thi4go:
I can work on this, if anybody isn't yet.

tiagoalvesdulce:
I am, @thi4go. Self-assigned the issue.

PR #200, opened 2018-01-31
Dropdown

tiagoalvesdulce:
Closes https://github.com/decred/politeiagui/issues/193


tiagoalvesdulce:
Would it look like this? I think I prefer the menu items centered.

image

src/components/Dropdown.js
```diff
@@ -0,0 +1,65 @@
+import React, {Component} from "react";
```
fernandoabolafio: Please, insert spaces around 'Component' for keep the code pattern. 
To be like this:
`import React, { Component } from "react"`

src/components/Dropdown.js
```diff
@@ -0,0 +1,65 @@
...
+import ReactDOM from "react-dom";
+
+class Dropdown extends Component {
```
fernandoabolafio: Would be nice to have to have Typechecking with `prop-types` for this component since it might be reused in other parts of the codebase.

PR #211, opened 2018-02-16
[WIP] Additional changes to paywall

sndurkin:
This PR adds polling to both dcrdata and decred (insight) block explorers for a potential payment transaction when the user is logged in and hasn't paid the registration fee. When it finds a transaction that meets the paywall requirements, it makes a request to `/v1/user/verifypaymenttx` so that Politeia can update that the user has paid in its database.


src/selectors/api.js
```diff
@@ -115,10 +116,19 @@ export const isAdmin = bool(or(
...
-));
+export const hasPaid = bool(state => {
+  if(state.api.me.response) {
+    return state.api.me.response.paywalladdress === "";
+  }
+  if(state.api.login.response) {
+    return state.api.login.response.paywalladdress === "";
+  }
+  if(state.api.verifyPaywallPaymentServer) {
+    return state.api.verifyPaywallPaymentServer.response.haspaid === true;
```
tiagoalvesdulce: `return state.api.verifyPaywallPaymentServer.response.haspaid` will do the job here

src/actions/external_api.js
```diff
@@ -36,15 +32,24 @@ const checkForPayment = (poll, addressToMatch, amount) => {
...
+  poll.every((transaction) => {
     txid = checkTransaction(transaction, addressToMatch, amount);
+    return !txid;
   });
   return txid;
 };
 
 const checkTransaction = (transaction, addressToMatch, amount) => {
   let addressSeen = false;
   let addressValue = 0;
```
tiagoalvesdulce: These two variables are not being used.

src/selectors/api.js
```diff
@@ -51,6 +51,7 @@ const apiMeResponse = getApiResponse("me");
 const apiInitResponse = getApiResponse("init");
 const apiPolicyResponse = getApiResponse("policy");
 const apiNewUserResponse = getApiResponse("newUser");
+const apiVerifyUserPaymentResponse = getApiResponse("verifyPaywallPaymentServer");
```
tiagoalvesdulce: This is assigned but never used

src/actions/external_api.js
```diff
@@ -36,15 +32,24 @@ const checkForPayment = (poll, addressToMatch, amount) => {
...
+
+  if (transaction.amount > amount &&
```
vctt94: Here should be `>=` I think, shouldn't it?

vctt94:
tests are failing

sndurkin:
@tiagoalvesdulce @vctt94 I think this is ready for another review. One issue I ran into that I couldn't resolve is it didn't seem to be automatically re-enabling things like the Submit Proposal button when the `/user/verifypaymenttx` request returned with `{ "haspaid": true }`. I couldn't figure it out, do you have any idea why?

Also, if you want to test the full end-to-end flow, the corresponding Politeia changes haven't been merged to master yet but they're available in decred/politeia#232.

sndurkin:
> When verifying the payment after polling with RECEIVE_VERIFY_PAYWALL_PAYMENT_EXPLORER, I am always getting a null object as response. I believe because this is a logged in route, and after creating account the user is not considered "logged in", but not really sure. Do you reiceve the {haspaid} response?

I will remove the status on the `/user/signup/next` page and only display it after the user has logged in.

> Even after approving the paywall payment with polling and checking the payment against the server, the status is as waiting for payment. Maybe would be good changing for enough confirmations, checking server or something like that, what do you think?

Yeah, I'll add at least one other status for "waiting for enough confirmations".

> When polling, it is always making both request for both urls, not sure if that is a problem, but maybe we could use request timeouts and only make one request. But we can have an issue for that.

Thanks, I'll make sure to fix this bug.

> with the server, I noticed that even when we have confirmations, the haspaid returns false, sometimes, like with `/v1/user/verifypaymenttx?txid=7b5f5b71dc155fabddac54b548d5f97195753050e65ecfda1b859399ecdc8901`

I've noticed this as well and will do some more testing to fix server-side issues.

> And if we try to signup with a repeated email, we do not receive paywall information, even if not paid.

This is currently the expected behavior of the server.

sndurkin:
Closing this PR in favor of #225.

PR #222, opened 2018-03-20
Set "Key mismatch Message" on Login

### LINKED ISSUE #169: Need to display a message when there is a key mismatch between client and server

fernandoabolafio:
I took over the PR #184 which fixes #169. 
Since I can't commit to the already opened PR itself I've incorporated the changes
here and added a few more things.


tiagoalvesdulce:
LGTM. @sndurkin can you review it as well?

PR #278, opened 2018-04-19
Check vote by block height

### LINKED ISSUE #273: Mark when vote is complete
We need to makr a vote when it has been completed. Easiest is to prob ask dcrdata and compare against the end block of the tally.

vctt94:
fixes #273 

I added a package.json field `network`, which we need to know in order to get the right block height from dcrdata.


src/components/snew/Content.js
```diff
@@ -72,8 +74,22 @@ class Loader extends Component {
...
     }
+    this.props.getLastBlockHeight();
   }
 
+  componentWillReceiveProps(nextProps){
+    if (!nextProps.activeVotes)
+      return;
+    const {activeVotes} = nextProps;
+    const endHeightByToken = {};
+    activeVotes.map((value) => {
```
fernandoabolafio: This section is generating a warning on console
```
 Line 88:  Arrow function expected a return value  array-callback-return
```
You can use `forEach` instead of `map` here.

src/components/snew/ThingLink.js
```diff
@@ -82,15 +84,26 @@ const ThingLinkComp = ({
...
-          Proposal active for voting
-          </span>
+        {
+          vote && activeVotesEndHeight ?
+            activeVotesEndHeight[vote.token] >= lastBlockHeight ?
+              (<span style={{
+                padding: "4px 8px",
+                borderRadius: "8px",
+                fontSize: "12px",
+                color: "green"
```
fernandoabolafio: I think we can use a different color to signalize ended voting proposals. Maybe red?

PR #285, opened 2018-04-25
Fix bug with refreshing the /user/account page

### LINKED ISSUE #281: Refreshing on /user/account doesn't stay on the same page
When on the `/user/account` page, pressing F5 doesn't bring up the same page but goes back to `/login` which redirects to the `/user/proposals` page.

sndurkin:
Fixes #281.


sndurkin:
Travis is currently failing because it's installing the newly released Node 10, which is incompatible with upath (which is a dependency of webpack): https://github.com/anodynos/upath/issues/14

PR #287, opened 2018-04-25
Misc bugfixes and changes

sndurkin:
- Re-enable user action restrictions if paywall hasn't been paid.
- Fix bug where paywallTxNotBefore wasn't being utilized when checking
  block explorers for transactions.
- Fix bug where UI wasn't properly being updated after user registration was paid.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,6 @@ addons:
       - google-chrome-stable
 language: node_js
 node_js:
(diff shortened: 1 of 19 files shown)


sndurkin:
Thanks for your review, I've made the button look more disabled and removed the travis job that was failing (due to a dependency that doesn't work with the latest node version).

PR #294, opened 2018-05-01
Identity Mgmt bugfixes

### LINKED ISSUE #291: More Identity Mgmt issues
There are a couple issues with Identity Mgmt that need to be resolved:

- Formatting bug for the Key Updated message; missing space between email and other text
- Change Key Updated message to say something else, like "Verification Required" and the message should be orange/yellow instead of green so the user knows further action is required
- Replace the "Action Needed" message when the user imports an identity whose public key matches the server's with a success message
- Add to the identity help text that Politeia stores all past identities
- Add help text for the identity mgmt buttons
- Change "secretKey" to "private" or "privateKey" to be consistent

sndurkin:
Fixes #291.


src/components/Verify/Page.js
```diff
@@ -0,0 +1,55 @@
...
+import Message from "../Message";
+import PageLoadingIcon from "../snew/PageLoadingIcon";
+
+
+const Page = ({
+  verifyNewUser,
+  isRequestingVerifyNewUser,
+  verifyNewUserError,
+}) => {
+  console.log(verifyNewUser);
```
fernandoabolafio: This console.log can be removed, right?

sndurkin: Thanks, I removed it.

fernandoabolafio:
LGTM

PR #298, opened 2018-05-02
Improve proposal submit page UX

### LINKED ISSUE #296: Fix awful UX on submit proposal page
Several issues:

1. Error message appears at the top of the page but it doesn't autoscroll into view - it should either be next to the Submit button or should autoscroll into view so the user is aware what's happening when he clicks the Submit button and nothing seems to happen.

2. Error messages for file uploads are extremely vague. Don't make the user guess what's wrong:

   image

3. The page doesn't inform the user at all what the proposal limitations are (upload file size, mime types, etc). This should be made clear to the user in a dismissable `<Message />` at the top.

tiagoalvesdulce:
Closes #296

### DIFF
--- a/src/components/MultipleItemsBodyMessage.js
+++ b/src/components/MultipleItemsBodyMessage.js
@@ -3,7 +3,6 @@ import React from "react";
 const MultipleItemsBodyMessage = ({items}) =>
   Array.isArray(items) ? (
     <ul>
-      {console.log("I am here", items)}
       {items.map((error, i) =>
         <li key={i} style={{padding: "3px 0px"}}>
           {error}
--- a/src/components/MultipleItemsBodyMessage.js
+++ b/src/components/MultipleItemsBodyMessage.js
@@ -1,0 +1,17 @@
+import React from "react";
+
+const MultipleItemsBodyMessage = ({items}) =>
+  Array.isArray(items) ? (
+    <ul>
+      {console.log("I am here", items)}
+      {items.map((error, i) =>
+        <li key={i} style={{padding: "3px 0px"}}>
+          {error}
+        </li>
+      )}
+    </ul>
+  ) : (
+    (items instanceof Error) ? items.message : items
+  );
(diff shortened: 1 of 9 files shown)


src/validators/proposal.js
```diff
@@ -19,14 +18,20 @@ const validate = (values, dispatch, props) => {
 
   if (values.files) {
     if(values.files.length > props.policy.maximages) {
-      throw new SubmissionError({ _error: "Only 5 attachments are allowed" });
+      throw new SubmissionError({ _error: "Only 5 attachments are allowed." });
```
sndurkin: This should use `props.policy.maximages` instead of a hardcoded number.

src/components/ProposalInfo.js
```diff
@@ -0,0 +1,24 @@
...
+
+const ProposalInfo = ({policy}) => (
+  <ul style={{listStyle: "disc"}}>
+    <li style={liStyle}>You must provide both a proposal name and description.</li>
+    <li style={liStyle}>
+      The proposal name must be between {policy.minnamelength} and {policy.maxnamelength} characters long{" "}
+      and only contain the following characters: {policy.supportedcharacters.join(" ")}
+    </li>
+    <li style={liStyle}>Only 5 attachments are allowed.</li>
+    <li style={liStyle}>The maximum size of an attached image is {policy.maximagesize} kb.</li>
```
sndurkin: I believe the `maximagesize` is in bytes, not kb - it's 512kb.

src/components/MultipleItemsBodyMessage.js
```diff
@@ -0,0 +1,17 @@
+import React from "react";
+
+const MultipleItemsBodyMessage = ({items}) =>
+  Array.isArray(items) ? (
+    <ul>
+      {console.log("I am here", items)}
```
sndurkin: Last nit - log statement that should be removed.

PR #343, opened 2018-05-25
Display the correct public key on account page and set message while …

### LINKED ISSUE #342: Wrong public key displayed on account page
- Go to /user/account
- Refresh the page
The public key being displayed isn't correct, now:
- Click on the username and account. 
The page will be updated and the correctly public key will be displayed:

wrong-pk

fernandoabolafio:
…loading it from storage
closes #342

### DIFF
--- a/src/components/AccountPage.js
+++ b/src/components/AccountPage.js
@@ -6,6 +6,7 @@ import Message from "./Message";
 import { myPubKeyHex } from "../lib/pki";
 import Paywall from "./Paywall";
 import accountConnector from "../connectors/account";
+import { PUB_KEY_STATUS_LOADED, PUB_KEY_STATUS_LOADING } from "../constants";
 
 const UpdatedKeyMessage = ({ email }) => (
   <span>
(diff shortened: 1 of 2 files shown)


src/components/AccountPage.js
```diff
@@ -22,17 +23,26 @@ class KeyPage extends React.Component {
...
-    myPubKeyHex(loggedInAsEmail).then(pubkey => {
-      if(!this.unmounting) {
-        this.setState({ pubkey });
-      }
-    });
+    this.resolvePubkey();
+  }
+
+  componentDidUpdate() {
+    this.resolvePubkey();
```
sndurkin: Should we re-resolve the public key on every update?

fernandoabolafio: The function is being called but the code inside only executes if the condition `!this.state.pubkey && this.props.loggedInAsEmail` is fulfilled. I can move the condition out of the `resolvePubKey` if you think it can get more readable.

fernandoabolafio: Btw, I couldn't figure it out with another lifecycle since it is a mix of state change and prop changes.

sndurkin: That's a good point that the condition is already inside the function, I missed that.

PR #349, opened 2018-05-31
[WIP] Add usernames to comments.

### LINKED ISSUE #348: Create username cache
Comments returned by www won't have the username populated, because in proposals with a lot of comments, the username would be duplicate information. It will provide a separate request to fetch usernames by user id, which the UI will use when the proposal details is loaded and request usernames for all user ids it doesn't know about. It will then cache them so it doesn't have to make duplicate requests. The cache will only live while the user stays on Politeia; refreshing the browser or navigating away will wipe it away.

Proposal lists will still have the username sent, so the UI can cache those by user id as they're received.

sndurkin:
This PR depends on decred/politeia#305.

Closes #348.


sndurkin:
@fernandoabolafio, could you provide some guidance on how best to test this new functionality, and also the `verify user payment with politeia action` test?

src/actions/app.js
```diff
@@ -68,3 +75,39 @@ export const onLocalStorageChange = (event) => (dispatch, getState) => {
...
+  // to make a server request.
+  if(userIdsToFetch.length === 0) {
+    return dispatch(act.RECEIVE_USERNAMES({ usernamesById }));
+  }
+
+  return dispatch(onFetchUsernamesByIdApi(userIdsToFetch))
+    .then(() => {
+      const apiUsernamesByIdResponse = get(getState(), ["api", "usernamesById", "response"], undefined);
+      if(apiUsernamesByIdResponse) {
+        for(let i = 0; i < userIdsToFetch.length; i++) {
```
fernandoabolafio: Instead of for loops, you could use `userIdsToFetch.forEach(userId => {...})`. Its just a more functional syntax.

src/selectors/app.js
```diff
@@ -51,3 +62,34 @@ export const userCanExecuteActions = state => {
...
+  for(let comment of comments) {
+    if(comment.userid in globalUsernamesById) {
+      comment.username = globalUsernamesById[comment.userid];
+    }
+  }
+
+  return comments;
+};
+
+export const unvettedProposals = state => {
```
fernandoabolafio: What do you think if, instead of using the selectors for caching the proposal usernames, use the store subscription  (https://github.com/decred/politeiagui/blob/master/src/App.js#L18) to detect changes on unvetted/vetted response and caching the values there? So we can create a module like `handleCacheUsernamesFromProposals`. It would keep selectors cleaner. It's just an implementation suggestion if you find the current approach is good enough we can tackle the tests for it as well.

sndurkin: I'm fine with doing that as long as we don't have to preserve the cache in local/session storage?

sndurkin:
Regarding your comments:

1. Implemented this in my most recent commit.
2. Do you think this test is really worth it? The code it would be verifying is simply: `author: username || userid`
3. I added this in my most recent commit, but didn't include any parameters; I think in order to test this more thoroughly, I would have to create a proposal with a comment, and then fetch it?
4. I'm holding off on this implementation because I'm assuming this approach will change once I implement it through the store subscription, right?

src/selectors/app.js
fernandoabolafio: You don't need to use local/session storage. I would recommend you creating a module (such as lib/cache_usernames.js) the you can export the variable ‘globalUsernamesById‘ from there and all needed handlers functions. This way we keep all the related code to caching grouped together.

fernandoabolafio:
About your last comment:
1 -great!

2- Isn't that much effort to test it. You can modify the test "comments to T1" inside `lib/test/snew.test.js` to:
```javascript
test("comments to T1", async() => {
    const { comments } = await import(`../../../mocks/api/v1/proposals/${PROPOSAL_TOKEN}/comments/GET.json`);
    let result = snew.commentsToT1(comments);
    let comment = comments[0];
    let resultComment = result[0];
    let { kind, data } = resultComment;
    expect(kind).toEqual("t1");
    expect(data.author).toEqual(comment.userid);
    expect(data.parent_id).toEqual("0");
    expect(data.name).toEqual(comment.commentid);
    expect(data.body).toEqual(comment.comment);
    expect(data.permalink).toEqual(`/proposals/${comment.token}/comments/${comment.commentid}`);

    //test username will be used instead of id for author
    comment.username = "foobar";
    result = snew.commentsToT1([comment]);
    resultComment = result[0];
    expect(resultComment.data.author).toEqual(comment.username);
  });
```
I've tested the code above with your latest commit and the test is passing :).
 
3- I don't think you need to create it before. You can mock the prop and the comment and fetch it. In order to mock the response for a given route you can use:
```
fetchMock.get(path, response)
```
4- Right

sndurkin:
@fernandoabolafio, I tried to address your concerns about the unit tests and wasn't able to fix them. I wasted a couple hours on it so I'm merging this and opening issues for the remaining stuff so I can move on - it's cutting too much into my productivity.

#364 
#365

PR #350, opened 2018-06-01
Add specific messages for empty proposal lists.

### LINKED ISSUE #199: Empty proposals list message is too generic
Currently it says "There are no proposals yet." This message is too generic; after #177 goes in, we will we have multiple different proposal lists, so it should be tailored specifically to each list.

We should have a unique message for at least the following:

* Public proposals
* Admins' proposals (unreviewed/censored)
* My proposals
* Another user's proposals

sndurkin:
Also change the `<select>` filter field on the admin proposals list page to a radio group, for easier switching.

Fixes #199.


fernandoabolafio:
The radio buttons are indeed way better than the <select>. Everything else also LGTM.

PR #353, opened 2018-06-02
Fix bug with generating a new identity.

### LINKED ISSUE #351: Bug when trying to generate a new identity
When clicking the "Generate new identity" button, it creates a new identity locally and then sends a request to Politeia, which sends an email to the user to verify it. If the user clicks the button again (accidentally or just out of curiosity), it generates a new identity and sends another request to Politeia, but this request fails because the original identity verification email hasn't expired yet. But, the 2nd new identity has already been generated locally, so it's impossible for the user to verify the 1st identity by clicking on the email link.

To fix this issue, I think the UI should revert back to the original identity if the server request fails.

sndurkin:
Fixes #351.

### DIFF
--- a/src/actions/api.js
+++ b/src/actions/api.js
@@ -230,13 +230,11 @@ export const onUpdateUserKey = (loggedInAsEmail) =>
   withCsrf((dispatch, getState, csrf) => {
     dispatch(act.REQUEST_UPDATED_KEY());
     return pki.generateKeys().then(
-      keys => pki.myPubKeyHex(loggedInAsEmail).then(
-        publickey => api.updateKeyRequest(csrf, publickey)
-          .then(response => {
-            pki.loadKeys(loggedInAsEmail, keys);
-            return dispatch(act.RECEIVE_UPDATED_KEY({ ...response, success: true }));
-          })
-      )
+      keys => api.updateKeyRequest(csrf, pki.toHex(keys.publicKey)).then(
+        response => {
+          pki.loadKeys(loggedInAsEmail, keys);
+          return dispatch(act.RECEIVE_UPDATED_KEY({ ...response, success: true }));
+        })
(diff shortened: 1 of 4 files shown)


src/actions/api.js
```diff
@@ -228,9 +229,18 @@ const statusName = key => ({ 3: "censor", 4: "publish" }[key]);
 export const onUpdateUserKey = (loggedInAsEmail) =>
   withCsrf((dispatch, getState, csrf) => {
     dispatch(act.REQUEST_UPDATED_KEY());
-    return api.updateKeyRequest(csrf, loggedInAsEmail)
-      .then(response => dispatch(act.RECEIVE_UPDATED_KEY({ ...response, success: true })))
-      .catch(error => { dispatch(act.RECEIVE_UPDATED_KEY(null, error)); throw error; });
+    return pki.generateKeys().then(
+      keys => pki.myPubKeyHex(loggedInAsEmail).then(
+        publickey => api.updateKeyRequest(csrf, publickey)
```
fernandoabolafio: Here you are generating new keys but you are sending the old pubkey to the server. 
Since you've changed `pki.generateKeys()` to only generate a new key pair without saving it to localforage you cannot request `pki.myPubkeyHex` because it will give you the current saved pubkey.
You could change this to:
```javascript
return pki.generateKeys().then(
      keys => api.updateKeyRequest(csrf, pki.toHex(keys.publicKey))
        .then(response => {
          pki.loadKeys(loggedInAsEmail, keys);
          return dispatch(act.RECEIVE_UPDATED_KEY({ ...response, suc
```

sndurkin: Good catch @fernandoabolafio, thanks! I've updated the PR.

fernandoabolafio:
LGTM

PR #371, opened 2018-06-13
Fix comment self reply

### LINKED ISSUE #369: [comments] Can't reply to self after creating comment.
1. Create or reply to comment
2. Click on reply to the new comment, doesn't work
3. Refresh page
4. Reply now does work

fernandoabolafio:
closes #369 
I also added two minor fixes for:
 - Clear a react warning on comments listing
 - Remove an unnecessary console log on MarkdwonEditorsField component


tiagoalvesdulce:
LGTM

PR #390, opened 2018-06-25
Fix issues with paying with testnet faucet.

sndurkin:
This depends on decred/testnetfaucet#29 being merged.

### DIFF
--- a/src/selectors/external_api.js
+++ b/src/selectors/external_api.js
@@ -1,12 +1,19 @@
 import get from "lodash/fp/get";
+import compose from "lodash/fp/compose";
 import { bool, or } from "../lib/fp";
 
 const getIsApiRequesting = key => bool(get(["external_api", key, "isRequesting"]));
 const getApiResponse = key => get(["external_api", key, "response"]);
+const getApiError = key => get(["external_api", key, "error"]);
 
 export const isApiRequestingPayWithFaucet = or(
   getIsApiRequesting("payWithFaucet"),
   getIsApiRequesting("blockHeight"),
+);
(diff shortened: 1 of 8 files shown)


src/selectors/external_api.js
```diff
@@ -1,12 +1,19 @@
...
 const getIsApiRequesting = key => bool(get(["external_api", key, "isRequesting"]));
 const getApiResponse = key => get(["external_api", key, "response"]);
+const getApiError = key => get(["external_api", key, "error"]);
 
 export const isApiRequestingPayWithFaucet = or(
   getIsApiRequesting("payWithFaucet"),
   getIsApiRequesting("blockHeight"),
 );
 
+export const payWithFaucetError = or(
```
fernandoabolafio: It looks a bit weird. 
Why don't you change the action to:
```javascript
export const payWithFaucet = (address, amount) => dispatch => {
   dispatch(act.REQUEST_PAYWALL_PAYMENT_WITH_FAUCET());
   return external_api.payWithFaucet(address, amount)
     .then(() => dispatch(act.RECEIVE_PAYWALL_PAYMENT_WITH_FAUCET({faucetPayment: true})))
     .then(json => json.Error ?
     dispatch(act.RECEIVE_PAYWALL_PAYMENT_WITH_FAUCET(null, json.Error)) :
     dispatch(act.RECEIVE_PAYWALL_PAYMENT_WITH_FAUCET({faucetPayment: true}))
}).catch(...)
```
By doing that you can use only `getApiError("payWithFaucet")`

PR #393, opened 2018-06-26
Add LoginSignup Modal

### LINKED ISSUE #386: Handle like comments errors (user logged out)
When an error occurs while trying to vote up/down on a comment the GUI isn't handling it.
There are two main cases to be handled:
- User is logged out. 
Right now the score just doesn't change. It should at very least display an error message but ideally, it should instruct the user to login (similar to reddit).

tiagoalvesdulce:
Closes #386

### DIFF
--- a/src/actions/api.js
+++ b/src/actions/api.js
@@ -2,7 +2,7 @@ import Promise from "promise";
 import * as sel from "../selectors";
 import * as api from "../lib/api";
 import * as pki from "../lib/pki";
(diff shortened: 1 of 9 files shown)


tiagoalvesdulce:
Hey @fernandoabolafio, thanks for the review. I addressed two of your concerns. I think that the signup bug should be solved in a different PR, so I'd recommend we get this PR in and open an issue for that.

PR #398, opened 2018-06-29
Auto verify key when in testnet

### LINKED ISSUE #356: Auto-verify new indentity and new user
When the verification token is returned from the server the UI can auto-redirect to the respective verification route to speed up generating new identities. 
For instance, when clicking "Generate new identity" the server will return:
```
{"verificationtoken":"2c68763844920c14719a66914f2d6eed85f39eb173ccd0d2e0b05e1578b05da5"}
```
Then we can redirect to:
`/user/key/verify/?verificationtoken=2c68763844920c14719a66914f2d6eed85f39eb173ccd0d2e0b05e1578b05da5`

The same has to be done for verifying a new user.
This will speed up testing and avoid losing the token sent by the server.

vctt94:
closes #356


vctt94:
This PR is ready for another review

PR #409, opened 2018-07-02
Issue 379: No server running on development

### LINKED ISSUE #379: Internal server error needs to be handled gracefully
Steps to reproduce: 
 - Shut down `politeiawww`
 - Go to `https://localhost:3000` or wherever `politeiagui` is running on your machine
screen shot 2018-06-19 at 12 21 11 pm

rrecuero:
Closes #379


rrecuero:
@fernandoabolafio do you mean an actual ui component or a browser alert window is enough?

fernandoabolafio:
@rrecuero I am thinking about a UI component, nothing really fancy though. Maybe we could have a route for server errors? So every time we detect the server error (e.g status code > 500) we redirect to that route where we present a friendly message.

rrecuero:
@fernandoabolafio  Like that idea. I'll implement a 500 route

rrecuero:
@fernandoabolafio following your suggestion, I created a 500 route and redirected to it if init fails.

We can also remove all the other throws in api and capture them all in the loader component if we want.

What do you think?

rrecuero:
thank you @fernandoabolafio . I was aware of some of these issues, I just wanted to know if you envision this path to the be the right one. I'll fix the issues this afternoon and send it back your way.

fernandoabolafio:
@rrecuero np, just make sure to add "[wip]" while you think the PR is not completely finished yet.
About "removing all the other throws from the API" I am not sure if I understood what you meant. Anyway, I think we can keep this PR specific for the 500 route so it is easier to review. 
We can discuss further enhancements of the error handling mechanism on slack or you can write an issue proposing that. Sounds good?

src/App.js
```diff
@@ -57,17 +60,15 @@ class Loader extends Component {
...
+      <div className="appWrapper">
+        <ModalStack />
+        {this.props.children}
+      </div>
     );
   }
 }
 
-const LoaderComponent = loaderConnector(Loader);
+const LoaderComponent = loaderConnector(withRouter(Loader));
```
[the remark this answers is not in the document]
rrecuero: Thank you

rrecuero:
@fernandoabolafio this should fix the issues. I am not sure about the one you mention that happened after waiting. I waited for several minutes and didn't see it. If you can repro it again, can you give me specific steps

src/components/ErrorPage.js
```diff
@@ -6,8 +6,25 @@ const ErrorPage = ({ error }) => (
     <div className="page error-page">
       <Message
         type="error"
-        header="Server error"
-        body={error} />
+        header="Internal Server Error"
```
sndurkin: To be consistent with other message headers, can you change this to "Internal server error"?

src/components/ErrorPage.js
```diff
@@ -6,8 +6,25 @@ const ErrorPage = ({ error }) => (
...
+        children={
+          <div>
+            <p>
+              Something is wrong with the server.
```
sndurkin: This is redundant and can be removed, because the error itself should say "The server encountered an unexpected error, please contact Politeia administrators and include the following error code: X"

rrecuero:
@sndurkin Made those changes. I also added a wrapper component for the error page so it always displays the last api error. This way you can either:

- Use just the error message and pass the error to the message
- Or redirect to /500 and the connector will automatically pass the last api error to ErrorMessage

Right now, we are only automatically capturing the init error in `App.js` and redirecting to `/500` but we can either modify the loader selector to grab any `ApiError` or decide on an individual basis whether this error is a fatal error and push `/500/` in that case

src/components/ErrorPage/index.js
```diff
@@ -0,0 +1,18 @@
+import React, { Component } from "react";
+import { autobind } from "core-decorators";
+import ErrorMessage from "./ErrorMessage";
+import changePasswordConnector from "../../connectors/changePassword";
```
sndurkin: Shouldn't this be using the `errorPage` connector?

src/components/ErrorPage/ErrorMessage.js
```diff
@@ -0,0 +1,29 @@
+import React from "react";
+import Message from "../Message";
+
+const ErrorMessage = ({ error }) => (
+  <div className="content" role="main">
+    <div className="page error-page">
+      <Message
+        type="error"
+        header="internal server error"
```
sndurkin: The "I" should be capitalized.

src/components/ErrorPage.js
```diff
@@ -6,12 +6,9 @@ const ErrorPage = ({ error }) => (
     <div className="page error-page">
       <Message
         type="error"
-        header="Internal Server Error"
+        header="internal server error"
```
sndurkin: The "I" should be capitalized.

src/components/ErrorPage/ErrorMessage.js
```diff
@@ -0,0 +1,29 @@
+import React from "react";
+import Message from "../Message";
+
+const ErrorMessage = ({ error }) => (
```
sndurkin: This should probably be named `InternalServerErrorMessage` instead, because it's specifically for displaying that type of error.

src/components/ErrorPage/index.js
rrecuero: yes, sorry about that

src/components/ErrorPage/ErrorMessage.js
rrecuero: changed

src/components/ErrorPage/ErrorMessage.js
rrecuero: good call

rrecuero:
ping @fernandoabolafio

rrecuero:
@fernandoabolafio  Added several changes:

- The error message gets passed through the URL so it is persisted and easy to share and tell the users to copy the address if something went wrong.
- In App.js we are listening to all apiError and if it is an internal error we automatically go to 500.
- Changed logout page to use the same pattern and redirect to 500 if there is an error. This pattern can be used in any other components to send non internal errors to the 500 route

PR #410, opened 2018-07-02
Adds link instructions to signup confirmation

### LINKED ISSUE #295: Ensure the user knows to verify email links with the same browser
The user needs to open verification links sent via email with the same browser that was used to generate them, because the UI signs them with the identity. The user needs to be aware of this for each of these cases (e.g. new user registration, etc).

rrecuero:
Closes #295


rrecuero:
thanks @fernandoabolafio. Updated it

fernandoabolafio:
@rrecuero your PR looks good. I added two minor changes.
One of them was my mistake, the next page after password reset doesn't need the "open with the same browser" instruction. Only action which modify the key pair stored in the browser needs that message.

PR #422, opened 2018-07-06
Fix infinite request bug when no csrf is available.

sndurkin:


fernandoabolafio:
it was closed on #424

PR #426, opened 2018-07-07
Fix signup gaps

sndurkin:
This PR is the corresponding UI work for decred/politeia#349. It does the following:

* Adds a "Resend verification email" link to the signup page.
* Warns the user about identities before signing up so he's aware of the verify-with-same-browser rule.

Closes #410.


src/connectors/signup.js
```diff
@@ -49,24 +49,32 @@ class Wrapper extends Component {
...
-        throw new SubmissionError({
-          _error: e.message,
-        });
-      });
+
+    if (!this.props.isShowingSignupConfirmation) {
+      return this.props.onSignup();
     }
+
+    return this.props.onSignupConfirm(args).catch(e => {
```
tiagoalvesdulce: You are not passing `onSignupConfirm` action to this `connector`.

src/connectors/signup.js
```diff
@@ -49,24 +50,32 @@ class Wrapper extends Component {
...
   onSignup(args) {
     args = {
       ...args,
-      email: args.email.trim()
+      email: (args.email || "").trim()
     };
+
     const policy = this.props.policy || {};
     validate(policy, args);
-    if (this.props.onSignup(args)) {
```
tiagoalvesdulce: This if statement is also important, basically we want to make sure `this.props.onSignupConfirm(args)` (in the case of your PR) exists. `this.props.onSignupConfirm(args)` can be `undefined` eventually and we have to make sure it's not before returning `this.props.onSignupConfirm(args).catch(e)...`

PR #447, opened 2018-07-13
Make UX improvements to the proposal lists.

### LINKED ISSUE #440: Minor improvements to the public proposal list filters
2 improvements:

* We should display the number of proposals in each category, so the user can know right away if they're empty or not:

   image

* If there are no proposals in active voting, then the default filter should be pre-voting

sndurkin:
Closes #440.


fernandoabolafio:
LGTM

PR #448, opened 2018-07-13
Block up/down comment vote on active/finished proposals

fernandoabolafio:
This PR is the corresponding work of https://github.com/decred /politeia/pull/386. However, it doesn't depend on it to get merged.
Only allow comment up/down voting for proposals which it’s voting hasn’t started


tiagoalvesdulce:
LGTM

PR #457, opened 2018-07-14
Fix multiple request on signup

### LINKED ISSUE #434: Forbidden when signing up
I get a forbidden error when signing up as soon as I submit the information. (I also get that error for "resend verification" and "reset password") This could be caused by a past account for test proposals for which registration never fully completed. 

But if it's an old account I don't believe the system should give that kind of error because it could leak information about my email address being used for an account already.

Other informations: I tried mutiples usernames and passwords to be sure it wasn't content related.

image

vctt94:
fixes #434

### DIFF
--- a/src/connectors/signup.js
+++ b/src/connectors/signup.js
@@ -22,6 +22,7 @@ const signupFormConnector = connect(
     apiNewUserError: sel.apiNewUserError,
     apiVerifyNewUserError: sel.apiVerifyNewUserError,
     isShowingSignupConfirmation: sel.isShowingSignupConfirmation,
+    csrf: sel.csrf,
   }),
   {
     onFetchData: act.onGetPolicy,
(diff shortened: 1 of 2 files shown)


vctt94:
Thanks for reviewing my PR. It is fixed now

sndurkin:
Thanks, that issue is fixed. I'm noticing one last problem, though: If I am on the site and I clear cookies and try to login, I get a forbidden error (which is fine, because I cleared the cookies without pressing F5). But then I click the "Back to home" button and try to log in again, and I still get a forbidden error. Is that fixable?

vctt94:
I do not think so @sndurkin. We do not touch cookies, who controls it is the browser and the server. If the client clear cookies there is not a way of the server to know who is making the request, without reloading the page and getting a new csrf.

PR #463, opened 2018-07-15
Issues 428 infobox

### LINKED ISSUE #428: Remove the "Important information about proposals" box
We should remove the "Important information about proposals" box as it takes up too much vertical space, which should be allocated to the proposal description and attachments. Here's how we can do that:

* Resize the Submit button to a normal width
* Disable the Submit button when proposal name or description is missing and add a little `?` icon next to it which has a tooltip explaining what is needed.
* Automatically prevent the user from entering invalid characters in the proposal title, and display a message below the title indicating only certain chars are supported, and also indicate when the user is getting close to 80 characters.
* Display a message next to the "Attach" button indicating how many attachments remaining the user can upload - also show the max file size and expected mime types.
* Rename "Attach" to "Attach a file"

rrecuero:
Closes #428 

- Removes info box
- Adds synchronous validation and synchronous warning for the proposal name length
- Adds errors as you type when description or name is missing or the name has weird characters or extra length
- Disables submit button when errors or submitting
- Cleans up layout

screen shot 2018-07-15 at 1 40 33 pm
screen shot 2018-07-15 at 1 40 40 pm
screen shot 2018-07-15 at 1 40 50 pm

@sndurkin


rrecuero:
Fixes all the issues but still uses the Message component. It is much easier to do it this way instead of passing the errors individually and choosing where to display them.

Also cleaned up the display of the images:

screen shot 2018-07-15 at 4 40 51 pm

Let me know what you think @sndurkin

rrecuero:
Thanks @fernandoabolafio @sndurkin . I'll change the errors/warnings to be inline below the inputs/textarea so the page doesn't jump around and the user doesn't get distracted but still gets the info.

I'll change it tonight or tomorrow

rrecuero:
ready again @sndurkin @fernandoabolafio. Made a couple changes:

- Created an InputFieldWithError that displays errors and warnings from Redux form
- Modified our MarkdownEditorField to support this as well
- Adjusted the design and removed the error message boxes. Now they display below the appropriate field.

cc @marcopeereboom

sndurkin:
Tested & working great, I noticed one or two minor issues which we can clean up in follow-up commits.

PR #488, opened 2018-07-30
Enable comments sorting by “new”, “old” and “top”

### LINKED ISSUE #444: Comment sorting options
Comments seem to be sorted newest-first, which will be a mess on any proposals that see sustained discussion. Some options would be good. I don't know if access to reddit's Best algorithm is easy, but basic options like Old, New and Top-scoring would be nice.

fernandoabolafio:
closes #444 
depededs on https://github.com/decred/snew-classic-ui/pull/7

screen shot 2018-07-30 at 3 50 59 pm


src/components/ProposalDetail.js
```diff
@@ -1,12 +1,11 @@
 import React from "react";
-import get from "lodash/fp/get";
 import orderBy from "lodash/fp/orderBy";
 import { Content } from "./snew";
 import { commentsToT1, proposalToT3 } from "../lib/snew";
 import { getTextFromIndexMd } from "../helpers";
 import Message from "./Message";
+import { SORT_BY_NEW, SORT_BY_OLD, SORT_BY_TOP } from "../constants";
```
brunobraga95: nice.

fernandoabolafio:
@sndurkin I agree with you that it would be better to not update the sort immediately when comments are voted. I am trying to figure out a way to make the list only re-sort when the sort option has changed.

fernandoabolafio:
@sndurkin I addressed your comments. Now every update case is handled separately (e.g: new comment, comment vote, sort option change). The logic to implement it got a bit heavy so I moved it to `ProposalDetails/helpers.js` so the component file itself is kept cleaner.

PR #490, opened 2018-07-31
Change proposal filter options to tabs.

sndurkin:


tiagoalvesdulce:
I'd recommend renaming tabIndex state (controls which tab is active) because it can be confused with the tabIndex `jsx` attribute.

tiagoalvesdulce:
LGTM

PR #495, opened 2018-07-31
Add tooltip component and apply it to TimeAgo

### LINKED ISSUE #492: Hovering over the proposal relative date should show the absolute date
Hovering over this relative date under each proposal should show the absolute date in a tooltip or something, so the user can see when exactly it was submitted:

image

tiagoalvesdulce:
Closes #492

### DIFF
--- a/src/components/snew/ThingLink.js
+++ b/src/components/snew/ThingLink.js
@@ -11,10 +11,11 @@ import {
   PROPOSAL_STATUS_UNREVIEWED,
   PROPOSAL_VOTING_NOT_STARTED
 } from "../../constants";
-import { getProposalStatus } from "../../helpers";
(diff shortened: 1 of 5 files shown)


src/components/snew/ThingLink.js
```diff
@@ -96,10 +97,22 @@ const ThingLinkComp = ({
       <p className="tagline">
         <span className="submitted-by">
           submitted{" "}
-          <TimeAgo
-            style={{ cursor: "pointer" }}
-            datetime={created_utc * 1000}
-          />
+          <Tooltip
+            text={dateConverter(created_utc)}
```
sndurkin: I think we could just use `new Date(created_utc * 1000).toString()` here.

PR #503, opened 2018-08-06
Fix bug when traversing children nodes on Markdown render

fernandoabolafio:


src/components/snew/Markdown/helpers.js
```diff
@@ -1,9 +1,10 @@
 import React from "react";
 import xssFilters from "xss-filters";
 
-export const transverseChildrens = (el, cb) => {
-  const children = el.children || (el.props && el.props.children);
-  const filterChildren = (c) => c.map(child => transverseChildrens(child, cb));
+export const traverseChildrens = (el, cb) => {
```
tiagoalvesdulce: Can you rename it to `traverseChildren`?

PR #507, opened 2018-08-08
Improvements on local storage handling for state modifications

### LINKED ISSUE #494: Persist draft proposals on logout
Right now, if a user logs out and refreshes the browser, the draft proposals are gone.
We should manage to keep some parts of the state, such as draft proposals, even after log out.

### LINKED ISSUE #510: Allow users to reach submit proposal page
Since now there is an option to save the proposal as a draft, it makes sense aways let the user access the submit proposal route, where he will be able to save the proposals as a draft and submit it later.

fernandoabolafio:
This PR intends to improve how the state is being saved in the local storage by:
 - [x] Save the state under a key which references the user email (e.g: state-email@example.com)
 - [x] Allow specifying which parts of the state should be cleaned up on logout
-  [x] Draft Proposals Modifications
    - [x] Use unidirectional data flow for draft proposals `component -> store -> local storage`
    - [x] Use unique id for draft proposal keys
- [x] Fix/add tests
closes #510 
closes #494

Good to test for this PR:
- Multiple tabs react to user logout/login
- Draft proposals can be saved/retrieved and deleted
- Editor's content isn't lost when navigating away from the submit proposal route


tiagoalvesdulce:
It looks really good. I just found a bug with two tabs in the same draft. To reproduce:
1. Open two tabs in the same draft
2. Edit one of the drafts and save it
3. Go to the other tab and refresh it
4. GUI will break

PR #511, opened 2018-08-11
Hash user’s password before sending it to the server

### LINKED ISSUE #443: Hash user's password before sending to server
We should hash the user's password during signup/login/change password before sending it to the server:

*  https://www.npmjs.com/package/bcrypt
* use technique 2

fernandoabolafio:
closes #443 

**Note:** In case someone is wondering why I haven't used the crypto-js lib (which was included already in the project) is because their SHA3-256 algorithm is slightly different from the standard one. From their [docs](https://code.google.com/archive/p/crypto-js/):
> SHA-3
[...]
>NOTE: I made a mistake when I named this implementation SHA-3. It should be named Keccak[c=2d]. Each of the SHA-3 functions is based on an instance of the Keccak algorithm, which NIST selected as the winner of the SHA-3 competition, but those SHA-3 functions won't produce hashes identical to Keccak.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,6 +10,7 @@
     "font-awesome": "^4.7.0",
     "js-file-download": "^0.4.1",
     "js-sha256": "^0.7.1",
+    "js-sha3": "^0.8.0",
     "localforage": "^1.5.4",
     "lodash": "^4.17.4",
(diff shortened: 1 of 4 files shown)


tiagoalvesdulce:
LGTM

PR #514, opened 2018-08-14
Enable user details page access to anyone

fernandoabolafio:
This PR is the corresponding work of decred/politeia#415

### DIFF
--- a/src/components/UserDetail/GeneralTab.js
+++ b/src/components/UserDetail/GeneralTab.js
@@ -58,8 +58,26 @@ const TokenFields = ({
   </div>
 );
 
+const IdentityField = ({ identities }) => (
+  identities && identities.length ? (
+    <div>
+      {identities.map(id => {
+        if(id.isactive) {
+          return (
+            <Field key={id.pubkey} label="Active identity">{" " + id.pubkey + " "}</Field>
+          );
+        }
+
+        return null;
+      })}
+      <FieldSeparator />
+    </div>
+  ) : null
+);
+
 const GeneralTab = ({
   user,
+  loggedInAsUserId,
   dcrdataTxUrl,
   isApiRequestingMarkAsPaid,
   isApiRequestingMarkNewUserAsExpired,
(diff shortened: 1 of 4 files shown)


src/components/UserDetail/GeneralTab.js
```diff
@@ -58,115 +58,125 @@ const TokenFields = ({
   </div>
 );
 
+const IdentityField = ({ identities }) => (
```
brunobraga95: what you think about this to improve readability (may need better idendation)?

```
const IdentityField = ({ identities }) => { 
    if(identities && identities.length) {
      const activies = identities.filter((id) => id.isactive);
      return (
        <React.Fragment> // this does not add a unnecessary div to the DOM
          {identities.map(id =>
             ( <Field key={id.pubkey} label="Active identity">{" " + id.pubkey + " "}</Field>))}
          <FieldSeparator />
        </React.Fragment>
      )
    }
    return null;
}
```

src/components/UserDetail/GeneralTab.js
```diff
@@ -58,115 +58,125 @@ const TokenFields = ({
...
-      </div>
-    )}
-    {user.newuserpaywalltx && ([
-      <Field label="Transaction">
-        {user.newuserpaywalltx === "cleared_by_dbutil" ?
-          <span>Cleared by dbutil</span> :
-          <a href={dcrdataTxUrl + user.newuserpaywalltx} target="_blank">{user.newuserpaywalltx}</a>
-        }
-      </Field>,
+  !(isAdmin || (loggedInAsUserId === user.id))  ?
```
brunobraga95: what about applying some boolean algebra and make this
`!isAdmin && loggedInAsUserId !== user.id`

In my opnion it is easier to understand and reason about.
also maybe it would be good to extrack to a variable that gives it a bit more meaning.

something like:
`const isNotLoggedIn = !isAdmin && loggedInAsUserId !== user.id`

Not sure this name makes sense in the context, but you got the idea.

src/components/UserDetail/GeneralTab.js
```diff
@@ -58,115 +58,125 @@ const TokenFields = ({
...
+    <div className="detail-form">
+      <IdentityField identities={user.identities} />
+    </div> :
+    <div className="detail-form">
+      {!user.newuserverificationtoken ? (
+        <Field label="Verified email">Yes</Field>
+      ) : (
+        <div>
```
brunobraga95: consider using  ´Fragment´ from react 16, this way you wont need to add somewhat of a useless div to the DOM. There are a few benefits of making the DOM smaller, but a quick one to think about is that it always easier to debug the dom tree when you dont have things you dont actually need.

src/components/UserDetail/GeneralTab.js
```diff
@@ -58,115 +58,125 @@ const TokenFields = ({
...
+          <span>No</span>,
+          isAdmin && <ButtonWithLoadingIcon
+            className="c-btn c-btn-primary button-small"
+            text="Mark as paid"
+            disabled={isApiRequestingMarkAsPaid}
+            isLoading={isApiRequestingMarkAsPaid}
+            onClick={() => onEditUser(user.id, EDIT_USER_CLEAR_USER_PAYWALL)} />
+        ])}
+      </Field>
+      {user.newuserpaywalladdress && user.newuserpaywallamount && (
```
brunobraga95: what about extracting this to a variable with some meaningful name?

src/components/UserDetail/GeneralTab.js
```diff
@@ -58,115 +58,125 @@ const TokenFields = ({
...
+        <div>
```
brunobraga95: `React.Fragment`

src/components/UserDetail/Page.js
```diff
@@ -41,11 +41,11 @@ const UserDetailPage = ({
             </div>
             <div className="detail-email">{user.email}</div>
             <Tabs>
-              <Tab
+              {user.isadmin && <Tab
```
brunobraga95: consider:

   {user.isadmin ? <Tab
                title="General"
                selected={tabId === USER_DETAIL_TAB_GENERAL}
                tabId={USER_DETAIL_TAB_GENERAL}
                onTabChange={onTabChange} /> : null}

when you don't actually say you will return `null` when the expression fails, you end up returning `false`, which makes sense. The "problem" is that `false` is a valid value, and you will end up adding it the DOM. On the other hand, `null` is not added to the DOM tree.

Not a huge deal, but I am sure that is not what you want, up to you.

fernandoabolafio:
Thanks for your review @brunobraga95! I really like the `React.Fragment` option, I didn't know about this React feature.

tiagoalvesdulce:
LGTM

PR #515, opened 2018-08-17
Allow users to edit proposals

fernandoabolafio:
**This PR implements the modifications on the GUI to enable proposals editing**

This is the corresponding GUI work of: https://github.com/decred/politeia/pull/420 and https://github.com/decred/politeia/pull/424

- [x] Open submission page pre-populated with the existent proposal content
- [x] Integrate edit proposal endpoint
- [x] Add/fix tests


tiagoalvesdulce:
Unreviewed proposals are changing to the same status as "voting started" proposals. See the image below.

screen shot 2018-08-28 at 15 34 56

Works really nice for vetted proposals but I think the user be able to download all the different versions of a proposal on "Download Proposal Bundle"? It's being saved on d, right?

Another really small issue is that the unreviewed proposal version doesn't change on edit, which is fine but can lead to confusion among users. I'd say we shouldn't show version for unvetted proposals.

Everything else looks good.

fernandoabolafio:
Thanks for the review @tiagoalvesdulce. Regarding your comments respectively:

- It should not happen. I will check what is going on.
- The user will be able to retrieve different versions in the future but that hasn't been implemented on the backend yet. So, for now, it will be the user responsibility to keep the different versions of his proposals bundles saved. 
- Yeah, I think is fine to hide the version info for unvetted props.

tiagoalvesdulce:
LGTM

tiagoalvesdulce:
LGTM

Issue #506, opened 2018-08-08
Implimenting CSP


degeri:
Since the politeia website is going to be the core of the Decred ecosystem an XSS vulnerability in the website could prove disastrous. Though utmost care has been placed to make sure we are XSS free before public release I suggest that we implement CSP to lessen the impact of cross site scripting if there is any.  

CSP:https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP

degeri:
I will be working on this.

PR #519, opened 2018-08-21
Update Dockerfile-serve

degeri:
This implements CSP.
It has the following rules. 
Allows "self" , *.decred.org  and *.dcrdata.org

This fixes https://github.com/decred/politeiagui/issues/506

### DIFF
--- a/Dockerfile-serve
+++ b/Dockerfile-serve
@@ -5,4 +5,4 @@ LABEL version="1.0"
 LABEL maintainer "holdstockjamie@gmail.com"
 
 COPY ./docker-build/ /usr/local/apache2/htdocs/
-RUN sed -i -e "$ a #CSP Headers\nHeader set Content-Security-Policy \"default-src 'self' *.decred.org *.dcrdata.org;\"" /usr/local/apache2/conf/httpd.conf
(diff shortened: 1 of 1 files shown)


jholdstock:
Change looks fine for me. Just a couple of things to consider:
- Are we adding  `*.decred.org` and `*.dcrdata.org` because they are required right now? Or because we _may_ require them in future? If the latter, I would suggest they are added when we actually need them, rather than adding them pre-emptively.
- If I recall the production setup correctly, there is an nginx instance reverse proxying to multiple docker containers. Perhaps this kind of protection is actually better suited to live in the nginx instance? In fact, perhaps it is already there?

degeri:
1. Yes we need them now. We are loading content for the following.
https://faucet.decred.org/requestfaucet request is made for purchasing credits in testnet and requests to *.dcrdata.org (testnet and explorer) are made for payment confirmation. We can go for a more granular rule like so (Content-Security-Policy  "default-src 'self' testnet.dcrdata.org explorer.dcrdata.org testnet.decred.org mainnet.decred.org faucet.decred.org;) but I wanted to be safe and make sure nothing breaks.


2. Yes I believe that other HTTP security headers come from nginx (X frame options,HSTS and XSS protection) but no CSP.

image

fernandoabolafio:
@dajohi can you take a quick look at this

PR #525, opened 2018-08-23
Refactor Component Lifecycle Code

tiagoalvesdulce:
`componentWillReceiveProps` and `componentWillMount` lifecycles are considered legacy and the  [React docs](https://reactjs.org/docs/react-component.html) advises to avoid using them. Soon they will be completely discontinued. I've refactored the lifecycle of our components that used them using `componentDidUpdate` and the `constructor` instead.


fernandoabolafio:
LGTM but I am waiting on #515 before getting it in since there will be the need for a bit more of lifecycle changes after that.

Issue #524, opened 2018-08-22
Issue with last submitted proposal message


tiagoalvesdulce:
<img width="868" alt="screen shot 2018-08-22 at 16 30 38" src="https://user-images.githubusercontent.com/13955303/44486015-d3603380-a628-11e8-870b-00cc33f49623.png">

This message should show up at the after submitted page ONLY. But, if you submit a proposal and navigate through the website without refreshing it and then go back to the proposal's page (by clicking it) the message stays there. We should clear the `lastSubmitted` state on route change.

The message should also be rephrased after https://github.com/decred/politeia/pull/420 goes in saying the user won't be able to get the proposal's content only if it gets censored.

thi4go:
I'll catch this one! 👍

PR #549, opened 2018-08-31
Correcting issues with last submitted proposal message

thi4go:
Added an redux action to clear lastSubmitted when ProposalDetail gets unmounted, correcting it's render behaviour

Fixes https://github.com/decred/politeiagui/issues/524


tiagoalvesdulce:
LGTM

PR #559, opened 2018-09-05
Fix logout when multiple tabs are opened

### LINKED ISSUE #558: Problem logging out when logged in in multiple tabs
Looks like the user info is being erased from the localStorage in which the log out is done but the other tab is rewriting the info into localStorage again.

fernandoabolafio:
closes #558

### DIFF
--- a/src/reducers/api.js
+++ b/src/reducers/api.js
@@ -4,6 +4,7 @@ import map from "lodash/fp/map";
 import cloneDeep from "lodash/cloneDeep";
 import { DEFAULT_REQUEST_STATE, request, receive, reset, resetMultiple } from "./util";
 import { PROPOSAL_VOTING_ACTIVE } from "../constants";
+import { clearStateLocalStorage } from "../lib/local_storage";
 
 export const DEFAULT_STATE = {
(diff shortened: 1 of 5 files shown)


src/reducers/api.js
```diff
@@ -313,6 +314,7 @@ const api = (state = DEFAULT_STATE, action) => (({
   [act.RECEIVE_LOGOUT]: () => {
     const tempState = DEFAULT_STATE;
     tempState.init = state.init;
+    clearStateLocalStorage();
```
tiagoalvesdulce: It's OK to do this here by now, but please write an issue to fix it in the future so our reducers stays as pure functions.

tiagoalvesdulce:
Tested and it works great. As soon as the tests are fixed, it can go in but please note my inline comment.

src/reducers/api.js
fernandoabolafio: Actually, there are multiple places where we need to fix that. Here is the issue:
https://github.com/decred/politeiagui/issues/562

tiagoalvesdulce:
LGTM

Issue #376, opened 2018-06-15
Testing - next steps


fernandoabolafio:
- Write unit testing for selectors and reducers
- Write component testing for key components to make sure they behave correctly.
 Examples of components to be tested are:
   - Header alert: make sure it will be rendered when either paywall is missing or/and the user needs to generate a new identity.
   - ThingLink: make sure it renders the correct content, given different situations (censored proposal, just created proposal, proposal on voting, proposal with finished voting, etc)
   - Buttons triggering key actions (submit proposal, submit comment, approve/censor proposal) make sure they are all blocked when there is a missing verification (paywall/identity).

thi4go:
I can write some more tests, I'll work on this @fernandoabolafio 👍

PR #577, opened 2018-09-08
Test coverage for redux selectors module

thi4go:
Straightforward unit tests for the selectors, focusing on conditional returns behaving as expected with state changes and such. Covers all modules with threshold of 80%.

Adds a mocked state used to test our selectors, which with some little refactor can be used as a complete reference lookup to the state structure used by the gui, and used for something.

image

Part of https://github.com/decred/politeiagui/issues/376


tiagoalvesdulce:
LGTM

Issue #562, opened 2018-09-05
Reducers must remain as pure functions


fernandoabolafio:
Audit app reducers and spot places where side effects are introduced. All reducers must be pure functions.

PR #581, opened 2018-09-10
Reducers to pure functions

thi4go:
Audit reducers to be pure functions only, and moved logic like assignments and declarations, iterations, to action/lib.

Only two reducers from api module needed a rewrite.

closes https://github.com/decred/politeiagui/issues/562

### DIFF
--- a/src/lib/api.js
+++ b/src/lib/api.js
@@ -228,3 +228,33 @@ export const usernamesById = (csrf, userids) => {
 export const proposalsVoteStatus = () => GET("/v1/proposals/votestatus").then(getResponse);
 export const proposalVoteStatus = (token) => GET(`/v1/proposals/${token}/votestatus`).then(getResponse);
 
+
+export const reduceCommentLikes = (commentsvotes, commentid, token, newAction) => {
+  let reducedVotes = null;
+  const cvfound = commentsvotes && commentsvotes.find(
+    cv => cv.commentid === commentid && cv.token === token
+  );
+
+  if (cvfound) {
+    reducedVotes = commentsvotes.reduce(
+      (acc, cv) => {
+        if (cv.commentid === commentid && cv.token === token) {
+          const currentAction = parseInt(cv.action, 10);
+          acc.oldAction = currentAction;
+          cv = {
+            ...cv,
+            action: newAction === currentAction ? 0 : newAction
+          };
+        }
+        return { ...acc, cvs: acc.cvs.concat([cv]) };
+      }, { cvs: [], oldAction: null });
+  } else {
+    const newCommentVote = { token, commentid, action: newAction };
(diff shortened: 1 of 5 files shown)


src/reducers/api.js
```diff
@@ -125,43 +113,8 @@ export const onResetSyncLikeComment = (state) => {
 };
 
 export const onReceiveSyncLikeComment = (state, action) => {
```
fernandoabolafio: Why did you change this function?

src/lib/api.js
```diff
@@ -228,3 +228,33 @@ export const usernamesById = (csrf, userids) => {
 export const proposalsVoteStatus = () => GET("/v1/proposals/votestatus").then(getResponse);
 export const proposalVoteStatus = (token) => GET(`/v1/proposals/${token}/votestatus`).then(getResponse);
 
+
+export const reduceCommentLikes = (commentsvotes, commentid, token, newAction) => {
```
fernandoabolafio: I think it shouldn't be here

src/reducers/api.js
thi4go: It had some side effects, like the usage of cloneDeep, some assignments and iterations. Thought It would be better to move everything to the action dispatching it, build the data needed there, and pass down to the reducer.  The actions module where quite clean as well, mostly handling dispatches and callbacks. So I thought that it would be more clean to handle that logic elsewhere, and putted it on lib. 
This is the code that was removed from onReceiveSyncLikeComment

```
const commentsvotes = state.commentsvotes.response &&
    state.commentsvotes.response.commentsvotes;
const backupCV = cloneDeep(commentsvotes);
const comments = state.proposalComments.response &&
    state.proposalComments.response.comments;

  let reducedVotes = null;
  const cvfound = commentsvotes && commentsvotes.find(
    cv => cv.commentid === commentid && cv.token === token
  );

  if (cvfound) {
    reducedVotes = commentsvotes.reduce(
      (acc, cv) => {
        if (cv.commentid === commentid && cv.token === token) {
          const currentAction = parseInt(cv.action, 10);
          acc.oldAction = currentAction;
          cv = {
            ...cv,
            action: newAction === currentAction ? 0 : newAction
          };
        }
        return { ...acc, cvs: acc.cvs.concat([cv]) };
      }, { cvs: [], oldAction: null });
  } else {
    const newCommentVote = { token, commentid, action: newAction };
    reducedVotes = {
      cvs: commentsvotes ? commentsvotes.concat([newCommentVote]) : [newCommentVote],
      oldAction: 0
    };
  }

  const { cvs: newCommentsVotes, oldAction } = reducedVotes;```

thi4go:
Correcting the tests @fernandoabolafio ! Thanks for the review 👍

PR #590, opened 2018-09-12
Add max height to comments

### LINKED ISSUE #567: Add max-height to comments
If some user creates a very long comment, the scroll gets really painful. We should set a max-height to comments and a Read more at the end to show the entire comment.

largecomment

camus-code:
Resolves #567

### DIFF
--- a/src/components/snew/ThingComment.js
+++ b/src/components/snew/ThingComment.js
@@ -15,6 +15,27 @@ class ThingComment extends React.PureComponent {
     e && e.preventDefault && e.preventDefault();
     this.props.onCensorComment(this.props.loggedInAsEmail, this.props.token, this.props.id);
   }
+  handleCommentMaxHeight = () => {
+    const insertAfter = (newNode, referenceNode) => referenceNode.parentNode.insertBefore(newNode, referenceNode.nextSibling);
+    const commentsDiv = Array.prototype.slice.call(document.querySelectorAll(".comment > .entry > .usertext"));
+    const overflowDivs = commentsDiv.filter(function isOverflown(element) {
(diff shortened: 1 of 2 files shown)


dajohi:
@camus-code rebase needed.

src/components/snew/ThingComment.js
```diff
@@ -15,6 +15,24 @@ class ThingComment extends React.PureComponent {
     e && e.preventDefault && e.preventDefault();
     this.props.onCensorComment(this.props.loggedInAsEmail, this.props.token, this.props.id);
   }
+  componentDidMount(){
```
fernandoabolafio: Could you extract and encapsulate this logic into another function? 
something like: 
```js
handleCommentMaxHeight = () => {
//logic goes here
}
componentDidMount() {
  this.handleCommentMaxHeight()
}
```

src/components/snew/ThingComment.js
```diff
@@ -15,6 +15,24 @@ class ThingComment extends React.PureComponent {
...
+    const insertAfter = (newNode, referenceNode) => referenceNode.parentNode.insertBefore(newNode, referenceNode.nextSibling);
+    const commentsDiv = Array.prototype.slice.call(document.querySelectorAll(".comment > .entry > .usertext"));
+    const overflowDivs = commentsDiv.filter(function isOverflown(element) {
+      return element.scrollHeight > element.clientHeight || element.scrollWidth > element.clientWidth;
+    });
+    for (let i = 0; i < overflowDivs.length; i++) {
```
fernandoabolafio: A suggestion: here you could use the `forEach` method instead, for easier readability:
```js
overflowDivs.forEach((overflowDiv, index) => {
// logic goes here
})
```

camus-code:
Everything should be up to par now

PR #613, opened 2018-09-20
Update flow for onboarding

### LINKED ISSUE #607: Improve onboarding flow
We want to show the user a "Welcome! Not sure what Politeia is? Want to read more about it?" message (or something like this) and two options "Sure, show me more" and "No, I know how Politeia works" and then follow to the onboarding text. This can be implemented as a "first step" in the onboarding modal to make the implementation easier. We should also reword the "Read more" link in the sidebar that opens the onboarding modal to something like "Open onboarding modal to read more" or something similar. Note that when clicking on this link in the sidebar, the "first step" saying Welcome ... should not be presented.

camus-code:
resolves #607


tiagoalvesdulce:
LGTM

PR #614, opened 2018-09-20
Improve proposal credits management flow

### LINKED ISSUE #606: Improve proposal credits buying flow
Right now, to buy proposal credits the user has to go to the `Account` page and there is where the user manages identity, username, and password. Proposals Credits don't fit well there. We want to create a new proposal credits buying flow similar to what @fernandoabolafio did on https://github.com/decred/politeiagui/pull/605 for the paywall.

**Edit:** A good place to put a "Buy proposal credits" button is where the refresh button is. After https://github.com/decred/politeiagui/pull/598 we will not need this refresh anymore so we could use that space.

tiagoalvesdulce:
Closes #606


tiagoalvesdulce:
@thi4go Thanks! I agree we should paginate this table. I think it can be done on the frontend because in the real world people wouldn't be buying thousands of credits.

PR #621, opened 2018-09-21
Polishes copy for onboarding modal

s-ben:
Was checking out the latest politeiagui and couldn't resist polishing some copy on the onboarding modal. Mainly just rewriting for flow. Feel free to close if changes are off message or review would slow down launch.

### DIFF
--- a/src/components/Modal/contents/OnBoard.js
+++ b/src/components/Modal/contents/OnBoard.js
@@ -86,12 +86,10 @@ const onBoardingContent = [
     body: `
 **What is Politeia?**
 
-Politeia is a platform which is being developed to support Decred’s governance.
-It facilitates the submission, tracking, and discussion of Decred governance proposals by the stakeholder community.
+Politeia is a platform that supports Decred’s governance model. It facilitates the submission, tracking, and discussion of Decred governance proposals by the stakeholder community.
 
-Decred’s approach to decision-making is grounded in the principle of stakeholder voting.
-Proof of Stake Voters lock DCR to buy tickets, while tickets are live they can be used to cast a vote for or against
-each open Politeia proposal.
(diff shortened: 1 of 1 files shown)


src/components/Modal/contents/OnBoard.js
```diff
@@ -86,28 +86,23 @@ const onBoardingContent = [
     body: `
 **What is Politeia?**
 
-Politeia is a platform which is being developed to support Decred’s governance.
-It facilitates the submission, tracking, and discussion of Decred governance proposals by the stakeholder community.
+Politeia is a platform that supports Decred’s governance model. It facilitates the submission, tracking, and discussion of Decred governance proposals by the stakeholder community.
```
RichardRed0x: Don't agree with adding "model", makes it sound like it's supporting some concept.

src/components/Modal/contents/OnBoard.js
```diff
@@ -86,28 +86,23 @@ const onBoardingContent = [
...
+
+Decred’s approach to decision-making is grounded in the principle of stakeholder voting. To vote, stakeholders must buy “tickets” that lock their DCR until they are called to vote. Stakeholders can vote on any open proposal.
```
RichardRed0x: tickets "that lock their DCR" doesn't sound right, the ticket doesn't lock anything, the locking is done to obtain the ticket.

RichardRed0x: Would be better to get rid of "buy" altogether.

> To vote, stakeholders must time-lock DCR to buy tickets. While tickets are live they can be used to cast a vote for or against each open Politeia proposal.

s-ben:
@RichardRed0x , agree with both of your suggested changes. Have implemented and pushed to branch.

PR #626, opened 2018-09-25
Display indicator when session is about to expire

### LINKED ISSUE #623: [logout] Add visual indicator that session has ended
After a session expires the GUI does not provide feedback and it only becomes obvious after reloading the site. We need some sort of indicator that the session is about to expire or has expired.

fernandoabolafio:
closes #623 
Depends on https://github.com/decred/politeia/pull/482
This PR adds an indicator when the session is about to expire (less than 10 minutes left). Once the Session is ended, the user is explicitly logged out and a modal for log in shows up on the screen.

**Indicator is active:**
screen shot 2018-09-25 at 12 07 54 pm

**Session has ended:**
screen shot 2018-09-25 at 12 10 42 pm

### DIFF
--- a/src/components/SessionExpiresIndicator.js
+++ b/src/components/SessionExpiresIndicator.js
@@ -1,22 +1,16 @@
 import React from "react";
 import distanceInWordsToNow from "date-fns/distance_in_words_to_now";
+import IntervalComponent from "./IntervalComponent";
 import currentUserConnector from "../connectors/currentUser";
 import * as modalTypes from "./Modal/modalTypes";
 
 class SessionExpiresIndicator extends React.Component {
-  interval = null
   constructor(props) {
     super(props);
     this.state = {
-      timer: null
+      timer: null,
+      activateInterval: false
     };
-  }
-  finishInterval = () => {
-    this.props.onLogout();
-    this.setState({ timer: null });
-    this.props.openModal(modalTypes.LOGIN, {
-      title: "Your session has expired. Please log in again."
-    }, null);
   }
   intervalProcedure = () => {
     let { timer } = this.state;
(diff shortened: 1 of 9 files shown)


src/components/SessionExpiresIndicator.js
```diff
@@ -0,0 +1,70 @@
+import React from "react";
```
brunobraga95: This is a very handy component, in fact it is so good that I suggest you to change your approach and decouple it from the name `SessionExpiresIndicator` so that it can be used under different contexts. i.e A component should appear (and maybe disappear as well) after X seconds. I suggest something like the following:
```
<Router>
   <LoaderComponent>	         
     <LoaderComponent>
     <SessionExpiresIndicator />
      <HeaderAlertComponent />
                 ......
```

```
const SessionExpiresIndicator = ({openModa, onLogout}) => 
    (<TimedComponent appearsAfter={10} onTimePassed={() => onLogout()}>
      {openModal(modalTypes.LOGIN, {
      title: "Your session has expired. Please log in again."
    }, null)}>)
```

and then your component (which I named TimedComponent) can render its children on your `finishInterval` method

src/components/SessionExpiresIndicator.js
```diff
@@ -0,0 +1,70 @@
...
+import distanceInWordsToNow from "date-fns/distance_in_words_to_now";
+import currentUserConnector from "../connectors/currentUser";
+import * as modalTypes from "./Modal/modalTypes";
+
+class SessionExpiresIndicator extends React.Component {
```
brunobraga95: well done.

src/components/SessionExpiresIndicator.js
fernandoabolafio: hey @brunobraga95, thanks for your review! I actually like this idea. The example you gave me isn't exactly applicable to this case since I'm implementing an interval which should be up given a specific circumstance. However, it sparked me that I could split the interval logic into another component, which is very close to what you suggested. I've done the changes already, in case you want to take a look.

PR #634, opened 2018-09-27
Always use rel="noopener noreferrer" with target="_blank"

dajohi:


degeri:
https://github.com/decred/politeiagui/issues/635

src/components/snew/Markdown/helpers.js
```diff
@@ -86,15 +86,15 @@ export const customRenderers = (filterXss, confirmWithModal) => ({
   image: ({ src, alt }) => {
     return <a
       target="_blank"
-      rel="nofollow"
+      rel="nofollow noopener noreferrer"
       onClick={(e) => confirmWithModal && verifyExternalLink(e, src, confirmWithModal)}
```
fernandoabolafio: In order to have the correct behavior for the markdown links you need to change the javascript code inside the `verifyExternalLink` function because external links are openned programmtically. I've done a few tests and this code seems to implement the expected behavior:
```
const verifyExternalLink = (e, link, confirmWithModal) => {
  e.preventDefault();
[...]
```
```
}).then(confirm => {
      if (confirm) {
        const newWindow = window.open();
        newWindow.opener = null;
        newWindow.location = link;
        newWindow.target = "_blank";
      }
    });
[...]
```

PR #638, opened 2018-09-28
Add Pagination

### LINKED ISSUE #552: Implement Proposals Pagination
Proposals are paginated on the backend, the GUI should handle it.

tiagoalvesdulce:
Closes #552 

Depends on https://github.com/decred/politeia/pull/465


thi4go:
Tested and pagination is working great, on every proposal listing screen. I think that paginating from 10 proposals instead of 20 would make the listing a bit more clean at first moment, but that would be my only suggestion.

LGTM.

src/selectors/api.js
```diff
@@ -253,6 +255,27 @@ export const getPropVoteStatus = state => token => {
 export const userid = state => state.api.me.response && state.api.me.response.userid;
 export const censoredComment = state => state.api.censorComment.response;
 
+
+export const lastLoadedUnvettedProposal = state => {
```
brunobraga95: expect for the first line of the functions below, the rest is same, it could then be extracted to its own method and avoid replicated code, in fact maybe even the first line could be avoided.

```
const lastValue = (proposalMethod, state) => {
  const vp = proposalMethod(state);
  const last = Object.keys(vp).length - 1;
  return last > -1 ? vp[last] : {}
}
...
...
export const lastLoadedVettedProposal = state => lastValue(apiVettedProposals, state);
export const apiUnvettedProposals = state => lastValue(apiUnvettedProposals, state);
export const lastLoadedUserDetailProposal = state => lastValue((state) => user(state) ? user(state).proposals : [] , state);
export const lastLoadedUserProposal = state => lastValue(apiUserProposals, state);
```

src/components/UserDetail/ProposalsTab.js
```diff
@@ -2,17 +2,45 @@ import React from "react";
 import userConnector from "../../connectors/user";
 import { Content as ProposalListing } from "../../components/snew";
 
+const removeDuplicates = (arr1, arr2) => {
+  const proposalsObj = arr1 ? arr1.reduce((acc, cur) => {
```
brunobraga95: I think you are doing the same thing on both arrays, could extract this to a single method and make it more readable.

src/components/snew/Content.js
```diff
@@ -42,18 +47,39 @@ export const CustomContent = ({
...
-            key: "content",
-            lastBlockHeight: props.lastBlockHeight,
-            listings: listings || [
-              {
-                allChildren:
-                proposals.map((proposal, idx) => formatProposalData(proposal, idx, activeVotes))
-              }
-            ]
-          }} />
+        (listings && listings.length > 0) || proposals.length > 0 || proposalCounts[filterValue] !== 0 ? (
```
brunobraga95: in my opinion it is always good to avoid too much condition like this and extract it to a more descriptive name.

src/components/snew/Content.js
```diff
@@ -42,18 +47,39 @@ export const CustomContent = ({
...
+                <button
+                  style={{ marginTop: "15px" }}
+                  className="c-btn c-btn-primary"
+                  onClick={() => onFetchData ?
+                    onFetchData(lastLoadedProposal ? lastLoadedProposal.censorshiprecord.token : null)
+                    :
+                    onFetchUserProposals(userid, lastLoadedProposal ? lastLoadedProposal.censorshiprecord.token : null)
+                  }
+                  disabled={
+                    (count && count <= proposals.length) ||
```
brunobraga95: same as above

src/helpers.js
```diff
@@ -153,3 +159,24 @@ export const multiplyFloatingNumbers = (num1, num2) => {
   }
   return (num1*num2)/(Math.pow(10, cont1+cont2));
 };
+
+export const countPublicProposals = (proposals) => {
```
brunobraga95: :+1:

src/reducers/api.js
```diff
@@ -32,10 +32,56 @@ export const DEFAULT_STATE = {
...
   verifyUserKey: DEFAULT_REQUEST_STATE,
   likeComment: DEFAULT_REQUEST_STATE,
+  unvettedStatus: DEFAULT_REQUEST_STATE,
   email: "",
   keyMismatch: false
 };
 
+export const onReceiveProposals = (key, state, { payload, error }) => {
+  if (state[key] && state[key].response && state[key].response.proposals) {
+    const proposalsObj = state[key].response.proposals.reduce((acc, cur) => {
```
brunobraga95: seems to be the same method you used in a previous case, maybe it could be extracted to a helper function in order to centralize the logic (this way you reduce the code and who is reading just need to understand one function that is applied in other places)

PR #640, opened 2018-10-01
Replace buttons loading icon

### LINKED ISSUE #599: Replace buttons loading icon
Replace the loading icon on buttons as soon as we get a negative version of [politeia loading gif](https://github.com/decred/politeiagui/blob/master/src/style/politeia-loading.gif).

victorgcramos:
Fixes #599 
- Added new loading icon files
- Removed "spin" attribute, since the icon is already animated
- Changed the files on CSS to get the icons from the new files


victorgcramos:
It's not possible to opacify the background image of our assets, so I created a new element that is only shown when user hovers the button. That element has an opacity closer to our button's. This is not 100% accurate, but it's the closest possible.

tiagoalvesdulce:
Good enough.

PR #642, opened 2018-10-01
Malformed links redirecting to proposals/:id page

### LINKED ISSUE #633: Malformed links redirecting to proposals/:id page
Links without `http://` or `https://` are being treated as a politeiagui route.

Go to https://test-proposals.decred.org/proposals/0873814acca0322986c94dd7408c54dbab17815bbed64b938780514a811ee2f9 and click the first comment to reproduce.

victorgcramos:
Fixes #633

### DIFF
--- a/src/components/snew/Markdown/helpers.js
+++ b/src/components/snew/Markdown/helpers.js
@@ -3,13 +3,16 @@ import xssFilters from "xss-filters";
 import * as modalTypes from "../../Modal/modalTypes";
 
 function extractHostname(url) {
+  console.log("HOSTNAME 1", url, typeof url);
   //find & remove protocol (http, ftp, etc.) and get hostname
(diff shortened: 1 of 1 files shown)


src/components/snew/Markdown/helpers.js
```diff
@@ -2,6 +2,16 @@ import React from "react";
 import xssFilters from "xss-filters";
 import * as modalTypes from "../../Modal/modalTypes";
 
+function extractHostname(url) {
```
dajohi: Can this be simplified to https://nodejs.org/api/url.html#url_url_hostname?

tiagoalvesdulce:
`new URL()` doesn't work properly if the provided URL is not prefixed with `http` or `https`, which is the exact use case @victorgcramos is handling. So, his function looks good.

PR #644, opened 2018-10-01
Fix proposal credits inconsistencies

### LINKED ISSUE #639: Weird behaviours on proposal credit's number display
Some unwanted behaviours on the credit number display. To note, when we refresh/reload the pgui, it gives us the correct number of credits, but continuing interactions with pgui without reloading it leads to these bugs.

This one happens right after a transaction to buy more credits has been confirmed.
pcredit

And this one right after submitting a new proposal. 
pcredit2

tiagoalvesdulce:
Closes #639


thi4go:
Tested and working great! Every interaction with proposal credits is now solid. Thanks @tiagoalvesdulce. 

LGTM.

PR #652, opened 2018-10-03
Make proposal credits status more explicit

### LINKED ISSUE #646: [Proposal Credits] Explicit display what the purchase status is
Instead of displaying a loading spinner for unconfirmed transactions or a check for the confirmed ones, I think it would be better to explicit show what the status is. Spinners usually indicate that there is an ongoing request waiting for the reply.
Maybe also consider displaying the confirmations number (e.g.: 2/3)

**Example:**
screen shot 2018-10-02 at 9 08 39 am

camus-code:
closes #646

### DIFF
--- a/src/components/ProposalCreditsManager/Summary.js
+++ b/src/components/ProposalCreditsManager/Summary.js
@@ -1,4 +1,5 @@
 import React from "react";
+import { CONFIRMATIONS_REQUIRED } from "../../constants";
 
 const DcrdataTxLink = ({
   isTestnet,
@@ -17,7 +18,8 @@ const formatDate = (date) => {
   const day = d.getUTCDate();
   const year = d.getUTCFullYear();
   const month = d.getUTCMonth();
-  const minutes = d.getUTCMinutes() < 10 ? `0${d.getUTCMinutes()}` : d.getUTCMinutes();
+  const _minutes = d.getUTCMinutes();
+  const minutes = _minutes < 10 ? `0${_minutes}` : _minutes;
   const time = d.getUTCHours() + ":" + minutes;
   return year + "-" + month + "-" + day + "  |  " + time;
 };
@@ -47,6 +49,7 @@ const ProposalCreditsSummary = ({
       numberPurchased: Math.round(lastPaymentNotConfirmed.amount * 1/proposalCreditPrice),
       txId: lastPaymentNotConfirmed.txid,
       price: proposalCreditPrice,
+      confirmations: lastPaymentNotConfirmed.confirmations,
       confirming: true
     };
     proposalCreditPurchases.push(transaction);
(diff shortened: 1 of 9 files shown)


src/components/ProposalCreditsManager/Summary.js
```diff
@@ -17,8 +17,7 @@ const formatDate = (date) => {
   const day = d.getUTCDate();
   const year = d.getUTCFullYear();
   const month = d.getUTCMonth();
-  const minutes = d.getUTCMinutes() < 10 ? `0${d.getUTCMinutes()}` : d.getUTCMinutes();
```
tiagoalvesdulce: This line was added so we have `20:08` instead of `20:8` for example. Please keep it.

src/components/ProposalCreditsManager/Summary.js
```diff
@@ -27,26 +26,14 @@ const ProposalCreditsSummary = ({
...
-        confirming: false,
-        datePurchased: "just now"
-      };
-      if (!proposalCreditPurchases.find(el => el.txId === transaction.txId)) proposalCreditPurchases.push(transaction);
-    });
-  }
   if (lastPaymentNotConfirmed) {
     const transaction = {
-      numberPurchased: Math.round(lastPaymentNotConfirmed.amount * 1/proposalCreditPrice),
+      numberPurchased: lastPaymentNotConfirmed.amount * 10,
```
tiagoalvesdulce: `proposalCreditPrice` is a constant that will indicate the credit price. It won't always be 0.1, which would need us to go into the code and change it. Let's keep the constant.

src/components/ProposalCreditsManager/Summary.js
brunobraga95: also, you can extract `d.getUTCMinutes()` into a variable and avoid multiple call to the same function.

camus-code: Thanks for the suggestion implementing it now @brunobraga95

src/components/ProposalCreditsManager/Summary.js
```diff
@@ -79,8 +81,9 @@ const ProposalCreditsSummary = ({
                 </div>
                 <div className="credit-purchase-cell credit-purchase-status">
                   { creditPurchase.confirming ?
-                    (<div className="user-proposal-credits-cell"><div className="logo"></div></div>)
-                    : "✔"
+                    (<div className="user-proposal-credits-cell" style={{ color: "#ff8100", fontWeight: "bold" }}><div>
+											Awaiting confirmations: </div>({creditPurchase.confirmations} of 3)</div>)
```
tiagoalvesdulce: This should be 2 confirmations. We have a constant for it here: https://github.com/decred/politeiagui/blob/fcd0b93660b1914ff493f225804ed183f01e5b93/src/actions/external_api.js#L10 Can you please move this constant to `src/constants` and use it in the places of the code that mention the number of confirmations?

camus-code: @tiagoalvesdulce done and done. Lol I thought there was a constant for it but couldn't find it in `src/constants`.

PR #673, opened 2018-10-09
Fix Malformed URLs

### LINKED ISSUE #647: [Malformed URLs] Show error message instead of treating it as an external link
If the user types a malformed URL on the markdown editor, we should show an error message informing him and preventing the submission.

### LINKED ISSUE #654: Malformed links still working on right click and External Link warning bypass
So https://github.com/decred/politeiagui/pull/642 works perfectly when the user gives a "left mouse" click. 

But the URL still seems to point to "/proposals/[inputtext]"  and if a user gives a "right click" and open new tab then it still takes them to the problem URL. 

To reproduce go to URL:
https://test-proposals.decred.org/proposals/45316e5425a426a9a25f4145db92239029d446051dceae3d5ccac8bee1612519

1.

RIGHT click on my comment "test" and open in new tab the URL "https://test-proposals.decred.org/proposals/propppp" 
2.
This might also be a way around the "External Link Warning" that we implemented.  (though we hope anyone right clicking on a link and opening it is smart enough to understand what they are doing)
3.
Also if we create a link with the following code 

POC1:
\![clickmenowarning]\(htt://google.com) 
POC2:
\![ClickMeNoWarning2]\(f:ggggrr) 
POC3:
\![ClickMeNoWarning3]\(bitcoin:notanaddress)  --> this would openup in some clients. 

and give a normal left click it seems to bypass the "Warning message" and directly takes the user to the link

For a permanent fix I would suggest that the URL be "FORMED" only when the user loads the "Modal" object. Thus preventing this issue all together. 

image

victorgcramos:
Fixes #654  
Fixes #647


src/validators/reply.js
```diff
@@ -11,7 +11,10 @@ const validate = ({ values, keyMismatch }, policy) => {
       _error: `The comment must be less than ${policy.maxcommentlength} characters long`
     });
   }
-
+  const validUrl = urlValidator(values.comment);
```
tiagoalvesdulce: This is already used above on `validators/proposal.js`, you should create a function and use it in both places to avoid code duplication.

src/connectors/submitProposal.js
```diff
@@ -32,6 +32,7 @@ class SubmitWrapper extends Component {
   }
 
   render() {
+    console.log("LINKES PEGA", this.props);
```
tiagoalvesdulce: You left a log here.

src/connectors/newProposal.js
```diff
@@ -38,7 +38,9 @@ class NewProposalWrapper extends Component {
   constructor(props){
     super(props);
     this.state = {
-      initialValues: props.draftProposal || getNewProposalData()
+      // initialValues: "props.draftProposal || getNewProposalData()",
```
tiagoalvesdulce: This comment should go away.

src/validators/reply.js
```diff
@@ -11,7 +11,8 @@ const validate = ({ values, keyMismatch }, policy) => {
       _error: `The comment must be less than ${policy.maxcommentlength} characters long`
     });
   }
-  validateURL(urlValidator, values.comment);
+  validateURL(values.comment);
+  console.log("LINKES");
```
tiagoalvesdulce: You left a log here.

degeri:
Looks good :+1:  Great work :)

Issue #666, opened 2018-10-08
Cannot edit recently approved proposal before refresh


tiagoalvesdulce:
<img width="753" alt="screen shot 2018-10-08 at 10 45 38" src="https://user-images.githubusercontent.com/13955303/46612738-ce2b4b00-cae7-11e8-8fa5-7e52bb3cb6d9.png">

To reproduce:
Login as admin > submit a proposal > approve it (you will be redirected to its route) > try to edit it.

thi4go:
Working on it

tiagoalvesdulce:
Closing because it will be solved on https://github.com/decred/politeia/issues/514

tiagoalvesdulce:
Reopening, it should be fixed on GUI.

Issue #665, opened 2018-10-08
Userid on comment's author


tiagoalvesdulce:
<img width="530" alt="screen shot 2018-10-08 at 10 37 35" src="https://user-images.githubusercontent.com/13955303/46612644-873d5580-cae7-11e8-8766-d89a08b31b22.png">

To reproduce:
Login as admin > submit a proposal > approve it (you will be redirected to its route) > comment on it.

camus-code:
Working on it

tiagoalvesdulce:
Closing because it will be solved on https://github.com/decred/politeia/issues/514

tiagoalvesdulce:
Reopening, it should be fixed on the GUI.

PR #679, opened 2018-10-11
Updates proposal state with needed info onReceiveSetStatus

thi4go:
closes https://github.com/decred/politeiagui/issues/666 and https://github.com/decred/politeiagui/issues/665


src/reducers/api.js
```diff
@@ -54,7 +54,14 @@ export const onReceiveSetStatus = (state, action) => {
   if (action.error) return state;
   const getProposalToken = prop => get([ "censorshiprecord", "token" ], prop);
 
-  const { proposal: updatedProposal } = action.payload;
+  const { proposal: updatedProposal } = {
+    proposal: {
+      ...action.payload.proposal,
+      files: state.proposal.response.proposal.files,
```
tiagoalvesdulce: Just one small thing. Let's make sure the `response` object is not null before doing this assignment. `files: state.proposal.response ? state.proposal.response.proposal.files : []` and `username: state.proposal.response ? state.proposal.response.proposal.username : ""`

thi4go: Change commited ! Thanks for the suggestion @tiagoalvesdulce

src/reducers/api.js
```diff
@@ -54,7 +54,14 @@ export const onReceiveSetStatus = (state, action) => {
   if (action.error) return state;
   const getProposalToken = prop => get([ "censorshiprecord", "token" ], prop);
 
-  const { proposal: updatedProposal } = action.payload;
+  const { proposal: updatedProposal } = {
```
fernandoabolafio: You don't need to assign the constant using the constructor since you're describing the object structure.
It can be written straightforward like this: `const updatedProposal = { ... }`.

src/reducers/api.js
```diff
@@ -54,7 +54,14 @@ export const onReceiveSetStatus = (state, action) => {
...
+    proposal: {
+      ...action.payload.proposal,
+      files: state.proposal.response ? state.proposal.response.proposal.files : [],
```
fernandoabolafio: Adding to @tiagoalvesdulce comments, an even safer way of doing this is by using the `get` function from loadash:
```js
[...]
proposal : {
   ...action.payload.proposal,
   files: get(["proposal",  "response", "proposal", "file"], state) || []
[...]
```
The same thing works for the `username`. This is a good practice when you're accessing deeper levels os a nested object. The get won't throw an error in case one of the object keys aren't defined, it will simply return undefined in that case.

thi4go: Done! Thanks for the tips @fernandoabolafio , I'll remember to use `get` for accessing objects from now on :D

Issue #711, opened 2018-10-16
Proposal guidelines should be easily accessible from the Submit page


sndurkin:
I see that there are proposal guidelines on the introduction modal, which is accessible from the "Learn more about Politeia" link. But those guidelines would be better if they were accessible directly from the Submit page.

camus-code:
Mistakenly referenced this issue in my last PR. Picking up this one now though

camus-code:
Actually #726 will indeed close this. Just reissued my PR

PR #726, opened 2018-10-16
Limit access to /proposals/new route

### LINKED ISSUE #710: Account page link doesn't provide the user with more info
Before the user has finished registering, the Submit page shows this error:

image

This is a confusing message, because clicking on the link to open the Account Page doesn't provide a way to buy proposal credits because I haven't paid the registration fee yet.

camus-code:
resolves #710 and #711

### DIFF
--- a/src/components/Modal/contents/OnBoard.js
+++ b/src/components/Modal/contents/OnBoard.js
@@ -8,7 +8,7 @@ class OnBoard extends React.Component {
   constructor(props) {
     super(props);
     this.state = {
-      tab: 0
+      tab: this.props.me.payload && this.props.me.payload.tab ? this.props.me.payload.tab : 0
     };
   }
   render() {
--- a/src/components/Modal/contents/OnBoard.js
+++ b/src/components/Modal/contents/OnBoard.js
@@ -3,12 +3,13 @@ import ModalContentWrapper from "../ModalContentWrapper";
 import modalConnector from "../../../connectors/modal";
 import { Tabs, Tab } from "../../Tabs";
 import Markdown from "../../snew/Markdown";
+import get from "lodash/fp/get";
 
 class OnBoard extends React.Component {
   constructor(props) {
     super(props);
     this.state = {
-      tab: this.props.me.payload && this.props.me.payload.tab ? this.props.me.payload.tab : 0
+      tab: get([ "me", "payload", "tab" ], this.props) || 0
     };
   }
   render() {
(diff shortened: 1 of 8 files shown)


src/connectors/newProposal.js
```diff
@@ -41,6 +46,7 @@ class NewProposalWrapper extends Component {
       initialValues: props.draftProposal || getNewProposalData(),
       validationError: ""
     };
+    this.checkPaywallNewProposal(props);
```
fernandoabolafio: The constructor isn't the right place to call this function because the API may have not returned the paywall data yet. You can test that it is wrong if you have the paywall paid and try to access the route manually; you'll be redirected to the homepage even having the paywall all set.
It should be placed in the `componentDidUpdate` by confirming that the API data is available before calling `this.checkPaywallNewProposal(props);`.

camus-code: Thanks for the suggestion! Just now realizing that it does a redirect to the homepage before it lands on the route. I'll be addressing this and the updated message in the next commit.

camus-code:
@fernandoabolafio I believe I addressed everything we talked about. Let me know what you think

src/connectors/newProposal.js
```diff
@@ -61,6 +70,25 @@ class NewProposalWrapper extends Component {
     }
   }
 
+  checkPaywallNewProposal() {
```
fernandoabolafio: We don't need to be checking the state from local storage here since the Submit route is an authenticated route. All inner components of routes should always check from the in-memory cache (e.g from the store). By doing that we avoid propagating two sources of truth in the application.
Your code can be dramatically simplified by applying the following approach:

- connector:
```js

const newProposalConnector = connect(
  sel.selectorMap({
    ...
    userHasPaid: sel.userHasPaid
  }),
  {
    ...
  }
);
```

- componentDidMount:
```js
componentDidMount(){
    this.checkPaywallNewProposal();  
}
```

- checkPaywallNewProposal:
```js
checkPaywallNewProposal = () => {
    const { location, history, userHasPaid } = this.props;
    if (!userHasPaid) {
      history.replace({
        pathname: "/",
        state: { from: location }
      });
    }
  }
```

src/components/Modal/contents/OnBoard.js
```diff
@@ -8,7 +8,7 @@ class OnBoard extends React.Component {
   constructor(props) {
     super(props);
     this.state = {
-      tab: 0
+      tab: this.props.me.payload && this.props.me.payload.tab ? this.props.me.payload.tab : 0
```
fernandoabolafio: This is more a suggestion for the code style, you can make this sentence more concise by importing the `get` function from `lodash` and write:
```js
this.state = {
    tab: get(["me", "payload", "tab"], this.props) || 0
}
```

PR #750, opened 2018-10-19
Bundle together account page with user/id page

### LINKED ISSUE #740: [UI/UX] Unable to get to user page from account page and mismatched info
There is information on the user page that is not on the account page, such as whether the email address has been verified, whether the registration fee has been paid, and if so, the transaction used to pay it.

Unfortunately, the user page is also fairly well hidden in that the account page has no link to get there (although it does display your user ID).  As a result, in order to get to the user page, you either need to have to performed some action such as commenting on a proposal in order to get a link to it, or simply have to know that URL and manually put your user ID in there.

A relatively quick change that would at least provide a means to get to the user page from the account page would be to make the displayed user ID on the account page a link to the user page.

However, I think, ultimately, the pages probably need to be combined.

camus-code:
Closes #740 by bundling the account page functionality into the user/id route.


camus-code:
Thanks for the suggestions @thi4go ! I'm wrapping this up tonight so it should be good to go by the morning.

1. I made it so that the account identity error message is displayed right above the buttons. In addition, if this error message is active, the "manage identity" section will start out as open (as to better signal where to solve this issue)

2. Taken care of as well. I left the functionality as is (with the verification route)

3. Taken care of (and for the username modal as well).

src/components/UserBadge.js
```diff
@@ -0,0 +1,58 @@
...
+import userBadgeConnector from "../connectors/userBadge";
+import {
+  PAYWALL_STATUS_PAID, PAYWALL_STATUS_LACKING_CONFIRMATIONS, PAYWALL_STATUS_WAITING
+} from "../constants";
+
+const UserBadge = ({
+  userPaywallStatus,
+  loggedInAsEmail
+}) => {
+  return !loggedInAsEmail ? null : userPaywallStatus === PAYWALL_STATUS_WAITING ? (
```
brunobraga95: consider extracting this to a separate variable to improve readability.
```
const isWaiting = loggedInAsEmail && userPaywallStatus === PAYWALL_STATUS_WAITING;
return isWaiting ? (
....
) : null

```

fernandoabolafio:
This PR was moved to https://github.com/decred/politeiagui/pull/761 due to some strange behavior from Github on this thread.

PR #758, opened 2018-10-22
Add admin/users

tiagoalvesdulce:
**To test it, make sure to run `yarn` before `yarn start` to upgrade the dependencies.**

Depends on https://github.com/decred/snew-classic-ui/pull/12


src/components/UserLookupPage.js
```diff
@@ -0,0 +1,85 @@
+import React from "react";
+import LoadingIcon from "./snew/LoadingIcon";
+import { Field, reduxForm } from "redux-form";
+import Link from "./snew/Link";
+import Message from "./Message";
+import searchUserConnector from "../connectors/userSearch";
+
+class UserLookupPage extends React.Component {
```
brunobraga95: Prefer `React.PureComponent`
https://stackoverflow.com/questions/41340697/react-component-vs-react-purecomponent

src/components/UserLookupPage.js
```diff
@@ -0,0 +1,85 @@
...
+            </div>
+          }
+        </form>
+        {error && (
+          <Message type="error" header="Cannot search for user" body={error} />
+        )}
+        {userSearch && (
+          <div className="user-search-results">
+            <h3>Matches {userSearch.totalmatches} of {userSearch.totalusers} users</h3>
+            <table>
```
brunobraga95: is it necessary to use tables? just thinking..it has been a while since I dont see it.

probably makes no difference tho

tiagoalvesdulce: Yeah, we probably want to create a Table component to unify the styles of proposal credits table, markdown help table, and user search results table in the future.

tiagoalvesdulce: here, created an issue for that: https://github.com/decred/politeiagui/issues/760

brunobraga95: fair.

tiagoalvesdulce:
screen shot 2018-10-22 at 17 59 07

screen shot 2018-10-22 at 17 58 56

screen shot 2018-10-22 at 17 58 16

screen shot 2018-10-22 at 17 58 32

PR #764, opened 2018-10-23
Adds registration fee to history; minor changes to table ui

### LINKED ISSUE #716: User registration payment not showing in payment history.
A user registration payment will not appear in the payment history even after it has received the required number of confirmations.  A refresh fixes it.

screen shot 2018-10-16 at 8 53 08 am

thi4go:
Now user can see his tx from register fee on the history table.

Minor changes do UI, wanted to see how y'all feel about it. One thing I see as a bottleneck on our current UI is that we don't have a very harmonious colour palette. So I changed some things on the table to make it more neutral.

acchist

from: 

image

closes #716

### DIFF
--- a/src/components/ProposalCreditsManager/Page.js
+++ b/src/components/ProposalCreditsManager/Page.js
@@ -11,13 +11,22 @@ class ProposalCreditsPage extends React.Component {
       proposalCreditPrice,
       isApiRequestingProposalPaywall,
       proposalCredits,
-      proposalCreditPurchases,
 
       // Testnet only
       isTestnet,
       proposalPaywallPaymentTxid,
       ...props
     } = this.props;
+
+    const proposalCreditPurchases = this.props.proposalCreditPurchases.unshift({
+      numberPurchased: 1,
+      price: 0.1,
+      confirming: false,
+      amount: 1,
+      type: "fee",
+      txId: this.props.paywallTxid
+    });
+
     return isApiRequestingProposalPaywall ?
       <PageLoadingIcon />
       : (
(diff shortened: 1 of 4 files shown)


src/components/ProposalCreditsManager/Page.js
```diff
@@ -11,13 +11,22 @@ class ProposalCreditsPage extends React.Component {
...
 
       // Testnet only
       isTestnet,
       proposalPaywallPaymentTxid,
       ...props
     } = this.props;
+
+    const proposalCreditPurchases = this.props.proposalCreditPurchases.unshift({
+      numberPurchased: 1,
+      price: 0.1,
```
tiagoalvesdulce: We shouldn't hardcode this kind of stuff because if it changes, it will require changes all over the code. Let's use the `paywallAmount`: https://github.com/decred/politeiagui/blob/caf27a0859b1a7dce7766e4cf49789525afc8120/src/selectors/api.js#L221

thi4go: Thanks for the review @tiagoalvesdulce. Added a constant to represent that value.

thi4go:
Thanks for the review @tiagoalvesdulce ! Changed the spinner to a confirming text, but I really think we should use neutral colours to represent the transaction state. If you guys don't like the check sign, I can change it to a neutral coloured Confirmed text.

What are your thoughts on this, @fernandoabolafio ?

screen shot 2018-10-23 at 11 05 04

thi4go:
All changes done guys, think its good to go now.

PR #778, opened 2018-10-23
[proposal credits] Integrate admin capability of rescan user payments

fernandoabolafio:
This PR is the corresponding GUI work for https://github.com/decred/politeia/pull/548.

**Note**: I've moved the custom action handlers from the API reducer to its own file for better readability.

Preview:
screen shot 2018-10-23 at 12 07 14 pm

screen shot 2018-10-23 at 12 07 24 pm


src/components/UserDetail/GeneralTab.js
```diff
@@ -196,15 +196,47 @@ class GeneralTab extends React.Component {
...
+                text="rescan"
+              /> : null}
+          </Field>
+          {hasTheRescanResult ?
+            <Message
+              type="success"
+              body={<div>
+                {amountOfCreditsAddedOnRescan === 0 ?
+                  <span>User credits are up to date.</span> :
+                  <span><b>{amountOfCreditsAddedOnRescan} proposal credits </b>we're generated by the rescan.</span>
```
tiagoalvesdulce: Here should be `were`.

PR #796, opened 2018-10-25
Shows remaining blocks for voting expiration with some ui changes to voting div

### LINKED ISSUE #668: [vote] Add blocks left in vote period indicator
The GUI provides no feedback as to how many blocks/duration there is left on voting. We need to add some sort of counter as to ho many blocks there are left in the voting period.

thi4go:
closes #668 

prev:

image

now:

image

Any UI suggestions can be changed and commited quickly, so suggest away! 👍

### DIFF
--- a/src/components/VoteStats.js
+++ b/src/components/VoteStats.js
@@ -3,6 +3,7 @@ import voteStatsConnector from "../connectors/voteStats";
 import StackedBarChart from "./StackedBarChart";
 import distanceInWordsToNow from "date-fns/distance_in_words_to_now";
 import { getRandomColor } from "../helpers";
+import Tooltip from "./Tooltip";
 import {
   PROPOSAL_VOTING_ACTIVE,
   PROPOSAL_VOTING_FINISHED,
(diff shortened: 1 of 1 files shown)


thi4go:
Hey @fernandoabolafio thanks for the review. really liked the suggestion, working on it and will be commiting soon.

thi4go:
@fernandoabolafio messed with illustrator a bit (and was thrilled to discover dcrdesign repo on github hah), so came up with a proposal icon that I think will be useful for a visual on it's status. Also corrected alignment and some other minor things. It's really looking better now, I believe.

I changed the proposal finished color to our default dark blue, because red was giving out the impression of an error of some kind.

Also thought about putting the status text color a default black and semibold maybe, instead of the status color, but wasnt sure about it... if you could pitch in on this as well would be great.

Almost done with the PR, but it would be nice one last test on somebody's local dev environment to see proposals on all states. 

Attatching some images of how the div is looking like:

image

image


I think that we could better our onboarding in the near future using the icons, to demonstrate visually to the user all the states a proposal passes, and it's transition steps.

PR #799, opened 2018-10-26
Get correct URL for dcrdata and insight based on the testnet flag

fernandoabolafio:


src/components/VoteStats.js
```diff
@@ -111,7 +110,8 @@ class Stats extends React.Component {
     }))
 
   getTimeInBlocks = (blocks) => {
-    const blockTimeMinutes = NETWORK === "testnet" ? blocks*2 : blocks*5 ;
+    console.log("is testnet", this.props.isTestnet);
```
tiagoalvesdulce: You left a log here

PR #814, opened 2018-10-30
Validates UUID on user detail route

### LINKED ISSUE #798: [User details] Invalid ID leads to unhandled error
Access the User Details page with some invalid Id like:
`https://<hostname>/user/thisIsAnInvalidId`

While it should be handled gracefully and display a proper error, it is displaying a meaningless error:

screen shot 2018-10-26 at 12 06 18 pm

Ideally, I think we should have a regex to match the ID in the URL and if the request fails, display the error properly.

thi4go:
closes #798

### DIFF
--- a/src/actions/api.js
+++ b/src/actions/api.js
@@ -239,6 +239,10 @@ export const onFetchProposal = token => dispatch => {
 
 export const onFetchUser = userId => dispatch => {
   dispatch(act.REQUEST_USER(userId));
+  const regexp = /^[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$/i;
+  const valid = regexp.test(userId);
+  if (!valid)
+    return dispatch(act.RECEIVE_USER(null, "This is not a valid user ID."));
   return api
     .user(userId)
     .then(response => dispatch(act.RECEIVE_USER(response)))
--- a/src/actions/tests/api.test.js
+++ b/src/actions/tests/api.test.js
@@ -871,14 +871,14 @@ describe("test api actions (actions/api.js)", () => {
   });
 
   test("on fetch user details action", async () => {
-    const USER_ID = 0;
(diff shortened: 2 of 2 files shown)


src/actions/api.js
```diff
@@ -239,11 +239,15 @@ export const onFetchProposal = token => dispatch => {
...
+  const regexp = /^[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$/i;
+  const valid = regexp.test(userId);
+  if (!valid)
+    return dispatch(act.RECEIVE_USER(null, "This is not a valid UUID: "+userId));
   return api
     .user(userId)
     .then(response => dispatch(act.RECEIVE_USER(response)))
-    .catch(error => {
-      dispatch(act.RECEIVE_USER(null, error));
+    .catch( () => {
```
fernandoabolafio: Why have you changed it to send a generic message instead of passing the error gotten from the API?
We should keep passing the error from the API as it describes better what the error is about.

src/actions/api.js
```diff
@@ -239,11 +239,15 @@ export const onFetchProposal = token => dispatch => {
 
 export const onFetchUser = userId => dispatch => {
   dispatch(act.REQUEST_USER(userId));
+  const regexp = /^[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$/i;
+  const valid = regexp.test(userId);
+  if (!valid)
+    return dispatch(act.RECEIVE_USER(null, "This is not a valid UUID: "+userId));
```
fernandoabolafio: Since this message is targeting end users, I think the message can a bit more friendly, such as: "This is not a valid user ID: <user id>"

src/actions/api.js
thi4go: Confused myself on this one, but fixed it ! 👍

PR #821, opened 2018-10-31
[vote stats] Display quorum and pass percentage on vote stats

### LINKED ISSUE #810: Show proposal quorum requirement on UI
The quorum requirement for proposal votes is set at 20%, which will usually be around 8,192 votes but varies slightly depending on the ticket pool size when voting opens. It would be nice to show this on the UI, so that people can easily see how close proposals are to reaching the quorum requirement.

fernandoabolafio:
**Preview:**
Note: the percentage values in the chart are mocked and doesn't reflect the actual data

screen shot 2018-10-31 at 11 47 24 am

screen shot 2018-10-31 at 11 47 30 am

closes #810

### DIFF
--- a/src/components/StackedBarChart.js
+++ b/src/components/StackedBarChart.js
@@ -1,4 +1,5 @@
 import React from "react";
+import Tooltip from "./Tooltip";
 
 const barStyle = {
   display: "flex",
@@ -20,6 +21,7 @@ const labelStyle = {
 };
 
 const wrapperStyle = {
+  position: "relative",
   display: "flex",
   width: "100%",
   height: "24px",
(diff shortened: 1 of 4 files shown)


thi4go:
Tested and working great @fernandoabolafio. Really liked the new status bar color and the VoteStats component optimizations on passing down our props.

One thing that I'd maybe change is the color used on the quorum votes; red really bothers me and gives out the impression that something is wrong. What do you think if it were a grey colored text until it hits the minimum quorum, and then it can be green to represent its success? That's just personal preference though, other than that, everything LGTM.

fernandoabolafio:
thanks for reviewing it @thi4go.
The quorum color is updated now:
screen shot 2018-11-01 at 10 46 20 am
it will become green once the mininum quorum percentage is reached

thi4go:
Very appealing!! Thanks for this PR @fernandoabolafio. All is working great.

LGTM.

PR #848, opened 2018-11-05
Add user preferences.

sndurkin:
This commit adds a new Preferences tab on the user detail page, with
several checkboxes to opt into email notifications. The Preferences
tab is visible to admins to help with debugging purposes, but the
checkboxes are disabled as the admin can't change them.

It also changes the admin-only /v1/user/edit route to /v1/user/manage,
and reuses /user/edit for editing user preferences.

This PR depends on decred/politeia#577.


tiagoalvesdulce:

PR #851, opened 2018-11-05
Prevent unnecessary displaying of 'readMore' button for comments

### LINKED ISSUE #817: 'Read more' link improvement
Every comment has a `Read more` link at the bottom, even if all the text is already displayed.
image

This is minor, but annoying for new users. I just found myself clicking the `Read more` link a couple times, assuming there was more (as this is standard convention), only to have the `Read more` simply flip to `Read less` without the text changing.

camus-code:
Resolves #817

Classnames changed as well so CSS was modified to reflect that

### DIFF
--- a/src/components/snew/ThingComment.js
+++ b/src/components/snew/ThingComment.js
@@ -15,9 +15,9 @@ class ThingComment extends React.PureComponent {
   }
   handleCommentMaxHeight = () => {
     const insertAfter = (newNode, referenceNode) => referenceNode.parentNode.insertBefore(newNode, referenceNode.nextSibling);
-    const commentsDiv = Array.prototype.slice.call(document.querySelectorAll(".comment > .entry > .usertext"));
+    const commentsDiv = Array.prototype.slice.call(document.querySelectorAll(".entry > .usertext "));
     const overflowDivs = commentsDiv.filter(function isOverflown(element) {
-      return element.scrollHeight > element.clientHeight || element.scrollWidth > element.clientWidth;
+      return element.scrollHeight > element.clientHeight && element.scrollHeight > 530;
     });
     overflowDivs.forEach((overflowDiv) => {
       const readMore = document.createElement("a");
--- a/src/style/index.css
+++ b/src/style/index.css
@@ -1490,14 +1490,17 @@ form div.usertext-body.may-blank-within.md-container {
   float: right;
 }
 
-.comment > .entry > .usertext {
-	max-height: 500px;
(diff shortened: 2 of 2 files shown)


src/components/snew/ThingComment.js
```diff
@@ -15,10 +15,11 @@ class ThingComment extends React.PureComponent {
   }
   handleCommentMaxHeight = () => {
     const insertAfter = (newNode, referenceNode) => referenceNode.parentNode.insertBefore(newNode, referenceNode.nextSibling);
-    const commentsDiv = Array.prototype.slice.call(document.querySelectorAll(".comment > .entry > .usertext"));
+    const commentsDiv = Array.prototype.slice.call(document.querySelectorAll(".entry > .usertext "));
     const overflowDivs = commentsDiv.filter(function isOverflown(element) {
-      return element.scrollHeight > element.clientHeight || element.scrollWidth > element.clientWidth;
+      return element.scrollHeight > element.clientHeight && element.scrollHeight > 530;
     });
+    console.log(overflowDivs);
```
brunobraga95: I think you forgot this here

src/style/index.css
```diff
@@ -1490,14 +1490,15 @@ form div.usertext-body.may-blank-within.md-container {
   float: right;
 }
 
-.comment > .entry > .usertext {
-	max-height: 500px;
+.entry > .usertext {
+	max-height: 35.71428575em;
```
brunobraga95: how did you get to this number? :eyes:

src/components/snew/ThingComment.js
camus-code: Good catch mate. Sorry about that.

src/style/index.css
camus-code: Meant to put a comment for this one! Anyway, the line-height for comments is "1.42857143em", and so I set the max height at `line-height * 25` which is equal to `35.71428575em`.

src/style/index.css
```diff
@@ -1490,14 +1490,16 @@ form div.usertext-body.may-blank-within.md-container {
...
-	max-height: 500px;
+/* Max height is (line-height * 25)  */
+.entry > .usertext {
+	max-height: 35.71428575em;
 	overflow: hidden;
 }
 
 .collapsed {
-	max-height: 500px;
+	max-height: 35.71428575em;
```
tiagoalvesdulce: I think this line could use this comment (`Max height is (line-height * 25)`) too.

PR #859, opened 2018-11-06
Add option with and without redux-logger for dev

### LINKED ISSUE #762: Create a flag for yarn start script to select whether to use redux logs on dev env or not

tiagoalvesdulce:
Fixes #762

### DIFF
--- a/src/configureStore.js
+++ b/src/configureStore.js
@@ -16,8 +16,6 @@ const configureStore = preloadedState => {
       compose(applyMiddleware(thunkMiddleware))
     );
   } else {
-    const loggerMiddleware = createLogger();
-
     const composeEnhancers =
     typeof window === "object" &&
     window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__ ?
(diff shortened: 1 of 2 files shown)


src/configureStore.js
```diff
@@ -16,27 +16,38 @@ const configureStore = preloadedState => {
...
-        ...loadStateLocalStorage,
-        ...preloadedState
-      }
-      ,
-      composeEnhancers(applyMiddleware(
-        thunkMiddleware,
-        loggerMiddleware
-      ))
-    );
+    if (process.env.REACT_APP_USE_REDUX_LOGGER) {
```
brunobraga95: you are repeating some lines of code that could be reduced.

```
what about this approach? I have seen it before and it is cleaner and better at scalling.
const middlewares = [
  thunkMiddleware,
  reactRouterMiddleware,
  process.env.REACT_APP_USE_REDUX_LOGGER && loggerMiddleware,
].filter(Boolean);
return createStore(rootReducer, {...loadStateLocalStorage, ...preloadedState}, composeEnhacers(applyMiddleware(...middlewared)))
```

PR #864, opened 2018-11-07
Make login page & message more responsive on window resize

### LINKED ISSUE #612: Spacing issue on "Before you sign up..." info box (Chrome)
There is no margin on the right side of the "Before you sign up..." info box that shows after you create a new account.

https://localhost:3000/user/signup

Chrome (the browser that my localhost automatically starts up, even if I'm using another browser (Brave))
image

The spacing does appear to look OK in Firefox, though left-indenting generally looks weird IMHO.

Firefox
image

camus-code:
Resolves #612

### DIFF
--- a/src/style/theme/index.css
+++ b/src/style/theme/index.css
@@ -148,6 +148,13 @@ body>.appWrapper>.content {
     padding: 92px 352px 64px 16px;
     margin: 0;
 }
+body>.appWrapper>.content>#login {
+		margin-left: 0;
+		padding-right: 1em;
+    position: absolute;
+    width: 98.75%;
+    margin: 0;
+}
 
 .link {
     padding: 16px;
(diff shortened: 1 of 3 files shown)


src/style/theme/index.css
```diff
@@ -148,6 +148,13 @@ body>.appWrapper>.content {
     padding: 92px 352px 64px 16px;
     margin: 0;
 }
+body>.appWrapper>.content>#login {
```
brunobraga95: is this necessary? could you create a class and avoid ids and deep selectors?

camus-code: the reason I went this route is because these contents are being created in snew-classic and I had some issues directly modifying that .less for desired results.

if there's a big problem with this approach I can try again with snew though.

PR #867, opened 2018-11-07
Keep tab state for proposals list and fix redirect issues

### LINKED ISSUE #752: [proposals list] persist the filter state on refresh
If the user is seeing the active voting proposals and hit refresh, after refresh the list filter whatever is the default filter. 
We need a way to persist the user choice so after refreshing the page it will display the same list as before. 
IMO the best way to achieve this is by setting the filter as a query string parameter in the URL.

### LINKED ISSUE #868: Unwanted redirects after refresh
- log in as an admin
- go to the Admin Proposals Page and refresh the browser: the page will be redirected to the Public Proposals Page.
- go to the User Proposals page and refresh the browser: the page will be redirected to the Admin Proposals Page

fernandoabolafio:
closes #752
closes #868

### DIFF
--- a/src/components/ProposalFilter.js
+++ b/src/components/ProposalFilter.js
@@ -93,17 +93,29 @@ const mapHeaderToCount = {
 class ProposalFilter extends React.Component {
   constructor(props) {
     super(props);
+    this.handleUpdateFilterValueForQueryValue(props);
+  }
+  componentDidUpdate(prevProps) {
+    this.handleUpdateQueryForFilterValueChange(prevProps);
+  }
+  handleUpdateFilterValueForQueryValue = (props) => {
     const { location, header, handleChangeFilterValue } = props;
     const { tab } = qs.parse(location.search);
     const tabOptions = mapHeaderToOptions[header];
-    const tabValueIsValidForHeader = tab && tabOptions.find(op => op.value.toString() === tab);
-    if (tabValueIsValidForHeader) {
-      handleChangeFilterValue(parseInt(tab, 10));
+    const validTabOption = tab && tabOptions.find(op => op.label === tab);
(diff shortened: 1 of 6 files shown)


src/components/ProposalFilter.js
```diff
@@ -86,19 +90,57 @@ const mapHeaderToCount = {
...
+    this.handleUpdateQueryForFilterValueChange(prevProps);
+  }
+  handleUpdateFilterValueForQueryValue = (props) => {
+    const { location, header, handleChangeFilterValue } = props;
+    const { tab } = qs.parse(location.search);
+    const tabOptions = mapHeaderToOptions[header];
+    if (!tabOptions)
+      return;
+
+    const validTabOption = tabOptions.find(op => op.label === tab);
```
brunobraga95: where is op coming from?

src/components/Router/AuthenticatedRoute.js
```diff
@@ -23,15 +16,29 @@ class AuthenticatedRoute extends Component {
...
       && stateFromLocalStorage.api
       && stateFromLocalStorage.api.me
-      && stateFromLocalStorage.api.me.response)
-    ) {
+      && stateFromLocalStorage.api.me.response);
+
+    if(validUserLoginFromApi || validUserLoginFromLocalStorage) {
       return true;
     } else {
+      this.props.redirectedFrom(this.props.location.pathname);
```
brunobraga95: the else above is not necessary, you can cut it

src/components/ProposalFilter.js
tiagoalvesdulce: `op` is the first parameter of the callback passed to find. It's the current element being processed in the array.

PR #872, opened 2018-11-08
Add user option to paste identity & create upload identity modal

### LINKED ISSUE #698: [UI/UX] Allow paste for identity import
The only way to import an identity currently is by importing a file.  Ideally, the dialog would allow you to directly paste the identity as an alternative to requiring a file.

This is desirable because many security conscious users have specialized tools dedicated to storing this type of information encrypted, and, thus, it would be ideal to allow the user to paste it directly instead of first having to create a file to upload.

camus-code:
Closes #698

Also decoupled the PrivateKeyIdentityManager into three separate components (Import, Form, and Download). I think it makes things a little easier to work with.


tiagoalvesdulce:

camus-code:
@tiagoalvesdulce Thanks for the comments! The error message is updated and I think I've resolved the  merge conflicts

fernandoabolafio:
@camus-code  The reason the tests were failing is that you've exported constants from a component file and imported it on the `helpers.js` file, so the component code executed during the tests execution. Please, if you need constants to be used outside of the component scope, define them in the `constants.js` file and import them in the component and in the other files where you need it.

tiagoalvesdulce:
Thanks for fixing the tests, @fernandoabolafio.

I'm still seeing this:
image

Also, the errors are not being cleaned up when the modal is closed. So, if you show an error message, close the modal and then open it again, the messages will still be there.

PR #873, opened 2018-11-09
Add General Tab Functionality for all Users on /user/id Page

### LINKED ISSUE #871: Add a "General" tab to the user details page
There is currently a "General" tab when viewing your own user details page, but not when viewing the user details page of other users.

I think it would be helpful to add a "General" tab that displays the user's public data.  Specifically, the user's UUID, whether the user is an admin, the user's active identity, and a history all of the user's previous identities.  

All of this data is already returned in the response for the user details api call `/user/[uuid]`.

camus-code:
Resolves #871

### DIFF
--- a/src/components/PrivateKeyFormManager.js
+++ b/src/components/PrivateKeyFormManager.js
@@ -1,0 +1,83 @@
+import React from "react";
+import { reduxForm, Field } from "redux-form";
+import userConnector from "../connectors/user";
+import * as pki from "../lib/pki";
+
+const PUBLIC_KEY_MISMATCH = "The provided public key doesn't match the key stored in the server.";
+const INVALID_KEY_PAIR = "The provided key pair is not valid.";
+const LOAD_KEY_FAILED = "Sorry, something went wrong while importing the identity file, please try again. If the error persists, contact the Politeia support.";
+
+class PrivateKeyFormManager extends React.Component {
+  render(){
+    return (
+      <div>
+        <form className="import-identity-form" onSubmit={this.props.handleSubmit(this.handleKeys)}>
+          <div style={{ width: "45em" }} className="c-form-group">
+            <Field
+              className="c-form-control"
+              id="publicKey"
+              name="publicKey"
+              placeholder="Public Key"
+              component="input"
+              type="text" />
+          </div>
+          <div style={{ width: "45em" }} className="c-form-group">
+            <Field
+              className="c-form-control"
+              id="secretKey"
+              name="secretKey"
+              placeholder="Private Key"
+              type="password"
+              component="input" />
+            <button
+              type="submit"
+              tabIndex={7}
+              disabled={this.props.pristine || this.props.submitting}
+              style={{ margin: "2em 0 2em 2em", float: "right" }}
+              className="c-btn c-btn-primary c-pull-left">Submit Identity</button>
+          </div>
+        </form>
+      </div>
+    );
+  }
+
+// Handles user-inputted keys by using props to save form values into store
+// Uses same logic as OnSelectFiles from privKeyImportManager except 'props' is subsituted in place of 'keys'
+handleKeys = (props) => {
+  const { onIdentityImported, closeModal } = this.props;
+  props = { ...props,
+    publicKey: props.publicKey && props.publicKey.trim(),
+    secretKey: props.secretKey && props.secretKey.trim()
+  };
+  try {
+    this.auditIdentity(props);
+    pki.importKeys(this.props.loggedInAsEmail, props)
+      .then(() => {
+        onIdentityImported("Successfully imported identity") && closeModal();
+      })
+      .catch(e => {
+        console.error(e.stack);
+        onIdentityImported(null, LOAD_KEY_FAILED);
+      });
+  } catch(e) {
+    console.error(e.stack);
+    onIdentityImported(null, e);
+  }
+}
+
+//Same as one from privKeyImportManager except it checks form input instead of uploaded file
+auditIdentity = (props) => {
+  const { userPubkey } = this.props;
+
+  // check that the pubkey matches with the server one
+  if(props.publicKey !== userPubkey)
+    throw new Error(PUBLIC_KEY_MISMATCH);
+
+  // check that the key pair is valid
+  if(!pki.verifyKeyPair(props)) {
+    throw new Error(INVALID_KEY_PAIR);
+  }
+}
+}
+
+export default userConnector(reduxForm({ form: "form/import-identity" })(PrivateKeyFormManager));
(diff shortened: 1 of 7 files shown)


src/components/PrivateKeyImportManager.js
```diff
@@ -0,0 +1,88 @@
...
+          console.error(e.stack);
+          onIdentityImported(null, LOAD_KEY_FAILED);
+        });
+    } catch(e) {
+      console.error(e.stack);
+      onIdentityImported(null, e);
+    }
+  }
+
+  getJsonData = (base64) => {
```
brunobraga95: this method could be "key agnostic", since it is only about converting base64 to json, consider removing it from this component context and adding it to a helper function (or something similar). Even if it won't be used again (which we cannot be sure), it helps keeping the component logic restrict on what it proposes =)

src/components/UserDetail/GeneralTab.js
```diff
@@ -333,35 +340,58 @@ class GeneralTab extends React.Component {
...
+            <span>
+              <span style={{ fontWeight: "bold" }}>Expanded</span>
+              <a className="linkish" style={{ paddingLeft: "1em" }}onClick={() => this.setState({ showPastUserIdentities: false })}>(hide)</a>
+            </span> ) : (
+            <a className="linkish" onClick={() => this.setState({ showPastUserIdentities: true })}>Expand</a>
+          )}
+        </Field>
+        {showPastUserIdentities ?
+          <ul>{user.identities.map((identity, i) => (
+            <ul key={i} style={{ lineHeight: "2em" }}>
```
brunobraga95: dont use just the index to identify your elements

https://medium.com/@robinpokorny/index-as-a-key-is-an-anti-pattern-e0349aece318

src/components/UserDetail/GeneralTab.js
```diff
@@ -333,35 +340,58 @@ class GeneralTab extends React.Component {
...
-            <span>No</span>,
-            isAdmin && <ButtonWithLoadingIcon
-              className="c-btn c-btn-primary button-small"
-              text="Mark as paid"
-              disabled={isApiRequestingMarkAsPaid}
-              isLoading={isApiRequestingMarkAsPaid}
-              onClick={() => onEditUser(user.id, EDIT_USER_CLEAR_USER_PAYWALL)} />
-          ])}
-        </Field>
+        { isAdminOrTheUser ? <FieldSeparator /> : null }
```
brunobraga95: if you take a look here, you are doing the following:

```
if(X) {
  doFoo();
}

if(X) {
  dobar();
}
```

which is the same as 
```
if(X) {
  doFoo();
  doBar();
}
```

you can just 
```
{isAdmingOrTheUser ? 
  <React.Fragment>
    <FieldSeparator />
    <Field label = "has paid">
       .....
   </Field>
  </React.Fragment>
 :null
}
```

src/components/UserDetail/GeneralTab.js
```diff
@@ -419,7 +452,15 @@ class GeneralTab extends React.Component {
...
             </Field>
           </React.Fragment>}
+        { isAdminOrTheUser ? <FieldSeparator /> : null }
+        { !isAdminOrTheUser ?
+          <Field label="Admin (Y/N)">
+            {user.isAdmin ? <div>Yes</div> : <div>No</div> }
+          </Field> : null }
         <FieldSeparator />
+        <Field label="User ID">
+          <div>{user.id}</div>
```
brunobraga95: do you need this div?

src/components/UserDetail/GeneralTab.js
```diff
@@ -191,23 +191,25 @@ class GeneralTab extends React.Component {
       onResetRescan,
       rescanUserId
     } = this.props;
-    const { showIdentityHelpText, pubkey, pubkeyStatus } = this.state;
+    const { showIdentityHelpText, pubkey, pubkeyStatus, showPastUserIdentities } = this.state;
     const userHasActivePaywall = user && user.newuserpaywalladdress && user.newuserpaywallamount;
     const isUserPageOwner = loggedInAsUserId === user.id;
     const hasTheRescanResult = amountOfCreditsAddedOnRescan !== undefined && rescanUserId === user.id;
+    const isAdminOrTheUser = user && (isAdmin || loggedInAsUserId === user.id);
```
brunobraga95: This component has many checks like the following:
```
if(isAdminOrTheUser) {
  doX();
} else doY();
```
it is getting hard to understand it, it would be good to refactor it. You should not do this on this PR
but it is good to keep in mind.

I will open an issue about it and work on it.

tiagoalvesdulce:

src/components/PrivateKeyImportManager.js
camus-code: Nice suggestion. I'm going to implement this idea in #872 since it's a bit more relevant for that PR as opposed to this one.

src/components/UserDetail/GeneralTab.js
camus-code: Thanks for the link; I wasn't aware of that. I'm making it so the key is just the corresponding pubkey

src/components/UserDetail/GeneralTab.js
camus-code: Nope =)

src/components/PrivateKeyFormManager.js
```diff
@@ -0,0 +1,83 @@
...
+import * as pki from "../lib/pki";
+
+const PUBLIC_KEY_MISMATCH = "The provided public key doesn't match the key stored in the server.";
+const INVALID_KEY_PAIR = "The provided key pair is not valid.";
+const LOAD_KEY_FAILED = "Sorry, something went wrong while importing the identity file, please try again. If the error persists, contact the Politeia support.";
+
+class PrivateKeyFormManager extends React.Component {
+  render(){
+    return (
+      <div>
```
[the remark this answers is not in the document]
camus-code: div needed here either. Like your first comment, I'll also address this one in my latest commit to PR #872

src/components/UserDetail/GeneralTab.js
camus-code: Really good idea @brunobraga95 . I thought the same thing towards the end of finishing this up but I figured it'd be better to just make it functional for now and then work on encapsulating it.

the easiest way to refactor my opinion would just be to make two separate components: a view for all users and a view for page owners / admins which would make for minimal checking of things like `isAdminOrTheUser`

camus-code:
@brunobraga95 @tiagoalvesdulce @fernandoabolafio had some pretty rough merge issues with git so I decided to just open up a new PR at #883

tiagoalvesdulce:
Closing because of #883

PR #880, opened 2018-11-12
Distinguish Author's comment + add prettier

### LINKED ISSUE #877: Distinguish proposal author's comments
It would be useful to have a label on comments posted by proposal author. Similar to Reddit's `[S]` label, which stands for "submitter".

I suggest adding `author` label after the username. Alternatively, slightly change the background of the comment box using a calm color like pale light blue.

In any case it must be quite noticeable.

tiagoalvesdulce:
Closes #877


xaur:
Please pardon my criticism and feel free to ignore it since I'm not a politeia dev, but I've been noticing a pattern I consider dangerous.

Bundling a small feature patch with a large code reformatting (+7.5k, -5k lines) in a single commit is not a good idea. Just like mixing features in general. You can't cherry-pick or revert a small nicely isolated change, and digging throuh Git history becomes hard.

tiagoalvesdulce:
Hey @xaur, just saw your comment and I agree with you. We will avoid this from now on.

PR #883, opened 2018-11-12
Add general tab functionality for all users on user/page

### LINKED ISSUE #871: Add a "General" tab to the user details page

camus-code:
Resolves #871

### DIFF
--- a/src/components/UserDetail/GeneralTab.js
+++ b/src/components/UserDetail/GeneralTab.js
@@ -77,7 +77,8 @@ class GeneralTab extends React.Component {
       showIdentityHelpText: false,
       openedVerification: false,
       pubkey: "",
-      pubkeyStatus: PUB_KEY_STATUS_LOADING
+      pubkeyStatus: PUB_KEY_STATUS_LOADING,
+      showPastUserIdentities: false
     };
   }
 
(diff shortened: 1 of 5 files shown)


src/components/UserDetail/GeneralTab.js
```diff
@@ -213,26 +214,34 @@ class GeneralTab extends React.Component {
...
+      pubkey,
+      pubkeyStatus,
+      showPastUserIdentities
+    } = this.state;
     const userHasActivePaywall =
       user && user.newuserpaywalladdress && user.newuserpaywallamount;
     const isUserPageOwner = loggedInAsUserId === user.id;
     const hasTheRescanResult =
       amountOfCreditsAddedOnRescan !== undefined && rescanUserId === user.id;
+    const isAdminOrTheUser = user && (isAdmin || loggedInAsUserId === user.id);
```
tiagoalvesdulce: 👍

src/components/UserDetail/GeneralTab.js
```diff
@@ -213,26 +214,34 @@ class GeneralTab extends React.Component {
...
-            {isAdmin ? (
-              <ButtonWithLoadingIcon
-                className="c-btn c-btn-primary button-small"
-                isLoading={isLoadingRescan}
-                onClick={() => onRescan(user.id)}
-                text="rescan"
-              />
-            ) : null}
-          </Field>
+          {isAdminOrTheUser ? (
```
tiagoalvesdulce: if you aren't using an `else` clause, e.g `condition ? doSomething() : null`, there's no need to use the ternary operator, you can go with `condition && doSomething()` for simplicity.

src/components/UserDetail/GeneralTab.js
```diff
@@ -524,7 +603,14 @@ class GeneralTab extends React.Component {
             </Field>
           </React.Fragment>
         )}
+        {isAdminOrTheUser ? <FieldSeparator /> : null}
```
tiagoalvesdulce: Here you are doing something like: `condition ? doSomething() : null` and then `!condition ? doSomethingElse() : null`. Just do: `condition ? doSomething() : doSomethingElse()`

src/components/UserDetail/index.js
```diff
@@ -20,17 +17,11 @@ class UserDetail extends Component {
...
     this.props.onFetchProposalsVoteStatus();
   }
-
-  componentDidUpdate(prevProps) {
-    const { loggedInAsUserId, editUserResponse, user, isAdmin } = this.props;
-    if (editUserResponse) {
+  componentDidUpdate() {
+    const { editUserResponse } = this.props;
+    if(editUserResponse) {
       window.location.reload();
```
tiagoalvesdulce: Why do we need this?

camus-code: Not too sure to be honest. It was in here beforehand and I couldn't exactly figure out the purpose so I was a little worried to remove it.

src/components/UserDetail/GeneralTab.js
camus-code: thanks for the suggestion mate. went through and took out all the unnecessary ternaries

tiagoalvesdulce:
Depends on https://github.com/decred/politeia/pull/598

tiagoalvesdulce:
There is an issue showing the active public key for the user when accessing the route logged out. Also, I think it shouldn't be `unpaid_user's Public Key` but `Active Public Key` instead. Another thing that might be related to this PR is that the number of proposals is not showing in the `Proposals` tab when it's 0.

screen shot 2018-11-14 at 20 12 41

camus-code:
I think my last commit covered just about everything @tiagoalvesdulce . Ready for a look when you have a chance

tiagoalvesdulce:
Looks good but the paddings are a bit messed up when looking another user's tab. Look:

screen shot 2018-11-20 at 09 53 51

Issue #732, opened 2018-10-17
Feature Request: Highlight new comments since last visit


davecgh:
As the title suggests, it would be nice to have some type of visual indicator, perhaps a highlighted background, for comments that are new since your last visit.  It can be difficult to follow comments with several long branching threads without this feature.

victorgcramos:
I can take a look at that!

brunobraga95:
What about something like a button to jump to new comments. It would go to the right page and scroll down to until the right comment is in evidence.

Highlighting everything that hasn't been seen seems to be a big design change I believe (depending on how it is done of course). Just jumping into the right comment (the first one after the last viewed) sounds cleaner.

MariaPleshkova:
Created a design for this issue:
**Preview:** https://xd.adobe.com/view/16d02845-4b7e-452f-78dd-045a5755729b-60e1/
**Specs:** https://xd.adobe.com/spec/71272398-dc57-491f-73f8-52d17c4f81fc-0716/
![image](https://user-images.githubusercontent.com/8841018/51192043-bb9a8c00-18f6-11e9-8748-c5c585f6a3a6.png)

victorgcramos:
I'm working on #1021 and since it is related to that issue, I can add the new design.

xaur:
There is an advantage to highlight (change background) of the entire comment box: it gives user a visual clue of the clickable area. This is how it's done on Reddit now.

marcopeereboom:
New comments show up top. If this is not good enough please open a new issue.

PR #897, opened 2018-11-18
Feature/highlight new comments

victorgcramos:


victorgcramos:
Fixes #732

victorgcramos:
https://github.com/decred/snew-classic-ui/pull/24

src/components/snew/ThingComment.js
```diff
@@ -74,6 +74,7 @@ class ThingComment extends React.PureComponent {
       ...props
     } = this.props;
     const isProposalAbandoned = proposalStatus === PROPOSAL_STATUS_ABANDONED;
+    console.log(lastVisit);
```
tiagoalvesdulce: You left a log here.

victorgcramos:
This happens because the accesstime is set only when the proposal is fetched. This was discussed on Slack and I was told that this was the best way to deal with the situation, since I was not allowed to set the accesstime on a new request.

tiagoalvesdulce:
LGTM

PR #931, opened 2018-11-29
Split finished tab into approved and rejected

### LINKED ISSUE #904: Split "Finished voting" tab into "Approved" and "Rejected" tabs
If a user wants to see "finished proposals" I think they're usually going to be looking specifically for approved proposals (to see how they're doing) or maybe rejected proposals (to see what's not being approved). Therefore I think it would make sense to split the finished proposals into these tabs. The system should already know whether each proposal is approved/rejected because the criteria and results are stored.

tiagoalvesdulce:
Fixes #904


RichardRed0x:
I don't see this change on the live or test site UI, am I missing something?

PR #942, opened 2018-12-10
Improve structure of proposal's informations

### LINKED ISSUE #925: Show when proposals were approved for display on the UI
"Submitted X days ago" is confusing as the proposal may have been in the moderation queue for some time. The second research proposal for example was submitted 4/5 days before it was approved. When it appears on the proposals site it says "submitted 5 days ago" and it looks like it has been ignored with 0 comments from the community.

The "submitted" time doesn't line up with anything in the user's experience (i.e. they may have checked the site yesterday and not seen a proposal that is visible today and was submitted "2 days ago"). I think it would be better to replace or add "published" or "approved" x time ago.

fernandoabolafio:
This PR adds a better view formatting for the proposal's information such as version, token, author, etc..
screen shot 2018-12-10 at 1 44 50 pm

closes #925

### DIFF
--- a/src/components/snew/ThingLink.js
+++ b/src/components/snew/ThingLink.js
@@ -1,6 +1,5 @@
 import React from "react";
 import { DateTooltip } from "snew-classic-ui";
-import { getProposalStatus } from "../../helpers";
 import { withRouter } from "react-router-dom";
 import * as modalTypes from "../Modal/modalTypes";
 import actions from "../../connectors/actions";
@@ -62,6 +61,9 @@ class ThingLinkComp extends React.Component {
       subreddit,
       subreddit_id,
       created_utc,
+      publishedat,
+      abandonedat,
+      censoredat,
       title,
       url,
       permalink,
@@ -105,7 +107,10 @@ class ThingLinkComp extends React.Component {
       review_status === PROPOSAL_STATUS_UNREVIEWED ||
       review_status === PROPOSAL_STATUS_UNREVIEWED_CHANGES;
     const isVetted = review_status === PROPOSAL_STATUS_PUBLIC;
-    const displayVersion = review_status === PROPOSAL_STATUS_PUBLIC;
+    const displayVersion =
+      version &&
+      parseInt(version, 10) > 1 &&
+      review_status === PROPOSAL_STATUS_PUBLIC;
     const isVotingActiveOrFinished =
       voteStatus === PROPOSAL_VOTING_ACTIVE ||
       voteStatus === PROPOSAL_VOTING_FINISHED;
@@ -208,7 +213,7 @@ class ThingLinkComp extends React.Component {
             >
               {title}{" "}
               {review_status === PROPOSAL_STATUS_UNREVIEWED_CHANGES ? (
-                <span className="font-12 warning-color">(edited)</span>
+                <span className="font-12 warning-color">edited</span>
               ) : null}
             </Link>{" "}
             {expanded && displayVersion && version > 1 ? (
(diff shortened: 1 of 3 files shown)


src/components/snew/ThingLink.js
```diff
@@ -270,46 +270,92 @@ class ThingLinkComp extends React.Component {
...
             </div>
           </span>
-          <span className="tagline">
-            <span className="submitted-by">
-              {hasBeenUpdated ? "updated " : "submitted "}
-              <DateTooltip createdAt={created_utc} />
-              {author && (
-                <span>
-                  {" by "}
+          {!draftId ? (
```
brunobraga95: consider extracting to a different component, this component already has more than 600 lines

src/components/snew/ThingLink.js
```diff
@@ -270,46 +270,92 @@ class ThingLinkComp extends React.Component {
...
+            <ul className="proposal-details tagline">
+              <li className="field">
```
brunobraga95: these li items seems to be very similar, I believe than can extracted to a function which will make the code smaller and easier to understand

src/components/snew/ThingLink.js
```diff
@@ -270,46 +270,92 @@ class ThingLinkComp extends React.Component {
...
+                  <span className="value">
+                    <DateTooltip createdAt={created_utc} />
+                  </span>
+                </li>
+              </ul>
+              <div className="tagline proposal-draft">
+                Saved as draft
+                <span
+                  className="delete-draft"
+                  onClick={() => {
```
brunobraga95: https://medium.freecodecamp.org/why-arrow-functions-and-bind-in-reacts-render-are-problematic-f1c08b060e36

fernandoabolafio:
@brunobraga95 I addressed all of your comments.

fernandoabolafio:
@sndurkin thanks for your comments. These changes came from me only. In my opinion, the proposal information would be better summarized/structured by having the label-value format. Although, your comments make sense to me and I will try a different approach to fit the key information there.

fernandoabolafio:
@sndurkin I’ve done a few modifications on the information structure for proposals based on your comments on my PR.
This is how it looks like:
- Published date always displayed :white_check_mark:
- Version is only displayed if > 1 :white_check_mark:
- Keep other stuff as it was (token, status label) :white_check_mark:

The only thing I kept is to show the abandoned date as well. I am not sure how valuable is this info though.

**Public proposals:**
screen shot 2018-12-11 at 11 29 47 am

**Unvetted proposals:**
screen shot 2018-12-11 at 11 43 34 am

sndurkin:
That looks better to me, although I still don't think the token needs to be displayed except maybe on the detail page next to the download links, and it's probably just because I'm used to Reddit, but it would look better to me to move the comments next to the permalink. That would keep the amount of information displayed under the title to a minimum.

But please note - I don't have strong feelings about these things, just offering my 2 cents.

fernandoabolafio:
@sndurkin thanks for reviewing it again. I like your comments and I applied them. The token and status are no longer displayed as the former can be seen in the URL and the latter can be inferred (e.g the proposal is in the public list it is either public or abandoned (which is shown by the different color).
The comments links are now clickable and if you click it will redirect to the proposal page with the page scrollTop aligned with the comment area.

tiagoalvesdulce:
Looks good but the number 0 is not showing when the comments list is empty.

screen shot 2018-12-19 at 09 40 39

fernandoabolafio:
That was my purpose @tiagoalvesdulce. It simply displays "comment" when no comment has been made yet.

PR #943, opened 2018-12-10
Improve UX of external link warning

### LINKED ISSUE #770: External link warning is difficult to read
This is the current external link warning modal:

image

I have a couple problems with it:

* It's difficult to read due to the center-aligned text. Left-aligned text is easier to read because every new line starts at the same place. In other words, the user's eyes don't have to jump to a different spot for every new line.

* The red text grabs the user's attention, but in a bad way because the text is meant to be read in context. For example, the words "do not" are highlighted in red but the user doesn't immediately know what not to do without reading the rest of the sentence, making the red text just distracting.

* The domain doesn't really need to be on a separate line as it's a subset of the full link; that section of link text could just be highlighted red.

Here is/was Facebook's external link warning, to show an alternative approach:

image

Some minor UX things to note in Facebook's example:

* A warning icon which experienced users can use to quickly identify the page they're looking at.
* A user-friendly warning message ("Please be careful"), in contrast to Politeia's "External Link Warning" title, which seems to just describe what it is.
* Buttons which use action verbs as text instead of Yes/No (more info: https://hackernoon.com/let-the-action-buttons-in-the-dialog-box-scream-its-outcome-b319c5efc698)

camus-code:
Resolves #770 

This commit works to improve the styling of the external link warning modal. This code also highlights the domain name in red, as opposed to showing it on a separate line. Minor changes to the copy were made as well.

### DIFF
--- a/src/components/Modal/contents/ConfirmAction.js
+++ b/src/components/Modal/contents/ConfirmAction.js
@@ -19,6 +19,7 @@ class ConfirmAction extends React.Component {
     const { me } = this.props;
     return (
       <ModalContentWrapper
+        style={me.payload.style}
         title={me.payload.title || "Confirm Action"}
         submitText={me.payload.submitText || "Yes"}
         cancelText={me.payload.cancelText || "No"}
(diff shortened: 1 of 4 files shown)


camus-code:
Thanks for the review @sndurkin ! Completely missed the misaligned title / buttons. I wasn't too sure about the text either but decided to wait for some feedback. Working on this now

camus-code:
@sndurkin I think everything is up to par now.

PR #944, opened 2018-12-17
Directions for verifying proposal censorship

### LINKED ISSUE #688: Improve instructions for how a user can demonstrate censorship of their proposal
We should have a page somewhere that walks through this step by step, including steps like downloading the proposal bundle and how to demonstrate to others that a proposal was submitted.

camus-code:
Resolves #688 

Adds basic instructions to the onboard modal for verifying a proposal's censorship status.


RichardRed0x:
I'm not sure about this text, it seems to just explain how to look at the proposal bundle and see the status that means "censored". The intention with issue #688 was to show users how they can use their pi keys and censorship tokens to demonstrate that their proposal existed, even if it has been completely removed from proposals.decred.org.

PR #972, opened 2019-01-07
Fixes selection of comment sort options

### LINKED ISSUE #970: Comments "Sort by" select doesn't keep opened
The Select for sorting the comments it not keeping the opened state.

thi4go:
Turns out every click we do in the gui, we invoke `modalClickHandler`, and therefore it ran the `closeAllModals()` function. The select is being treated as a modal, so it would close out once we clicked the selection box. Now it checks if there is any modal opened by us in the state before closing anything.

Closes #970

### DIFF
--- a/src/components/Modal/ModalStack.js
+++ b/src/components/Modal/ModalStack.js
@@ -42,7 +42,11 @@ class ModalStack extends React.Component {
   }
   //function which closes all modals if user clicks outside of modal window
   modalClickHandler = event => {
-    if (event.target.closest(".modal-content")) return;
+    if (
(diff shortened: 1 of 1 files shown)


thi4go:
Hey @fernandoabolafio thanks for the explanation man. One thing that I thought was strange was that whenever we click pgui, `modalClickHandler` gets triggered, even if there isn't any modals opened. Is that the expected behaviour, or something we should address eventually? 👍

fernandoabolafio:
@thi4go good catch. This needs to be fixed. I created an issue for that here:
https://github.com/decred/politeiagui/issues/974

PR #984, opened 2019-01-12
Implements click listener component

### LINKED ISSUE #974: Modals: only set the click listener if there is a modal opened
Currently, the "click" listener is added in the `ModalStack` component which is always rendered. It needs to be associated with an opened modal and be removed once the modal is closed.

thi4go:
closes #974

### DIFF
--- a/src/components/Modal/Modal.js
+++ b/src/components/Modal/Modal.js
@@ -1,24 +1,14 @@
 import React from "react";
-import PropTypes from "prop-types";
-import OnKeyListener from "../OnKeyListener";
-
-const ESC_CODE = 27;
 
 const Modal = props => {
   return (
-    <OnKeyListener keyCode={ESC_CODE} onKeyPress={props.onClose}>
-      <div className="modal">
-        <div className="modal-overlay" />
-        <div className="modal-container" style={props.style}>
-          {props.children}
-        </div>
+    <div className="modal">
+      <div className="modal-overlay" />
+      <div className="modal-container" style={props.style}>
+        {props.children}
       </div>
-    </OnKeyListener>
+    </div>
   );
-};
-
-Modal.prototypes = {
-  onClose: PropTypes.func
 };
 
 export default Modal;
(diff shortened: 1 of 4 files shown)


thi4go:
Hey @fernandoabolafio thanks for the review and the suggestions, that will be nice indeed! I'll turn that logic to a component, so that we can specify conditions for the click listeners.

thi4go:
Hey @fernandoabolafio finished creating the component! Turns out that OnKeyListener was being rendered two times, one in `Modal` and one in `ModalContentWrapper`. I encapsulated the listeners on the latter, and `Modal` now is a few lines of code. Should we refactor and get rid of it? I think it's only job right now is assigning diferent ID's to the childs.

fernandoabolafio:
@thi4go thanks for creating that component, and a nice catch of the listener being rendered twice! One thing that worries me about having the `OnKeyListener` and `OnClickListener` within the `ModalContentWrapper` is that not every Modal uses that wrapper. You can take a look at the Diff modal for instance.
So I think we need either to enforce that every modal uses the `ModalContentWrapper` in a way that its style is fully customizable or keep the listeners inside the `Modal` which is a wrap for every modal. What do you think?

thi4go:
@fernandoabolafio thanks for the review mate.

One thing that led me to encapsulate the listeners in the wrapper was that it already contained the props we needed to pass down to the listeners components, such as the `onClose` function and the `disabled` prop. For instance, the `OnKeyListener` on the `ModalContentWrapper` gets rendered with the following props:

```
<OnKeyListener
        ID={modalID}
        keyCode={ESC_KEY}
        onKeyPress={onClose || onCancel}
        disabled={disableCloseOnEsc}
      >
```

Rethinking now, I believe the best solution is to encapsulate the listeners on the `Modal` indeed, and leave it as the main wrapper for all modals. We'll just need to add a connector for the main wrapper, the `Modal`, with the props that the listeners need. That way we don't have to force new modals to use the content wrapper, and we'd avoid some refactor effort on making the wrappre generic for every modal style. What do you think?

PR #986, opened 2019-01-13
Remove onboard modal and link to dcrdocs

### LINKED ISSUE #947: Extract docs from modals into linkable pages, source them from .md
I suggest changing all documentation available in the modals to separate pages that can be linked to. Such pages can even work without javascript.

Ideally, the single source of truth for the docs are Markdown files. It is easier to engage copy writers to change Markdown files than HTML or JS files. Then you can build .html and .js from the .md.

camus-code:
Closes #947 

This commit removes our current onboarding modal and converts all existing links to DcrDocs


src/App.js
```diff
@@ -72,9 +75,16 @@ class Loader extends Component {
             </React.Fragment>
           ),
           cancelText: "Maybe later",
-          submitText: "Yes, show me more"
+          submitText: (
+            <Markdown
```
fernandoabolafio: Please, do not use the Markdown render here, it is meant to display text copy content when it's really needed. Use either anchor or button tags as appropriate;

src/components/snew/SubmitPage.js
```diff
@@ -123,12 +126,14 @@ class SubmitPage extends React.Component {
                             rows={20}
                             cols={80}
                           />
-                          <a
+                          <Markdown
```
fernandoabolafio: Don't use the Markdown render here. Use an anchor tag, please.

src/components/snew/SubmitPage.js
```diff
@@ -35,6 +36,8 @@ const normalizer = (value, previousValue) => {
   return files;
 };
 
+const proposalGuidelines = "[**Learn how to format your proposal**](https://docs.decred.org/governance/politeia/proposal-guidelines/)";
```
fernandoabolafio: Please, make the `https://docs.decred.org/governance/politeia/proposal-guidelines/` a constant we can use in the different places where it is linked.

camus-code:
Thanks for the review @fernandoabolafio ! Fixing up these inline comments now 😄

src/App.js
```diff
@@ -45,44 +45,18 @@ class Loader extends Component {
...
-    }
+      openModal(WELCOME_MODAL);
 
-    if (!prevProps.apiError && this.props.apiError) {
-      // Unrecoverable error
-      if (this.props.apiError.internalError) {
-        this.props.history.push("/500");
-      } else {
-        console.error("ERROR:", this.props.apiError.message);
+      if (!prevProps.apiError && this.props.apiError) {
```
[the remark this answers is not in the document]
camus-code: Thanks for the suggestions @fernandoabolafio . Agree that some of this logic could be better condensed inside of some variables. Also missed that I placed the API conditional inside of the on boarding one. Fixing these now

xaur:
Does this also remove what was added in #944?

PR #993, opened 2019-01-18
Add tab to url

### LINKED ISSUE #908: User Details: control selected tab from URL parameter
Implement the same tabs behavior from Public Proposals list where the tab can be selected from the URL.

tiagoalvesdulce:
Closes #908

### DIFF
--- a/src/components/UserDetail/index.js
+++ b/src/components/UserDetail/index.js
@@ -2,15 +2,60 @@ import React, { Component } from "react";
 import { autobind } from "core-decorators";
 import UserDetailPage from "./Page";
 import userConnector from "../../connectors/user";
-import { USER_DETAIL_TAB_GENERAL } from "../../constants";
+import {
+  USER_DETAIL_TAB_GENERAL,
+  USER_DETAIL_TAB_PREFERENCES,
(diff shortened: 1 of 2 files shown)


src/components/UserDetail/index.js
```diff
@@ -28,12 +73,9 @@ class UserDetail extends Component {
 
   render() {
     const { isTestnet } = this.props;
-    let dcrdataTxUrl;
-    if (isTestnet) {
-      dcrdataTxUrl = "https://testnet.dcrdata.org/tx/";
-    } else {
-      dcrdataTxUrl = "https://explorer.dcrdata.org/tx/";
-    }
+    const dcrdataTxUrl = isTestnet
```
fernandoabolafio: Shouldn't we use the DcrdatTxLink component here?
https://github.com/decred/politeiagui/blob/master/src/components/DcrdataTxLink.js

tiagoalvesdulce: Yeah, we should, but we have to refactor some stuff in the `GeneralTab` and makes more sense to switch the implementation while doing it.

fernandoabolafio: ok

PR #999, opened 2019-01-23
Move event listeners from modalContentWrapper to Modal

### LINKED ISSUE #997: Move modal listeners to main wrapper
We need to move the OnKeyListener and OnClickListener from the ModalContentWrapper into Modal which is the "universal modal wrapper" so every modal can have access to the listener's functionality. The Modal API needs to doc how to control this by the exposed props.

camus-code:
Resolves #997

### DIFF
--- a/src/components/Modal/Modal.js
+++ b/src/components/Modal/Modal.js
@@ -1,14 +1,49 @@
 import React from "react";
+import PropTypes from "prop-types";
+import OnClickListener from "../OnClickListener";
+import OnKeyListener from "../OnKeyListener";
 
-const Modal = props => {
-  return (
-    <div className="modal">
-      <div className="modal-overlay" />
-      <div className="modal-container" style={props.style}>
-        {props.children}
+const ESC_KEY = 27;
+
+class Modal extends React.Component {
+  render() {
+    const {
+      onClose,
+      onCancel,
+      disableCloseOnEsc,
+      disableCloseOnClick,
+      children,
+      style
+    } = this.props;
+    return (
+      <div>
+        <OnKeyListener
+          keyCode={ESC_KEY}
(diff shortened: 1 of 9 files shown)


src/components/Modal/Modal.js
```diff
@@ -1,14 +1,53 @@
...
+      onClose,
+      onCancel,
+      disableCloseOnEsc,
+      disableCloseOnClick,
+      children,
+      style
+    } = this.props;
+    return (
+      <OnKeyListener
+        ID={modalID}
```
fernandoabolafio: Do we really need this ID? Iif we do why it isn't specified as a required prop inside OnKeyListener?

src/components/Modal/Modal.js
```diff
@@ -1,14 +1,53 @@
...
+    }
+  };
+  componentDidMount() {
+    this.focusFirstInputOnModal();
+  }
+  render() {
+    const {
+      onClose,
+      onCancel,
+      disableCloseOnEsc,
```
fernandoabolafio: How can a caller specifiy this props to the Modal component?

If you take a look at ModalStack.js you'll find this:
```js
renderModalContent = modalData => (
    <Modal key={modalData.type}>
      <ModalContent modalData={modalData} />
    </Modal>
  );
```
If we use the modal without the Redux API, that's fine because we can pass props straight into it. 
However, if we do use the API, notice that there aren't any props being passed down to the Modal besides type. Maybe we need an extra configuration objection to be passed into `openModal`?

src/components/Modal/Modal.js
camus-code: You're right about this one. Removed the field and things work just the same.

PR #1010, opened 2019-02-05
Add proposal summary field

### LINKED ISSUE #1000: Add proposal summary field
Copying this issue here as I think it can be solved in the GUI alone:
https://github.com/decred/politeia/issues/569 

Add a length-restricted summary field to proposals. This will require changes to our index.MD file and its related functions so that the field can be handled properly. 

It will also require changes on the submit proposal page so there is another form for inputting the summary. The proposal detail page will also need to have a new, separate section for displaying this summary.

camus-code:
Resolves #1000 

Depends on https://github.com/decred/politeia/issues/685

Also will require changes to dcrdocs (https://github.com/decred/dcrdocs/issues/828)

This commit adds a summary field to the markdown file. A new form has been added to the submit page, along with an additional field being added to the proposal detail view.

Adding the summary field will require changes in the backend policy, along with updating Politeia on dcrdocs.

### DIFF
--- a/src/actions/api.js
+++ b/src/actions/api.js
@@ -333,11 +333,12 @@ export const onSubmitProposal = (
   username,
   name,
   description,
-  files
+  files,
+  summary
 ) =>
   withCsrf((dispatch, getState, csrf) => {
-    dispatch(act.REQUEST_NEW_PROPOSAL({ name, description, files }));
-    return Promise.resolve(api.makeProposal(name, description, files))
+    dispatch(act.REQUEST_NEW_PROPOSAL({ name, description, files, summary }));
+    return Promise.resolve(api.makeProposal(name, description, files, summary))
       .then(proposal => api.signProposal(loggedInAsEmail, proposal))
       .then(proposal => api.newProposal(csrf, proposal))
       .then(proposal => {
@@ -348,7 +349,8 @@ export const onSubmitProposal = (
             userid,
             username,
             name,
-            description
+            description,
+            summary
           })
         );
         resetNewProposalData();
@@ -368,11 +370,12 @@ export const onSubmitEditedProposal = (
   name,
   description,
   files,
-  token
+  token,
+  summary
 ) =>
   withCsrf((dispatch, _, csrf) => {
-    dispatch(act.REQUEST_EDIT_PROPOSAL({ name, description, files }));
-    return Promise.resolve(api.makeProposal(name, description, files))
+    dispatch(act.REQUEST_EDIT_PROPOSAL({ name, description, files, summary }));
+    return Promise.resolve(api.makeProposal(name, description, files, summary))
       .then(proposal => api.signProposal(loggedInAsEmail, proposal))
       .then(proposal => api.editProposal(csrf, { ...proposal, token }))
       .then(proposal => {
--- a/src/actions/app.js
+++ b/src/actions/app.js
@@ -41,10 +41,11 @@ export const onSetReplyParent = (
     dispatch(act.SET_REPLY_PARENT(parentId)),
     dispatch(reset("form/reply"))
   ]);
-export const onSaveNewProposal = ({ name, description, files }, _, props) => (
-  dispatch,
-  getState
-) =>
+export const onSaveNewProposal = (
+  { name, description, files, summary },
+  _,
+  props
+) => (dispatch, getState) =>
   dispatch(
     onSubmitProposal(
       props.loggedInAsEmail,
@@ -52,12 +53,13 @@ export const onSaveNewProposal = ({ name, description, files }, _, props) => (
       props.username,
       name.trim(),
       description,
-      files
+      files,
+      summary
     )
   ).then(() => sel.newProposalToken(getState()));
 
 export const onEditProposal = (
-  { name, description, files },
+  { name, description, files, summary },
   _,
   props
 ) => dispatch =>
@@ -67,18 +69,26 @@ export const onEditProposal = (
       name.trim(),
       description,
       files,
-      props.token
+      props.token,
+      summary
     )
   );
 
-export const onSaveDraftProposal = ({ name, description, files, draftId }) => {
+export const onSaveDraftProposal = ({
+  name,
+  description,
+  files,
+  draftId,
+  summary
+}) => {
   resetNewProposalData();
   return act.SAVE_DRAFT_PROPOSAL({
     name: name.trim(),
     description,
     files,
     timestamp: Date.now() / 1000,
-    draftId
+    draftId,
+    summary
   });
 };
 
--- a/src/actions/tests/api.test.js
+++ b/src/actions/tests/api.test.js
@@ -32,6 +32,7 @@ describe("test api actions (actions/api.js)", () => {
   const FAKE_CSRF = "fake_csrf_token";
   const FAKE_PROPOSAL_NAME = "Fake prop name";
   const FAKE_PROPOSAL_DESCRIPTION = "Fake prop description";
+  const FAKE_PROPOSAL_SUMMARY = "Fake prop summary";
   const FAKE_PROPOSAL_TOKEN = "fake_prop_token";
   const FAKE_PROPOSAL_VERSION = "2";
   const FAKE_COMMENT = "fake comment text";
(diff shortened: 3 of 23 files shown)


xaur:
Can you please paste a short snippet of how the field is encoded in the markdown file (with some surrounding context)?

camus-code:
@xaur 
This first snippet is where the newProposal structure is created from the markdown file. I separated the summary from the body with `\r` as to differentiate between the two.

```
export const makeProposal = (name, markdown, attachments = [], summary) => ({
  files: [
    convertMarkdownToFile(name + "\n" + markdown + "\r" + summary),
    ...(attachments || [])
```

I then added a new helper function (seen below) that grabs the summary. This function is then used 
[here](https://github.com/decred/politeiagui/blob/2ea3cc8f93b163272034cf0fad7157a803eaaf6a/src/components/ProposalDetail/ProposalDetail.js#L180) to pull the summary into ThingLink.
```
export const getSummaryFromIndexMd = file => {
  const text = atou(file.payload);
  const summary = text.substring(text.indexOf("\r"));
  //If statement to hide empty summaries for pre-existing proposals
  if (summary.length > 1) {
    return summary;
  } else {
    return null;
  }
};
```

xaur:
@camus-code sorry I should be more clear, I mean the markdown snippet -- how a piece of markdown that contains the field looks?

Also, using `\r` is very weird. Shouldn't it be `\n`?

camus-code:
@xaur sorry for the confusion. I've attached a snippet below.

As for the escape character, I wasn't too fond of the using `\r` either, but the problem lies in trying to determine where the body ends and the summary begins. Since newlines can be added into the summary field, it's hard to differentiate between the two based on an index of `\n`, hence why I used the other character. I'm obviously open to suggestions though

camus-code:
Sorry forgot the snippets @xaur. Let me know if you were looking for something different than this 

screen shot 2019-02-05 at 2 54 04 pm


screen shot 2019-02-05 at 2 54 14 pm

xaur:
That is not Markdown, can you please paste the snippet of Markdown that contains the summary and some surrounding context? I'd like to see how the data is stored.

```
Here, in a code block.
```

`\r` looks like an ugly hack to me. If multiline text is required in the summary, it is better to design it properly while Politeia is in early stages. Later it will be much more difficult to fix legacy warts, so better not to introduce them in the first place.

First option is to demand that the user encodes newlines explicitly: "Text\nNext line".

Second is to explicitly encode metadata as YAML (for example), so the document starts with:

```
---
title: Proposal title!
summary: Single-line text
---

And here's the proposal text.
```

or

```
---
title: Proposal title!
summary: |
  Multi-line
  Text
---

And here's the proposal text.
```

Third option is simplest and I would go with it - forbid newlines in the summary. Force it to single line text. I don't see any reason a short summary must be multiline.

Note I'm not a Politeia dev, may need input from them.

camus-code:
Here is the markdown snippet

```
Testing Proposal Title
Testing Proposal Body Testing Proposal Body Testing Proposal Body Testing Proposal Body Testing Proposal Body Testing Proposal Body Testing Proposal Body Testing Proposal Body Testing Proposal Body Testing Proposal Body Testing Proposal Body Testing Proposal Body Testing Proposal Body Testing Proposal Body

Testing Proposal Body Testing Proposal Body Testing Proposal Body Testing Proposal Body Testing Proposal Body Testing Proposal Body Testing Proposal Body Testing Proposal Body Testing Proposal Body Testing Proposal Body Testing Proposal Body Testing Proposal Body Testing Proposal Body Testing Proposal Body

Proposal summary proposal summary 
Proposal summary proposal summary 
```
My reasoning for using `\r` was that in the markdown [spec](https://spec.commonmark.org/0.25/#line-ending), line endings can be defined by either newline (`/n) or a carriage return (`/r`). 

Forbidding new lines in the summary is definitely the simpler approach, but I do like the idea of having a ~250 character space being a little more flexible.

xaur:
~250 chars is nice limit, agreed. Forbidding newlines may even stimulate people to write better summaries.

Looking at the snippet, is it correct that title is at the top and summary is at the bottom? So the format looks like

```
title + no newline
body
newline
summary
```

camus-code:
Perhaps you're right about it encouraging a more concise proposal. I would be interested in hearing from some of the other GUI devs. And yeah you are correct about the format. Switching the summary to the top might make more sense now that I think about it though

xaur:
If title is at the top and summary at the bottom, these are two pieces of metadata stuffed into the document in a hacky ad-hoc way. This will end bad.

Honestly even the way the title alone is encoded (without the extra newline after it) is poor, this would be better:

```
title

proposal begins
```

camus-code:
Hmm ok I'll move some things around and get back to you with a new snippet.  

@fernandoabolafio @thi4go @tiagoalvesdulce take a look over this whenever you get the chance

xaur:
I wonder how feasible it is to change to proper explicit metadata encoding (per above comments) now? If [proposal identifiers](https://github.com/decred/politeia/issues/560) or something else is ever planned, that would need metadata too.

fernandoabolafio:
I like having a summary for the proposal, my major question is why the summary is placed at the bottom of the proposal and not right after the title? 
If I am looking at the raw proposal (on GitHub) for instance, I have to scroll the proposal all the way down to find out that there is a summary in there. I think the whole point of having short text which can be quickly noted.

camus-code:
No reason honestly. I added it at the end for simplicity's sake when writing the code but it definitely belongs after the title. I'll address that in the next commit

camus-code:
@fernandoabolafio Fixed that with the latest commit. Also updated the tests to reflect the change in order

xaur:
How does the markdown sample look after the fix?

Also, a bit out of scope of this PR but this PR triggers a more general question: why embed the title and summary in markdown at all, when title is already saved in metadata ([example](https://github.com/decred-proposals/mainnet/blob/master/5431da8ff4eda8cdbf8f4f2e08566ffa573464b97ef6d6bae78e749f27800d3a/2/00.metadata.txt))?

fernandoabolafio:
@xaur about your question, the title is required in the markdown because behind the scenes, the title which goes in the metadata is always inferred from the markdown and also in the GUI, the title is extracted from the proposal markdown. 
I agree, however, that this is a valid question if we could input the title straight into the metadata instead of parsing the markdown file for it.

xaur:
@fernandoabolafio thanks, but if .md is primary source of truth for the title, then the question reverses: why is title _also_ stored in meta?

I can see some reason - the file is signed. Is metadata also covered by author's signature? In any case, the title and the summary field are two pieces of metadata that comes from the _author_, hence must be under his signature. Two ways to achieve it are to either stuff it into the file or put it separately.

My concern is "title is that first line in the file delimited by a single \n" is hacky implicit way to insert meta into the document. Attempting to insert even more meta this way will add even more warts. And then we have a problem that we can't "migrate" these documents because they are signed by authors.

s-ben:
Late to the party here, and perhaps out of scope for this PR, but this Issue just collided with another I'm working on. Namely that I'm unable to use the [politeia_verify](https://github.com/decred/politeia/tree/master/politeiad/cmd/politeia_verify) tool to verify censorship of a test proposal I submitted on testnet. If I input the proposal Markdown file I uploaded using politeiagui into politeia_verify, the Merkle roots don't match. I don't know Javascript, but looking at the code that encodes proposals, it appears that politeiagui is calculating the Merkle root differently than politeia_verify. And I'm wondering if this is due to how we encode/extract metadata? 

To prove censorship of content, the author will have to provide all raw files used to create the proposal. Adding metadata through the UI is therefore problematic currently, because the user doesn't have that metadata in a file they can input into politeia_verify. Proposals can also be submitted via the command-line tool politeiawwwcli, which presumably would need to be updated to accept metadata input. All this seems to point to having the metadata in the Markdown...

I could be mistaken here, not trying to create more work. Also, if anyone has other theories about politeia_verify I'm all ears.

xaur:
Interesting info and is relevant to this PR @s-ben, but it might be also useful as a standalone issue.

s-ben:
Yeah kind of just fishing for info too. Don't let me stop the PR train. Planning to open an Issues on /Politeia once I verify there is indeed a problem. Also, ready to include the summary field in the Pi docs whenever it's ready: https://github.com/decred/dcrdocs/issues/828

PR #1014, opened 2019-02-11
fix comment indentation index.css

### LINKED ISSUE #1011: Indentation issue with comments up/down buttons
The comments indentation around up/down votes need to be polished. It's currently hard to identify what is the comment level. 
The comment body should be entirely aligned at the right side of the up/down buttons.
screen shot 2019-02-06 at 15 49 46

camus-code:
Resolves #1011


camus-code:
@thi4go Yeah definitely! That message was a bit obscure

PR #1018, opened 2019-02-15
Markdown tab preview

### LINKED ISSUE #1017: Markdown full page preview
Dustin is currently writing a marketing proposal and ran into the issue of not being able to preview what his markdown tables would look like in the normal, full page proposal details view.

Is there a reason that the current markdown editor is setup the way it is and do you think it would be beneficial to have a full page preview instead?  Maybe a preview tab like Github has.

victorgcramos:
Fixes #1017

### DIFF
--- a/src/components/snew/SubmitPage.js
+++ b/src/components/snew/SubmitPage.js
@@ -36,6 +36,10 @@ const normalizer = (value, previousValue) => {
   return files;
 };
 
+const ToggledMarkdown = props => (
+  <MarkdownEditorField input={props.input} toggledStyle={true} />
+);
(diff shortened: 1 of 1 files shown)


fernandoabolafio:
I want to get more feedback before we move this code in. I tagged others who have written long proposals for politeia in the issue. Let see what are their preference.

fernandoabolafio:
Okay, we have one more vote for the "toggled" input. One small comment though: can you align the write/preview buttons with the right side of the input (same as the comments)? 

screen shot 2019-02-18 at 18 30 56

PR #1024, opened 2019-02-22
WIP: Add toggle to use CMS Registration instead

alexlyp:
This is a basic PoC to add/replace some new contractor management components and forms.  

I am still debating various ways we could "check" whether we are running Politeiagui in "cms mode" or not.  In this PoC I'm merely setting the option true/false in the selectors.  When we settle on a solution that will be replaced properly.

One possible idea I had was to add isCMS into the policy from politeiawww.  So politeiagui is started/launched the same way in both cases and just depending on which www "version" it is connected to it will show/hide various functionality.  Feel free to drop your thoughts on the topic here.

I've also added a rough stubbing in of a new route/form/action/lib/selector for admin invitation functionality.  If this general form looks acceptable we can start to fill in the other required forms for submitting invoices/logging work special features in CMS.


fernandoabolafio:
I like the idea of "toggling" the politeiagui mode based on www, this is pretty much the same thing we current do for setting the network (testnet/mainnet). In fact we could use the same route to add the `isCMS` boolean, see:
https://github.com/decred/politeia/blob/master/politeiawww/api/v1/v1.go#L458

I haven't run this code on my local end but the form code looks good to me.

alexlyp:
I've updated this PR with the addition of @fernandoabolafio's PoC of using HOCs to decide which connector to use depending on sel.isCMS

Issue #1022, opened 2019-02-20
New comment sorting algorithm


thi4go:
Currently, comments are being sorted by New as a default. This entails that good discussions started early on in the proposal gets burried down, potentially not being seen by new coming users. It was stated by @behindtext in riot that top comments would be best being sorted from old to new when going from top to bottom.

Reddit also implements a new sorting algo that aimed to increas "fairness" in the comment's sorting by blending comment votes and age, that could also improve how pgui sort comments. Initially, a quick improvement to the current status would be to make sorting from Top the default.

Quoting @RichardRed0x about the new reddit algo:

> yeah it's based on a Wilson score confidence interval iirc
> makes a guess about what the comments' scores (approval rates) would be by taking the lower > bound of a confidence interval around its current score, more votes means the interval is smaller > means the lower bound is higher.

Blog post about the new reddit sorting algo: [reddit link](https://redditblog.com/2009/10/15/reddits-new-comment-sorting-system/)

Here is a link to the discussion in riot: [riot link](https://matrix.to/#/!VFRvyndKpzcLrVslQD:decred.org/$15506331673048Eally:decred.org?via=decred.org&via=matrix.org&via=zettaport.com)

RichardRed0x:
To follow up on that discussion, I think switching to "Top" sorting method as the default would be a quick and easy move that improves the ordering of comments on politeia. Reddit used this as its default for some time, then developed the "Best" sort because of an issue whereby early comments would become locked in to the high-visibility spaces (appearing at the top of the list means the comment gets more votes so is more likely to remain at the top and get more votes, a feedback loop).

From what I have seen on politeia so far, I think Top could work quite well. People seem to be reading and voting fairly deep into the page. While I do think the kind of sorting reddit uses for "Best" is good and would be worth implementing, we may not be at a scale where it is really needed yet.

Also, reddit's Best algorithm relies on the upvote/downvote ratio for a comment as the basis for its ranking. So far Politeia has a much lowe rate of downvoting than reddit does, which could make this algorithm less useful.

xaur:
Please whatever is decided for the default sorting, I'd like to have "oldest first" or "newest first" as a personal default. The chronology of comments is important for my analysis and I hate when content is rearranged without request.

xaur:
In the special case of Politeia at its current (immature) stage of development, sorting oldest to newest by default has some merit: people who comment earlier are likely to be more active participants.

xaur:
[...]

PR #1034, opened 2019-02-28
Persist comment sort option

fernandoabolafio:
This PR:
1. Set "Top" as the default comment sort option.
2. Persist the comment sort option in the local storage.

1 and 2 were mentioned on #1022


fernandoabolafio:
Hold on, I will use the URL to persist the sort option instead of local storage.

fernandoabolafio:
Done! Now the comment sort option is set in the query string and update on the page load if a valid "sort" option is specified within the query string.
I also encapsulated some handful functions to deal with the query string. 
@tiagoalvesdulce can you take a look?

PR #1036, opened 2019-03-08
Updates README

s-ben:
Took a pass at fleshing out the README. Mainly just adding basic explanatory text so a new contributor will have an easier time. I'm a total node and JS noob though, so this definitely needs a thorough review. Open to feedback/criticism on my rewording/restructuring.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,10 +9,10 @@ Politeiagui uses node.js and the yarn package manager.
 
 - [node](https://nodejs.org/en/download/)
 - [yarn](https://yarnpkg.com/en/)
+- [politeiawww](https://github.com/decred/politeia)
 
+    **Note:**  Politeiagui can be run as a standalone application that uses mock APIs to simulate communicaiton with the Politeia backend (useful if you're only doing UI changes). In this case, `politeiawww` isn't needed. However, be aware that the Mock APIs are not always up-to-date with the latest changes. 
 
-
-Politeiagui can be built and run as a standalone application that uses mock APIs to simulate communicaiton with the Politeia backend (useful if you're only doing UI changes). However, if you want run the the full Politeia application, you'll first need to build and run `politeiawww`, Politeia's webserver. Instructions for doing so can be found [here](https://github.com/decred/politeia)
 
 
 ## Development
@@ -25,17 +25,19 @@ Politeiagui can be built and run as a standalone application that uses mock APIs
 
 1. Build and run locally.
 
-    **Mock APIs**
-
-    To run politeiagui using mock APIs that do not communicate with `politeiawww` (useful for working on UI changes only), run the following command:
-    
-    `yarn && yarn uidev`
-    
     **Against politeiawww**
     
     To run politeiagui against `politeiawww`, make sure `politeiawww` is running and use the following command:
     
     `yarn && yarn start`
+
+    **Mock APIs (WARNING: may be out of date)**
+
+    To run politeiagui using mock APIs that do not communicate with `politeiawww` (useful for working on UI changes only), run the following command:
+    
+    `yarn && yarn uidev`
+    
+
     
 
 ## Production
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,52 +1,79 @@
...
+
+It is necessary for politeiagui to connect to the testnet or mainnet in order to get the last block height, and determine the status of votes. 
 
-    This way it is possible to determine when a vote has finished or not.
 ---
 
 ## Testing 
 
-This code has currently unit tests for the core logic. Currently covering lib files (`/src/lib/*`) and actions files (`/src/actions/*`).
+Politeiagui currently has unit tests for its core logic. Test coverage is provied for lib files (`/src/lib/*`) and actions files (`/src/actions/*`).
```
fernandoabolafio: We currently test `actions/*`, `lib/*`, `reducers/*` and `selectors/*`. You can check the test coverage [here](https://github.com/decred/politeiagui/blob/master/package.json#L65).

README.md
```diff
@@ -1,52 +1,79 @@
...
+
+
+1. Clone this repository
+
+    `https://github.com/decred/politeiagui.git`
+    
+
+1. Build and run locally.
+
+    **Mock APIs**
```
fernandoabolafio: We don't have the Mock API's fully up to date. Worth to check first what is the current real usability of it to decide to really include it in the README or at least add a "deprecated" tag for it.

README.md
```diff
@@ -1,52 +1,79 @@
...
-### Requirements
 - [node](https://nodejs.org/en/download/)
 - [yarn](https://yarnpkg.com/en/)
 
-### With mock APIs
 
-    yarn && yarn uidev
 
-### Against politeiawww
+Politeiagui can be built and run as a standalone application that uses mock APIs to simulate communicaiton with the Politeia backend (useful if you're only doing UI changes). However, if you want run the the full Politeia application, you'll first need to build and run `politeiawww`, Politeia's webserver. Instructions for doing so can be found [here](https://github.com/decred/politeia)
```
fernandoabolafio: We don't have the Mock API's fully up to date. Worth to check first what is the current real usability of it to decide to really include it in the README or at least add a "deprecated" tag for it.

README.md
```diff
@@ -1,52 +1,79 @@
...
 ## Testing 
 
-This code has currently unit tests for the core logic. Currently covering lib files (`/src/lib/*`) and actions files (`/src/actions/*`).
+Politeiagui currently has unit tests for its core logic. Test coverage is provied for lib files (`/src/lib/*`) and actions files (`/src/actions/*`).
+
 Relevant libraries/modules used for testing are:
+
  - [Jest](https://facebook.github.io/jest/docs/en/getting-started.html) - Javascript testing
  - [fetch-mock](http://www.wheresrhys.co.uk/fetch-mock/api) - Mock http requests
  - [redux-mock-store](https://github.com/dmitry-zaets/redux-mock-store) and [redux-actions-assertions](https://github.com/redux-things/redux-actions-assertions) for testing redux actions.
```
fernandoabolafio: Would be nice to include the testing command here too: `yarn test`

README.md
s-ben: If the Mock APIs are more of a tool for regular contributors, which are prone to falling out of date, I would either exclude it from the README altogether, or include it, but not prominently and with a warning. I've gone with the latter for now, as we want to have these mentioned somewhere, if not the README. But can remove if you think it's best to just have people install `politeiawww`.

README.md
s-ben: Included. I assume this runs the full suite of unit tests, so have gone with the below text for now:

> Politeiagui has unit tests for its core logic. To run all unit tests run the command:

    yarn test

s-ben:
@fernandoabolafio, thanks for the quick review. Have made some updates to address your comments. This looks good to go, unless maybe you want to frame the Mock APIs differently. If they really aren't used, perhaps we can just take them out or put a `depricated` warning on them.

PR #1051, opened 2019-03-21
fix incorrect count of comments

### LINKED ISSUE #1045: New comments do not update Proposal Comment counter
The image shows "all 11 comments" on the comments list, and this number gets increased as comments are added. However, the "8 comments" (below "ABANDON" button) doesn't.

When you refresh the page, the number is changed.

Captura de Tela 2019-03-18 às 19 55 07

camus-code:
Closes #1045

### DIFF
--- a/src/components/snew/ThingLink/ThingLinkProposal.js
+++ b/src/components/snew/ThingLink/ThingLinkProposal.js
@@ -186,6 +186,15 @@ class ThingLinkComp extends React.Component {
       }
     };
 
+    // We use same component to render comment length on detail view and list view
+    // listOrDetailView checks if proposal is expanded
+    const detailViewComments = comments.length;
+    const listViewComments = numcomments;
+    const listOrDetailViewComments =
+      expanded && detailViewComments > listViewComments
+        ? detailViewComments
+        : listViewComments;
+
     return (
       <div
         className={`thing thing-proposal id-${id} odd link ${censoredorAbandoned()}`}
@@ -592,8 +601,8 @@ class ThingLinkComp extends React.Component {
                   data-event-action="comments"
                   href={`${url}?comments=true`}
                 >
-                  {numcomments || ""}{" "}
-                  {numcomments <= 1 ? "comment" : "comments"}
+                  {listOrDetailViewComments || ""}{" "}
+                  {listOrDetailViewComments <= 1 ? "comment" : "comments"}
                 </Link>
               </li>
             ) : null}
--- a/src/connectors/proposal.js
+++ b/src/connectors/proposal.js
@@ -7,7 +7,7 @@ import compose from "lodash/fp/compose";
 import { arg, or } from "../lib/fp";
 import * as sel from "../selectors";
 import * as act from "../actions";
-import { buildCommentsTree } from "../lib/snew";
+import { buildCommentsTree, buildSetOfComments } from "../lib/snew";
 
 const proposalConnector = connect(
   sel.selectorMap({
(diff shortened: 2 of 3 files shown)


src/selectors/app.js
```diff
@@ -104,6 +104,7 @@ export const getEditProposalValues = state => {
 
   const files = name ? getNotMarkdownFile(state) : [];
   const description = name ? getTextFromIndexMd(getMarkdownFile(state)) : "";
+  console.log(name, description);
```
victorgcramos: You left a log here!

camus-code: thanks for the catch! it's removed now

src/connectors/thingLink.js
```diff
@@ -1,9 +1,11 @@
+import React from "react";
 import { connect } from "react-redux";
 import * as sel from "../selectors";
 import * as act from "../actions";
 import get from "lodash/fp/get";
 import compose from "lodash/fp/compose";
 import { arg } from "../lib/fp";
+import { buildCommentsTree, buildSetOfComments } from "../lib/snew";
```
fernandoabolafio: Why do you need to build the comments tree just to get the number of comments? Wouldn't it be just as simple as getting the length of the proposal comments? See my comments for ThingLink.js.

src/components/snew/ThingLink.js
```diff
@@ -86,6 +86,7 @@ class ThingLinkComp extends React.Component {
       otherFiles,
       review_status,
       numcomments,
+      numofcomments,
```
fernandoabolafio:
```suggestion

```

src/components/snew/ThingLink.js
```diff
@@ -188,6 +189,12 @@ class ThingLinkComp extends React.Component {
       }
     };
 
+    // Fetches inital comment count from api to prevent flickering
+    const correctNumComments =
```
fernandoabolafio:
```suggestion
    const correctNumComments = comments.length;
```

src/lib/api.js
```diff
@@ -334,14 +334,14 @@ export const proposalSetStatus = (email, csrf, token, status, censorMsg) =>
...
+  POST("/proposals/new", csrf, proposal)
+    .then(({ response: { censorshiprecord } }) => ({
       ...proposal,
       censorshiprecord,
       timestamp: Date.now() / 1000,
       status: PROPOSAL_STATUS_UNREVIEWED
-    })
-  );
+    }))
+    .then(console.log({ ...proposal }));
```
fernandoabolafio: Remove this

src/connectors/thingLink.js
camus-code: Upon closer inspection you're right. I mistakenly believed that `numcomments` only updated upon page refresh, but getting the length of this prop returns the correct result. Addressing this now

fernandoabolafio: Just so you do not get confused, `numcomments` is a field returned with the proposal data which is indeed updated upon page refresh. The prop `comments` which I am suggesting you to use to check the length is updated every time a new comment is received.

camus-code: Ahh okay I appreciate the clarification. I had forgotten we also had the prop `comments` and assumed the tree had to be built to get the length.

Updated PR is up :)

src/selectors/app.js
tiagoalvesdulce: You left a log here

camus-code:
@fernandoabolafio rebased :)

PR #1052, opened 2019-03-25
fix external link handler and url preview

### LINKED ISSUE #1047: External link warning modal issues
1.
One of the issues highlighted in https://github.com/decred/politeiagui/issues/960 is back. The modal should close if the back button is pressed. 

2.

I also found that if the URL of the website contains a hash "#" (fragment identifier) then anything after that is not shown in the warning. This needs to be fixed since in many websites the content is dynamically loaded based on the fragment identifier. 

image

camus-code:
Closes #1047 

Fixes onClose behavior when path changes for modals. Also updates URL preview in external link modal to show entire url, including query strings and fragment identifiers

### DIFF
--- a/src/components/Modal/ModalStack.js
+++ b/src/components/Modal/ModalStack.js
@@ -13,7 +13,7 @@ class ModalStack extends React.Component {
 
   componentDidUpdate(prevProps) {
     const { openedModals, location, closeAllModals } = this.props;
-    const welcomeModal = openedModals.filter(
+    const welcomeModalOpened = openedModals.find(
       modal => modal.type === WELCOME_MODAL
     );
     const { modals } = this.state;
@@ -34,7 +34,10 @@ class ModalStack extends React.Component {
         else document.querySelector("body").style.overflowY = "scroll";
       });
     // closes modals if user path has changed except when it's the user's first access
-    if (location.pathname !== prevProps.location.pathname && !welcomeModal) {
+    if (
+      location.pathname !== prevProps.location.pathname &&
+      !welcomeModalOpened
+    ) {
       closeAllModals();
     }
   }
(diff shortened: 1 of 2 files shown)


src/components/Modal/ModalStack.js
```diff
@@ -16,6 +16,7 @@ class ModalStack extends React.Component {
     const welcomeModal = openedModals.filter(
```
fernandoabolafio: I think it could be better written like this:

```js
const welcomeModaIOpened = openModals.find(modal => moda.type === WELCOME_MODAL)
```

See next comment

src/components/Modal/ModalStack.js
```diff
@@ -34,7 +35,7 @@ class ModalStack extends React.Component {
         else document.querySelector("body").style.overflowY = "scroll";
       });
     // closes modals if user path has changed except when it's the user's first access
-    if (location.pathname !== prevProps.location.pathname && !welcomeModal) {
+    if (location.pathname !== prevProps.location.pathname && noWelcomeModal) {
```
fernandoabolafio:
```suggestion
    if (location.pathname !== prevProps.location.pathname && !welcomeModalOpened) {
```

src/components/Modal/ModalStack.js
```diff
@@ -16,6 +16,7 @@ class ModalStack extends React.Component {
     const welcomeModal = openedModals.filter(
       modal => modal.type === WELCOME_MODAL
     );
+    const noWelcomeModal = welcomeModal.length === 0;
```
fernandoabolafio:
```suggestion
```

src/components/Modal/ModalStack.js
camus-code: Good suggestion to use .find that's a lot cleaner than filter. PR is up

Issue #1060, opened 2019-03-31
User edit invoices


fernandoabolafio:
EditInvoice allows users to update their invoices based on comments by
administrators.

thi4go:
Working on this one!

fernandoabolafio:
closed on #1063

PR #1063, opened 2019-04-02
[CMS] Adds edit invoice feature

thi4go:
This PR still needs some testing, but implements edit invoice feature.

#1060


alexlyp:
Can you rebase this please?

thi4go:
Hey mates @alexlyp @fernandoabolafio this PR is ready to be reviewed.  Can you guys give it a go? I also had to check something with fernando but riot is not colaborating today 😅

alexlyp:
Sure thing.  Can you take a look at #1084 ?

alexlyp:
So looks like it updated properly for editing Contractor fields and any rows that were previously there, but adding new rows wasn't working and it kept saying it was a duplicate invoice (if i didn't change anything else).

alexlyp:
I also noticed that the edit button needs to be only shown in these situations:
- Only show for invoices in New, Updated or Disputed statuses.  Anything else it shouldn't be able to edit.
- Only show for invoices owned by the current user (this may be tougher to do).

alexlyp:
Also on Edit I noticed that a previously blacked out cell for expenses on a Labor (type 1) line item became editable.

alexlyp:
Oh and a "cancel" button next to Update would be great as well.

thi4go:
Right, working on the changes! Thanks for the review alex.

So is the `Expense` cell supposed to be Readonly at all times? I tested here and I'm actually getting that when creating new invoices as well. When I add a row, the Readonly cell shifts to the left it seems.

src/components/snew/RegisterForm.js
```diff
@@ -22,7 +22,7 @@ const RegisterForm = ({
   handleSubmit,
   isCMS
 }) =>
-  loggedInAsEmail ? null : isCMS ? (
```
fernandoabolafio: I don't think the diffs in this file belongs to this PR, right? Maybe a rebasing issue.

src/validators/signup.js
```diff
@@ -13,10 +13,7 @@ const validate = (policy, values, isCMS) => {
       !isRequiredValidator(values.verificationtoken) ||
       !isRequiredValidator(values.username) ||
       !isRequiredValidator(values.password) ||
-      !isRequiredValidator(values.password_verify) ||
```
fernandoabolafio: Looks to me that this diff doesn't belong to this PR. Maybe a rebasing issue?

src/selectors/app.js
```diff
@@ -63,9 +62,11 @@ import {
 } from "../constants";
 import {
   getTextFromIndexMd,
+  getTextFromJsonToCsv,
   countPublicProposals,
   isProposalApproved
 } from "../helpers";
+import { convertLineItemsToGrid } from "../components/InvoiceDatasheet/helpers";
```
fernandoabolafio: I think you'll need to make this function more robust in order to update the datasheet cells accordingly. There are some rules which you can see when you're creating new invoices, such as:
- if type cell === 1 -> expense cell is read-only
- if type cell === 2 or 3 -> labor cell is read-only 

If this function is called with readOnly enable, this check is not necessary. Otherwise, we need to make sure this happens when populating the table.

There is already a function which updates a grid based on the value of a given cell, maybe we can integrate that into this function. See: https://github.com/decred/politeiagui/blob/master/src/components/InvoiceDatasheet/helpers.js#L294.

thi4go: Nice Fernando, didn't know about those restrictions, thanks for clarifying. Right now I pushed an easy fix to this problem, without using the `processCellsChange` function, though I did try to place it somewhere within that function and wasnt quite able to...

src/components/snew/RegisterForm.js
thi4go: Trying to remove this from the PR, not sure how to exactly 🤔

src/validators/signup.js
thi4go: Trying to remove this from the PR, not sure how to exactly 🤔

alexlyp:
Everything looks good to me after testing.  Will let @fernandoabolafio do the final approval for any issues codewise. Great job!

thi4go:
Hey @fernandoabolafio, thanks for the thorough review mate. So I investigated it a bit further, and I stumbled into some weird behaviour. So there is two blocks on this form, the first one with input fields (name, contact, rate...) and the grid with line items. I noticed that:

- if I edit information on the input fields AND the line items, only the line items gets updated correctly;
- the input fields gets updated correctly only if you don't change line items on the same request;

Do you guys have any suspicion on why this is happening? @fernandoabolafio @alexlyp I'll keep investigating further anyhow.

fernandoabolafio:
Hey, @thi4go here is the video in another hosting service: https://sendvid.com/pck7wqaa.

About this I think is probably due to misuse of the connectors, but I haven't noticed that while testing the feature. The form updating looks fine to me as I type the new values. What I noticed as errored is what I captured in the video: the updated values in the form are not being sent to the API.
> Do you guys have any suspicion on why this is happening? @fernandoabolafio @alexlyp I'll keep investigating further anyhow.

thi4go:
Hey @fernandoabolafio, the form is fixed now, thanks for the hint, it was indeed a problem with the connector. 👍

PR #1048, opened 2019-03-21
Improve CMS customization

tiagoalvesdulce:
Now that I played with this a little, I'm opening this WIP so we can discuss some points and open it to collaboration.

* Politeia's presentational components use a lot of props with the word proposal eg. `fetchProposal`, `proposals`, `proposalCount`, etc. passed by connectors. IMO, to reduce the number of if/else inside these components we should rename all these variables to "registers" or "data". This way we could create custom connectors easily without touching presentational components. What do you think?

PR #1064, opened 2019-04-02
Remove duplication from RegisterForm.js

jholdstock:
#1048 added four new fields to the Registration form - Invite Token, Full name, Location and Extended Pubkey.

This was implemented by duplicating the source of the existing registration form, inserting the new fields into the duplicated form, and then switching between the forms using `isCMS`.

This removes that duplication, and instead inserts the new fields into the existing registration form using `isCMS`.

### DIFF
--- a/src/components/snew/RegisterForm.js
+++ b/src/components/snew/RegisterForm.js
@@ -120,54 +120,6 @@ const RegisterForm = ({
             />
           ) : null}
         </div>
-        {isCMS && (
-          <div className="c-form-group">
-            <label className="screenreader-only" htmlFor="user_reg">
-              name:
-            </label>
-            <Field
-              name="name"
-              placeholder="Full Name"
-              component="input"
-              type="text"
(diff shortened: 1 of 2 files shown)


tiagoalvesdulce:
Thanks for this PR @jholdstock. It looks good but broke the spacing between `Verify Password` and `Full Name` fields. See:

Screen Shot 2019-04-02 at 22 43 06

fernandoabolafio:
Name, location and extend pubkey won't be part of the signup form anymore (see https://github.com/decred/politeia/pull/769) so I think this PR makes even more sense now. Since that, for CMS we will be adding only "Invite token". Do you want to remove the outstanding fields in this PR?

fernandoabolafio:
@jholdstock if this PR is meant to close https://github.com/decred/politeiagui/issues/1074, please reference in the description so it can be closed automatically when we merged this PR.

Issue #1021, opened 2019-02-19
Manually mark new comments as read


xaur:
Governance discussions are serious and I'd like to not miss anything. I'd like to avoid an unintended loss of the "unread" marks by e.g. an accident page reload. This can be prevented if the "unread" marks are removed manually, e.g. with a click on the comment area (similar to reddit admin feed).

xaur:
Just had this problem yesterday. I opened a proposal to read new comments and had them highlighted. Then I couldn't finish due to interruptions and closed the browser. As all new comments were marked as read automatically, today I can't continue my reading of unread comments because I lost track of them.

I think the `unread -> read` transition should happen by click anywhere in the comment box since it's a common operation. Also there must be a way to `read -> unread` something I'd like to re-read tomorrow, but this is less common and can be done with tiny `mark unread` link-buttons near the `permalink` and `reply`.

xaur:
I'm destroyed. I opened a proposal to read newly highlighted comments. Suddenly a paralyzing "your session will expire in less than a minute" and I couldn't do anything. Now all highlighting is gone and I need to do the scrolling and looking at comment IDs again to figure out what I didn't read.

Filed a bug to not do any disturbing automatic view changes as #1031, but manual read state management would prevent state loss too.

xaur:
Insight: proposal owners can benefit from this feature. Current UX does not fully facilitate replying to all questions. Unread state is lost on page reload. Email does not solve it fully either: email notifications do not contain comment body and do not allow to reply via email.

With manual read state control, proposal owners have an extra tool to never miss a question. They can keep unread markers on comments until they are fully answered.

xaur:
Another insight: manual "read" marks can yield interesting data for analysis. For example, count how many people read a given comments and compare it to its score (suggested [here](https://matrix.to/#/!MgQoetFiyjrHAywokv:decred.org/$1550967899428957ycgNm:matrix.org)).

Upvoting or downvoting a comment should set the "read" mark.

victorgcramos:
I can work on that!

xaur:
@victorgcramos some thoughts on your [question](https://matrix.to/#/!VFRvyndKpzcLrVslQD:decred.org/$1553104433341207cSULd:matrix.org) in chat.

[...]

PR #1066, opened 2019-04-02
Feature: mark new comments as read manually

victorgcramos:
This commit allows users to manually mark comments as read, as we discussed on #1021

- adds "mark as read" option to each comment
    - read comments can be marked as unread
- style changes for supporting the new design for new comments
- adds www calls for persisting read comments data for each user
- clicking on unread comment's body sets it as read
- like action toggles read comment state for comment


victorgcramos:
#### [TODO] Set comments as read when:
- [ ] Upvoting/downvoting
- [ ] Reply
- [ ] Click on comment's box (?)
#### Done
- [x] Send new comment

xaur:
> Click on comment's box

Yes please, it's very fast to target because the box is big enough, and is perhaps touch friendly. This is how it works on Reddit btw.

> allows multiple comment selection to mark multiple comments as read or unread

This implies some selection and unselection UX. How? Checkboxes?

victorgcramos:
Yes! Checkboxes!

xaur:
Checkboxes on all comments would add a lot of visual noise. Honestly, I'm not sure right now that the feature to select multiple comments is necessary. This makes the process of marking them as read to complex: first click a dozen checkboxes and then press some "Mark as read" button. It would be much simpler to just click comment areas of the same comments.

victorgcramos:
I have the screen preview for that. Let me know what you think!
Captura de Tela 2019-04-06 às 15 21 30

xaur:
I still think checkboxes is an overkill.

Behavior like Reddit's unread message box would be ideal - the whole comment bounding box has a bit greyish or bluish background, and when you click anywhere on it it toggles the state between read/unread. I would just read through the comments and spend 1 click on each, instead 1 click on each checkbox and then an extra click on "mark as read".

Feel free to ask more people.

fernandoabolafio:
@victorgcramos what is the status of the backend work for this issue? Can you link it here?

fernandoabolafio:
This feature has been rejected as per @lukebp [comments](https://github.com/decred/politeiagui/issues/1021#issuecomment-486679698) so this PR is not applicable anymore. It is a good practice to always validate new features with other devs in the matrix channel. It saves you some hours of coding :)

xaur:
@fernandoabolafio to my memory a feedback on the issue and PR was requested multiple time in chat with zero reply. Perhaps pursuing it without any feedback was not a good idea, but devs not replying to multiple requests like this is a problem too.

fernandoabolafio:
I agree @xaur. I'm not here to blame anyone. Let's just try to improve the communication between the two sides next time so we make sure that every feature enters the coding stage with enough validation from the most experienced devs.

PR #1071, opened 2019-04-04
Add date-filter for invoices

### LINKED ISSUE #1056: Add invoice listing filter by month/year
This filter needs to be added to the user and admin invoices list

camus-code:
resolves #1056


src/components/DateFilter.js
```diff
@@ -0,0 +1,191 @@
...
+  YEAR_FIVE,
+  YEAR_SIX,
+  YEAR_SEVEN,
+  YEAR_EIGHT,
+  CMS_LIST_HEADER_ADMIN,
+  CMS_LIST_HEADER_USER,
+  CMS_DEFAULT_TAB_TITLE
+} from "../constants";
+
+const invoiceFilterMonthOptions = [
```
tiagoalvesdulce: Consider creating an array of months, e.g `["January", "February", ...]` and use `reduce` to adapt it to your needs. For example `invoiceFilterMonthOptions  = months.reduce((arr, m, idx) => [...arr, { label: m, value: idx + 1 }], [] )`.

src/constants.js
```diff
@@ -104,3 +104,27 @@ export const CMS_LIST_HEADER_ADMIN = "Admin Invoices";
 export const CMS_LIST_HEADER_USER = "Your Invoices";
 
 export const CMS_DEFAULT_TAB_TITLE = "Contractor Management System";
+
+// months && years
+export const YEAR_ONE = 2019;
```
fernandoabolafio: No need to define constants for the year or months values, you can use the plain value instead. 
Please, remove all those month/year constants and update the code accordingly.
Ps.: See [comments](https://github.com/decred/politeiagui/pull/1071#discussion_r273069146) from @tiagoalvesdulce

src/components/DateFilter.js
```diff
@@ -0,0 +1,191 @@
...
+    label: "november",
+    value: NOVEMBER
+  },
+  {
+    label: "december",
+    value: DECEMBER
+  }
+];
+
+const invoiceFilterYearOptions = [
```
tiagoalvesdulce: See above comment about months.

src/components/snew/Content.js
```diff
@@ -97,12 +102,20 @@ export const CustomContent = ({
         </div>
       )}
       {isCMS ? (
-        <InvoiceFilter
-          header={header}
-          handleChangeFilterValue={onChangeFilter}
-          filterValue={filterValue}
-          invoiceCounts={invoiceCounts}
-        />
+        <div>
```
fernandoabolafio: Consider using a React fragment here to avoid adding useless tags:
```suggestion
         <>
          <InvoiceFilter
            header={header}
            handleChangeFilterValue={onChangeFilter}
            filterValue={filterValue}
            invoiceCounts={invoiceCounts}
          />
          <DateFilter
            header={header}
            handleChangeDateFilter={onChangeDateFilter}
            monthFilterValue={monthFilterValue}
            yearFilterValue={yearFilterValue}
          />
        </>
```

src/actions/app.js
```diff
@@ -141,6 +143,9 @@ export const onChangePublicFilter = option =>
 export const onChangeUserFilter = option =>
   act.CHANGE_USER_FILTER_VALUE(option);
 
+export const onChangeDateFilter = (month = _MONTH, year = YEAR_ONE) =>
```
fernandoabolafio: Please, use names more meaningful for constants. One cannot tell what `_MONTH` stands for. 
Btw, I noticed that this constant has the value o `99`, why is that?

src/components/DateFilter.js
fernandoabolafio: Generally speaking, when you're going to create another array with the same size from the original one you can just use the `map` instead of `reduce` to simplify the code.

camus-code: Thanks for this advice! Just implemented it and the code is a whole lot cleaner

src/components/DateFilter.js
```diff
@@ -0,0 +1,112 @@
...
+const invoiceFilterMonthOptions = MONTHS.reduce(
+  (arr, m, idx) => [...arr, { label: m, value: idx + 1 }],
+  []
+);
+const invoiceFilterYearOptions = YEARS.reduce(
+  (arr, y) => [...arr, { label: y, value: parseInt(y, 10) }],
+  []
+);
+
+const mapDropdownToMonths = {
```
fernandoabolafio: No need for that. We can use the same filter options for every case.

src/components/DateFilter.js
```diff
@@ -0,0 +1,112 @@
...
+  [CMS_LIST_HEADER_ADMIN]: invoiceFilterMonthOptions,
+  [CMS_LIST_HEADER_USER]: invoiceFilterMonthOptions,
+  [CMS_DEFAULT_TAB_TITLE]: []
+};
+
+const mapDropdownToYears = {
```
fernandoabolafio: Same as above.

src/selectors/app.js
```diff
@@ -422,22 +429,42 @@ const getInvoicesByStatus = (invoices, status) =>
     ? invoices
     : invoices.filter(i => i.status === status);
 
+const getInvoicesByYear = (invoices, year) =>
+  year === getCurrentYear() ? invoices : invoices.filter(i => i.year === year);
```
fernandoabolafio: The invoice shape has been changed in the backend, see [here](https://github.com/decred/politeia/blob/master/politeiawww/api/cms/v1/v1.go#L107). 
This has to become:
```suggestion
  year === getCurrentYear() ? invoices : invoices.filter(i => i.input.year === year);
```

src/selectors/app.js
```diff
@@ -422,22 +429,42 @@ const getInvoicesByStatus = (invoices, status) =>
...
+
+const getInvoicesByMonth = (invoices, month) =>
+  month === FILTER_ALL_MONTHS
+    ? invoices
+    : invoices.filter(i => i.month === month);
```
fernandoabolafio:
```suggestion
    : invoices.filter(i => i.input.month === month);
```

camus-code:
Thanks @tiagoalvesdulce . Good catch on the details page I completely missed that. Most recent commit should fix it!

camus-code:
Thanks @tiagoalvesdulce . Good catch on the details page I completely missed that. Most recent commit should fix it!

src/components/DateFilter.js
```diff
@@ -66,9 +66,8 @@ class DateFilter extends React.Component {
     const selectedMonth =
       invoiceFilterMonthOptions.find(op => op.value === monthFilterValue) ||
       monthFilterValue === FILTER_ALL_MONTHS;
-    return invoiceFilterMonthOptions &&
-      invoiceFilterYearOptions &&
-      header !== CMS_DEFAULT_TAB_TITLE ? (
+
+    return header !== (CMS_DEFAULT_TAB_TITLE && undefined) ? (
```
tiagoalvesdulce: I think this condition is clearer: `header && header !== CMS_DEFAULT_TAB_TITLE `

camus-code: Yeah agreed. Just updated it

PR #1079, opened 2019-04-09
[cms] Remove unused cms register fields from form

### LINKED ISSUE #1074: Remove outstanding fields from cms signup form
The signup form needs to be updated, only the following fields must be kept:
- Email
- Username
- Password
- Verification token

alexlyp:
Close #1074

### DIFF
--- a/src/components/snew/RegisterForm.js
+++ b/src/components/snew/RegisterForm.js
@@ -120,46 +120,6 @@ const RegisterForm = ({
         </div>
       </div>
       <div className="c-form-group">
-        <label className="screenreader-only" htmlFor="user_reg">
-          name:
-        </label>
-        <Field
-          name="name"
-          placeholder="Full Name"
-          component="input"
-          type="text"
-          className="c-form-control"
-          id="username_reg"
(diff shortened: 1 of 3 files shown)


alexlyp:
Seems like this is kind of a dupe of #1064.  Feel free to close this one if that one is more appropriate.

src/components/snew/RegisterForm.js
```diff
@@ -120,46 +120,6 @@ const RegisterForm = ({
         </div>
       </div>
       <div className="c-form-group">
-        <label className="screenreader-only" htmlFor="user_reg">
```
fernandoabolafio: Now that the regular signup form and the cms form diverge by only one field, what do you think if instead of duplicating the email/password/password_confirm fields we simply add the verification token field based on the `isCMS` value ?  
Exactly as this diff: https://github.com/decred/politeiagui/pull/1064/files#diff-dd229c10752f743a5438ddb5c1e9d8a0

alexlyp: yah i think that's great.  I'll make this change.

PR #1083, opened 2019-04-11
Auto fill email and verification token fields on register/signup routes

### LINKED ISSUE #1082: [cms] Add a new route for invitiatiion links
Links in invitation emails are as follows:

`/register?email=ay-p@example.org&token=<token>`

This should resolve to the RegisterForm page and have the email and token fields auto populated.  We could do something fancier in the future and go to a new form that only takes username, password and verify password.  But for MVP, just the auto-population is fine.

tiagoalvesdulce:
Closes #1082

### DIFF
--- a/src/hocs/signupForm.js
+++ b/src/hocs/signupForm.js
@@ -6,11 +6,11 @@ import { withRouter } from "react-router-dom";
 import { SubmissionError } from "redux-form";
 import signupConnector from "../connectors/signup";
 import appConnector from "../connectors/app";
+import { getQueryStringValues } from "../lib/queryString";
 
 class SignupFormContainer extends Component {
   constructor(props) {
     super(props);
-
     this.state = {
       hasFetchedPolicy: false
     };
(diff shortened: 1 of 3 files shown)


alexlyp:
Hrm the token isn't populating for me, only the email address currently

tiagoalvesdulce:
Can you share the URL you are using to test? Here works fine with `/register?email=ay-p@example.org&token=<token>`. The email field is filled with `ay-p@example.org` and the invitation token is filled with `<token>`.

alexlyp:
https://localhost:3000/register?email=email%40email.com&verificationtoken=279edfcf5d422284d24690d9e8913e5be7f851631a4b87c4f7a1409ba169555a

alexlyp:
Whoops yah I mistaken said it was just "token" but it's actually verificationtoken in the link.

Worked when I changed it to just token

src/hocs/signupForm.js
```diff
@@ -97,8 +97,15 @@ const wrap = Component =>
     return <Comp />;
   });
 
+const { email, token: verificationtoken } = getQueryStringValues();
```
alexlyp: can just remove the "token:" I believe, sorry about that!

PR #1095, opened 2019-04-15
Reset date filter when no query strings params are available

### LINKED ISSUE #1091: Date filter should be reseted by default if no params in query string
Steps to reproduce: 
- Go to admin invoices list
- Set a dater filter (ex: 04/2019). The query params are updated accordingly.
- Go to another route (ex: home) 
- Go back to admin invoices list, the query string is blank (as expected) and the date filter is still showing the old filter (04/2019)

The date filter must be reset to its default state every time the query parameters don't exist or are blank.

camus-code:
Resolves #1091 

This commit adds a new action creator that resets the date filter values to their initial state. This action is called whenever there is a route change without any date params in the query string


src/components/Verify/Page.js
```diff
@@ -5,11 +5,12 @@ import PageLoadingIcon from "../snew/PageLoadingIcon";
 const Page = ({
   verifyNewUser,
   isRequestingVerifyNewUser,
-  verifyNewUserError
```
fernandoabolafio: It Looks to me that this diff in this file does not belong to this pull request, right?

camus-code: Yup you're totally right. Not sure how that snuck its way in there. Just updated

fernandoabolafio:
Great @camus-code! can you rebase it?

camus-code:
Yeah I just did! 

Let me know if there's any other issues @fernandoabolafio . I ran into a strange error at first about CMSHome not being found, but it went away and I can't seem to reproduce it again

PR #1104, opened 2019-04-16
Disable 'comments' tab

### LINKED ISSUE #1100: Comments tab on Account page not working
The tab is there but when clicked gives a message: "This feature has not been implemented yet." If the feature will not be implemented for some time it may be better to hide the tab.

victorgcramos:
Fixes #1100


src/components/UserDetail/Page.js
```diff
@@ -29,6 +27,7 @@ const UserDetailPage = ({
   numOfUserProposals,
   isCMS
 }) => {
+  console.log(tabId);
```
fernandoabolafio: Please, remove this log

PR #1127, opened 2019-04-23
Update successful Signup Page

### LINKED ISSUE #1080: [cms] Signup/next should be avoided
The success page for signup (signup/next) doesn't need to be shown since the information isn't accurate.  Probably a success modal with a short message should show up.

camus-code:
Closes #1080 

This commit updates the copy for the successful signup page on CMS. Once we finish instructions on how to properly format an invoice, these will be linked here as well.

Commit also removes the Pi welcome modal from CMS

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -49,7 +49,7 @@ class Loader extends Component {
       );
     }
 
-    if (!prevProps.onboardViewed && isUserFirstLogin) {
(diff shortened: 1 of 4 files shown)


alexlyp:
Looks good.  Can you also add a message that they can now login and add a link back to /user/signup

camus-code:
Yeah adding that back now

PR #1129, opened 2019-04-24
Update commentslikes test route

### LINKED ISSUE #1115: Update test for 'on fetch liked comments action'
https://github.com/decred/politeiagui/blob/cd6f067368d0d3b1634c05d5f4f93be6d6c8e11e/src/actions/tests/api.test.js#L511

This code is currently testing a non-existent route. This should be updated to the correct `commentslikes` route

We should also consider extending test coverage to more prominently warn for unmatched routes such as this one

camus-code:
Closes #1115 

This updates the incorrect route for `commentslikes`.

### DIFF
--- a/src/actions/tests/api.test.js
+++ b/src/actions/tests/api.test.js
@@ -508,20 +508,40 @@ describe("test api actions (actions/api.js)", () => {
   });
 
   test("on fetch liked comments action", async () => {
-    const path = `/api/v1/proposals/${FAKE_PROPOSAL_TOKEN}/commentslikes`;
-    const params = [FAKE_PROPOSAL_TOKEN];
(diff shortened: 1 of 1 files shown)


fernandoabolafio:
@camus-code thanks for this fix. However, if we are adding this option of failing in case of the route not matching, this should be the default behavior for the following functions which asserts that a route is called, for instance:

**actions/tests/api.test.js**:
Functions: `assertApiActionOnSuccess`, `assertApiActionOnError`

**lib/tests/api.tests.js**:
Functions: `assertGETOnRouteIsCalled`, `assertRouteIsCalledWithQueryParams`, `assertPOSTOnRouteIsCalled`

All these functions should fail if the routes passed to them is not matched in the subsequent request.

Can we have a mechanism for that? Maybe overriding the code in `beforeEach`?

PR #1132, opened 2019-04-24
Fetch lastblockheight on proposal detail view

### LINKED ISSUE #1130: Proposal info inconsistencies
If I click on a link to a proposal say from https://proposals.decred.org/?tab=active%20voting I will see information on the new page about how long the vote has until completion.
If I visit the proposal link directly this info will be missing (see screenshots below).

image
image

Refreshing the page after visiting the proposal from the active voting link causes the vote duration info to disappear.

Also if visited from the link the title of the page will not be correct until a refresh is performed.

camus-code:
Closes #1130 

This commit fixes the incorrect conditional statement which only fetched the last block height whenever `onFetchProposalsVoteStatus` was called. The statement was updated to also fetch the last block height whenever `onFetchProposalVoteStatus` was called.

Commit also fixes the bug where document.title was not being correctly set according to the proposal name

### DIFF
--- a/src/components/RecordDetail/RecordDetail.js
+++ b/src/components/RecordDetail/RecordDetail.js
@@ -21,7 +21,18 @@ class RecordDetail extends React.Component {
       accessTime: 0
     };
   }
-
+  resolveTabTitle = prevProps => {
+    const { isCMS, record } = this.props;
+    const proposal = !isCMS && record;
+    const proposalNameHasBeenUpdated =
+      (proposal &&
+        prevProps.record &&
+        prevProps.record.name !== proposal.name) ||
+      (proposal && proposal.name !== document.title);
+    if (proposalNameHasBeenUpdated) {
+      document.title = proposal.name;
+    }
+  };
   resolveFetchProposalVoteStatus = prevProps => {
     const { isCMS, record, onFetchProposalVoteStatus, token } = this.props;
     const proposal = !isCMS && record;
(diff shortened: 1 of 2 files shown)


fernandoabolafio:
@camus-code the fix for the issue #1130 looks good to me. What is this fix to the tab title you're referring to? I don't recall any bug in that regard, can you provide more info?

camus-code:
@fernandoabolafio So the problem was that updating the tab title worked for switching between proposals. However, if you clicked on a proposal, went back to the home screen, and then clicked on that proposal again, the tab title would not change

src/components/RecordDetail/RecordDetail.js
```diff
@@ -21,15 +21,7 @@ class RecordDetail extends React.Component {
       accessTime: 0
     };
   }
-  resolveTabTitle = prevProps => {
-    const { isCMS, record } = this.props;
-    const proposal = !isCMS && record;
-    const proposalNameHasBeenUpdated =
```
fernandoabolafio: I think you can keep this code structure as it keeps the `componentDidUpdate` cleaner. 
I tested and if you simpyl update `proposalNameHasBeenUpdated` check to: 
```js
const proposalNameHasBeenUpdated =
      (proposal &&
        prevProps.record &&
        prevProps.record.name !== proposal.name) ||
      (proposal && proposal.name !== document.title);
```
It solves the bug you mentioned

camus-code: Yeah good catch. Just updated this @fernandoabolafio

PR #1134, opened 2019-04-25
Lazy load public proposals 

### LINKED ISSUE #1013: Improving naming of state-tags: "Pre-voting" and "Active Voting"
Referring to this issue opened for Decrediton https://github.com/decred/decrediton/issues/2015
Fair point on consistency. 

I've introduced the term "Under Discussion" instead of "Pre-voting" in Decrediton previously, since pre-voting doesn't communicate this phase very easily nor accurately (e.g. is there's some sort of before-hand voting happening).

Decrediton lists currently the following: `Under Discussion`, `Under Vote`, `Finished Voting`, `Abandoned`. 

Bringing up the issue here, as perhaps the terms could be further ironed out altogether, and then updated in both Pi and Decrediton consistently. The main problem I see with Decrediton tags is the use of double "under". Ideally each state tag should be as simple adjective as possible as when they're all looked at together that also helps to understand each individual one. 

Some thoughts:
1) Pre-Voting → "New Proposals", "Proposals" "In Discussion" "New Submissions" "Submissions" etc
2) Active Voting → "Voting"
3) Approved
4) Rejected
5) Abandoned

### LINKED ISSUE #1107: Lazy load proposals using inventory route
Since https://github.com/decred/politeia/pull/809 we can now fetch the token for all public proposals already categorized into pre-voting, active, finished and abandoned.  
With that, the public proposals loading must be lazy loaded as the user scroll through the list or switch tabs.

fernandoabolafio:
This PR:
- Decouple Public Proposals list from snew
- Lazy load public proposals by firstly fetching the tokens inventory
- Paves the way for redesign changes. The structure of the page of public proposals is how I intend to implement for the other pages.

This PR depends on a change in the backend side in the tokens inventory response which will split the "finished tokens" into "approved" and "rejected".

depends on https://github.com/decred/politeia/issues/819
closes #1107 
closes #1013


src/components/snew/ThingLink/ThingLinkProposal.js
```diff
@@ -260,14 +258,14 @@ class ThingLinkComp extends React.Component {
               }}
             >
               {isEditable ? (
-                <Link
+                <a
```
camus-code: Why did you switch this from <Link> to an anchor tag?

camus-code:
This looks really good to me. I think you did a great job of neatly decoupling the proposals from our snew library. The infinite scrolling / lazy loading is working great on my end as well.

The only problem I'm seeing for now is the `Rejected` tab count is not updating properly. As you can see here, I have one rejected proposal but the count reads "0" The rest of the tabs seems to be updating properly.

rejected

I also left an inline comment as well

src/components/snew/ThingLink/ThingLinkProposal.js
fernandoabolafio: Ah, I replaced it because before the `Link` component we were using was from snew so I was having a few styling issues. However, I switched the Link component to the one from `react-router-dom` which removes one more dep on snew and also makes the route transition smoother. Anyway, good catch, I will replace it.

camus-code:
Does this PR close this issue as well: https://github.com/decred/politeiagui/issues/1013 ? At least on the Politeia side of things

fernandoabolafio:
Yes, it does. Well remembered.

Em ter, 30 de abr de 2019 22:38, Kevin <notifications@github.com> escreveu:

fernandoabolafio:
Yes, it does. Well remembered.

PR #1140, opened 2019-04-29
Add exchange rate to invoices

### LINKED ISSUE #1135: Include exchange rate on invoice submissions
With the this PR on politeia https://github.com/decred/politeia/pull/821, invoices now require that the exchange rate is set within the invoice.json payload.

In that PR, I've accomplished this on CLI by first requesting the the `invoiceexchangerate` for the month/year of the invoice, then adding that to the invoice.json and signing as normal.

I envision the process to be similar on the GUI.  When the user changes the month/year dropdowns, the exchange is rate is requested (and then displayed on the submit invoice form) which can then be added to the `newInvoice` invoice.json payload.

fernandoabolafio:
This PR:

 - Display the exchange rate on the Submit Invoice Form
 - Send exchange rate within the invoice payload (on edit and on submit)
 - Set the default month and year options, for a new invoice, to the current month and year
 - Limit the month options to months prior to the current month
 - Update error messages based on the API
 - Make the index route `/` protected to authenticated users only

closes #1135


tiagoalvesdulce:
Looks good @fernandoabolafio. I noticed two things while testing this PR:

1. If you submit an invoice without touching the month/year dropdown, it will throw an error but it's showing 01/2019. IMO we should change the default of the select component to Month/Year.

2. The error for a blank payment address is being shown as "Rate cannot be blank".

Those errors are not related to this PR, I'm just letting you know in case you wanna fix it before it goes in. I'm cool with merging this one and creating issues for the 2 points I've noticed.

alexlyp:
Yah I'll open issue for those 2 problems with exchangerate endpoint.

fernandoabolafio:
@tiagoalvesdulce I will handle the issues you pointed out in a different PR.

PR #1155, opened 2019-05-02
[cms] Ensure policy is populated for FilesField

alexlyp:
Checks to make sure policy is populated before attempting to render the Field


camus-code:
Closes #1159

PR #1156, opened 2019-05-02
[cms] Add submit invoice instructions to page

### LINKED ISSUE #1152: [cms] Add cms instructions to submit invoice page
The instructions at https://github.com/decred/politeia/blob/master/cms-instructions.md should be shown on the /invoices/new page.  Possibly to the right of the current contractor form.

alexlyp:
Close #1152

### DIFF
--- a/src/components/Form/Fields/FilesField.js
+++ b/src/components/Form/Fields/FilesField.js
@@ -53,50 +53,52 @@ export class FilesField extends React.Component {
       margin: 0
     };
     return (
-      <div className="attach-wrapper">
-        {policyErrors.length > 0 && <PolicyErrors errors={policyErrors} />}
-        <div>
-          <ReactFileReader
-            base64
-            multipleFiles
-            fileTypes={policy.validmimetypes}
-            handleFiles={this.handleFilesChange}
-          >
-            <div className="button-wrapper">
-              <button
-                className={`togglebutton access-required${
-                  !userCanExecuteActions ? " not-active disabled" : ""
-                }`}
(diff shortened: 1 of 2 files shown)


src/components/Form/Fields/FilesField.js
```diff
@@ -53,50 +53,52 @@ export class FilesField extends React.Component {
...
-                  Max file size:{" "}
-                  <span>
-                    {Math.floor(policy.maximagesize / 1024)} Kb.{" "}
-                  </span>{" "}
-                </div>
-                <div>
-                  {" "}
-                  Valid MIME types:{" "}
-                  <span>{policy.validmimetypes.join(", ")}</span>{" "}
+      policy && (
```
go1dfish: You could avoid indentation here with:

    if (!policy) return null;
    return (....);

since you are short circuiting the entire render.

src/components/snew/SubmitPage/SubmitPageInvoice.js
```diff
@@ -137,37 +157,62 @@ const InvoiceSubmit = props => {
...
-                      name="address"
-                      label="Payment address"
-                      type="text"
-                      component={InputFieldWithError}
-                    />
+                  <div
+                    className="usertext"
+                    style={{
+                      display: "flex",
+                      width: "100%"
```
fernandoabolafio: One way to prevent the overlapping described is to set the style: `justify-content: space-between` and remove the `width: "50%"` from the children. Just a suggestion, up to you how to deal with it.

Issue #990, opened 2019-01-15
Politeiagui code refactoring


fernandoabolafio:
As we approach the redesign of Politeiagui, we are considering a significant code refactoring to deal with DX and scalability issues that have been raised by the developers along the development process.
That way we could align a brand new design with better code and performance. See [this discussion](https://matrix.to/#/!VFRvyndKpzcLrVslQD:decred.org/$15475717344989MkbYi:decred.org?via=decred.org&via=matrix.org&via=zettaport.com) on Matrix for instance.

**Issues raised so far:**
- State Management: The app state management is getting way too complicated and hard to maintain. The use of Redux all over the place is resulting in a lot of boilerplate and edge cases which are hard to understand.
- Caching: Even though we should be preventing duplicate requests by using Redux, it isn't happening. Leading to higher data consumption.
- Snew dependency: Most of the code depends on the [snew-classic-ui](https://github.com/decred/snew-classic-ui) which is a React UI lib version of the old Reddit design. This dependency is hard to maintain and it's components weren't designed for responsivity across multiple screen sizes.

We need to discuss possible solutions to it and evaluate the feasibility of doing it.

Please, write your proposed solutions or thoughts in the comments.

victorgcramos:
Apollo looks like a good solution!

Also, what about the *snew-classic-ui*? I think it would be better for us to sit down an discuss the whole new design for Politeia, since our UI kit is outdated. Also, we can work on improving the mobile responsivity, which we don't have.

IMO, creating a new design kit from scratch would take less time than keep improving an existing one. 

**I'm able to help to create this!**

camus-code:
Completely agree @fernandoabolafio. 

snew-classic has become a big bottleneck and I don't really think it adds that much in terms of functionality. Not to mention it prevents people from easily being able to contribute to the GUI repo.

I'm not sure if the ideas is to create a UI lib from scratch or to implement an existing one, but I've come to really enjoy [Semantic UI - React](https://react.semantic-ui.com) for building responsive & customizable UX. They have a pretty expansive module / element collection that covers pretty much everything we already currently use within the GUI.

victorgcramos:
@camus-code this UI Kit is awesome! Great idea!

orthomind:
Just as an outsider to politeia code looking in, as a suggested alternative to Apollo Server, I've had good experiences myself using Apollo Client with the Go based [gqlgen](https://github.com/99designs/) as the GraphQL server.

[...]

PR #1160, opened 2019-05-03
Proposal redux hooks refactor

go1dfish:
This is an example of the refactor approach I describe here:

https://github.com/decred/politeiagui/issues/990#issuecomment-488499188

With some considerations to set up towards https://github.com/decred/politeia/issues/839#issuecomment-488734026

I recommend moving high level features/pages to top level components as we refactor them.  I've found this to be a pretty useful organizational strategy for react apps using this sort of provider approach.

The components/PublicProposalsPage/* files would move to /PublicProposalsPage/ and more logic from the PubllicProposals component would get moved into provider if I continue with this.  They are left as is to minimize the changes to highlight the more important bits.


src/Api/ApiProvider.js
```diff
@@ -0,0 +1,50 @@
...
+  const set = useCallback((key, value) => {
+    setState(s => ({ ...s, [key]: value }));
+  }, []);
+  const value = useMemo(() => [get, set], [get, set]);
+
+  return <ApiContext.Provider value={value}>{children}</ApiContext.Provider>;
+};
+
+export function useApiCall(commandAndParams) {
+  const [cmd, ...params] = commandAndParams;
```
fernandoabolafio: Not sure I like passing the names of the functions as a string here, it will require always checking the functions to make sure you're typing the correct name of the function. Maybe make `cmd` a function can be a good idea, then we pass the function and the params we want to apply in that function and delegate the execution to `useApiCall`. Something like:

```js
const [res, load, err] = useApiCall(api.fetchSomething, ...params);
```

src/Api/ApiProvider.js
```diff
@@ -0,0 +1,50 @@
...
+  const key = JSON.stringify(commandAndParams);
+  const [getResponse] = useApiState();
```
fernandoabolafio: Why are you not using the `setResponse` method in here? Shouldn't it be updating the `apiState` for every fetched response so it is available for other consumers?

src/PublicProposalsPage/PublicProposalsProvider.js
```diff
@@ -0,0 +1,91 @@
...
+      // This is an example of updating redux store with data
+      // For compatibility during refactors
+      redux.dispatch(act.RECEIVE_PROPOSALS_VOTE_STATUS(null, voteStatusError));
+    },
+    [voteStatusError]
+  );
+
+  const value = {
+    isLoading:
+      isLoadingProposals || isLoadingVoteStatus || isLoadingTokenInventory,
```
fernandoabolafio: The loading of proposals is handled in a different way in the inner (`components/PublicproposalsPage/PublicProposals.js`) component, so it should not trigger the global loading of the component. This loading management can, however, be pulled into the provider, if we want to.
```suggestion 
       isLoadingVoteStatus || isLoadingTokenInventory,
```

src/PublicProposalsPage/PublicProposalsProvider.js
```diff
@@ -0,0 +1,91 @@
...
+
+    // Component expects these two, they are intentionally
+    // blank for now to avoid changing design component
+    onFetchProposalsVoteStatus: useCallback(() => {}, []),
+    onFetchTokenInventory: useCallback(() => {}, [])
+  };
+
+  return (
+    <PublicProposalsContext.Provider
+      value={useMemo(() => value, Object.values(value))}
```
fernandoabolafio: Can you explain to me why is this memoized value needed here? Looks odd to me that it needs to be told to recalculate a value based on itself.

go1dfish: Contexts will trigger consumers to re-run when value changes.  Memozing value based on its values makes it so the reference identity of value only changes when one of its attributes actually changes.

src/PublicProposalsPage/PublicProposalsProvider.js
go1dfish: > The components/PublicProposalsPage/* files would move to /PublicProposalsPage/ and more logic from the PubllicProposals component would get moved into provider if I continue with this. They are left as is to minimize the changes to highlight the more important bits.

src/Api/ApiProvider.js
go1dfish: From the file comments:

>  set would only ever get used if we wanted to maintain local cache of previous requests while browsing

You could use ApiProvider to avoid doing multiple api requests on subsequent visits to the same page, but for now I was only concerned with using it as a way to preload initial state for when we want to do server side rendering.

Wasn’t sure if further caching was desirable, but would be easy to achieve this way.

src/Api/ApiProvider.js
go1dfish: Ok, like the idea of changing the first argument to a command, I think I’d prefer to refactor it to be used as...

useApiCall(api.fetchSomething, [...params], ...roomForFutureOptions)

Main reason I started with a string is the commandAndParms gets stringified to build the cache key, but we can probably derive this from function names as well.

go1dfish:
Regarding wanting a single source of approvals data, I realized that this would be desirable as I was building this, but wanted to keep the initial example focused on just public proposals as we discussed.

If the general pattern here is agreeable I can continue with it to build a shared proposal provider that will serve as a consistent source of truth for multiple pages.

fernandoabolafio:
Thanks for the answers in the inline comments. I see no reason to not proceed with this pattern. Starting with proposals, which is by far the most complex context will provide us a solid base for extending it to the other contexts. @tiagoalvesdulce can you take a look on this as well?

go1dfish:
Sounds good, I figure the proposals provider is likely the most important in terms of serverside rendering as well as that’s the most important dynamic content we want to render for non-js clients.

Note that there is no server side rendering in this branch and I don’t plan to tackle it in the branch but I do want to get the api interactions set up well for doing that in a subsequent branch.

go1dfish:
@fernandoabolafio this is at a good stopping point for review and merge.

From here I can move on to trying to improve redux code locality (moving that logic into src/Proposals) or on the legacy vs redesign component switching for these same components.

fernandoabolafio:
@go1dfish I've been debating about this with @tiagoalvesdulce and I think that would be better to use this refactoring for the redesign process but I don't want to merge this code in the current version we have in prod. This can lead to many bugs that I don't want to deal with right now. I will make sure to use this branch as a reference for the proposal part in the redesign but for now, let's put it on hold.

fernandoabolafio:
As per my last comment, the refactoring of the proposals code was included in the v2. Hence, this PR will be closed as we plan to not update the old GUI version anymore. Thanks for this PR. some of the concepts you used here inspired us during the proposals refactoring in the redesigned code.

PR #1161, opened 2019-05-03
Enable app customisation based on external configuration

fernandoabolafio:
This PR paves the way for further customization of Politeiagui by specifying settings and toggling features from an external configuration file.

To make this process easier I used [dotenv](https://github.com/motdotla/dotenv). 

Specify your config options by creating a `.env` file in the root folder of the project. All 
variables must be preceeded by "REACT_APP" as follow:

**/.env:**

```dosini
REACT_APP_TITLE="Politeia"
REACT_APP_STAGING=false
```

It is also possible to specify a preset to be used:

**/.env:**

```dosini
REACT_APP_PRESET="POLITEIA"
```

If the `REACT_APP_PRESET` is specified, all other specified options in the env
the file will be ignored.

The default preset is the POLITEIA, see it in `src/Config/presets.js`


package.json
```diff
@@ -10,6 +9,8 @@
     "date-fns": "^1.30.1",
     "diff": "^3.5.0",
     "dompurify": "^1.0.10",
+    "dotenv": "^7.0.0",
```
go1dfish: I may be wrong, but I think create react app already includes dotenv and support for .env out of the box.  May be worth trying without adding the explicit dependency here.

src/config/index.js
```diff
@@ -0,0 +1,4 @@
+import Config from "./Config";
+export * from "./ConfigProvider";
+
+export default Config;
```
go1dfish: I’d rename /config to /Config since it is effectively exporting “Config” as a component for use.

src/config/ConfigProvider.js
```diff
@@ -0,0 +1,30 @@
...
+export const ConfigConsumer = ConfigContext.Consumer;
+
+export const useConfig = () => React.useContext(ConfigContext);
+
+/**
+ * This wrap around the config provider is needed to add props validation for
+ * each custom option alllowed.
+ */
+export const ConfigProvider = ({ children, ...configOptions }) => (
+  <ConfigContext.Provider value={configOptions}>
```
go1dfish: May want to consider memoization  here.

If any parent component triggers a re-render this will currently create a new configOptions object causing anything using this context to also re-render

src/App.js
```diff
@@ -94,13 +100,19 @@ class Loader extends Component {
...
-  process.env.REACT_APP_STAGING ? (
-    <div className="staging-alert">
-      This is the politeia staging environment. DO NOT USE, YOU WILL LOSE YOUR
-      DECRED.
-    </div>
-  ) : null;
+const StagingAlert = () => {
+  const { isStaging } = useConfig();
+  return useMemo(
+    () =>
```
go1dfish: I don’t think there is any benefit to memoization here unless I’m missing something.

Rendering dumb JSX is very cheap so no need for the extra complexity here IMO

go1dfish:
Added some suggestions, but like this approach.

src/Config/ConfigProvider.js
```diff
@@ -0,0 +1,32 @@
...
+
+export const useConfig = () => useContext(ConfigContext);
+
+/**
+ * This wrap around the config provider is needed to add props validation for
+ * each custom option alllowed.
+ */
+export const ConfigProvider = ({ children, ...configOptions }) => (
+  <ConfigContext.Provider
+    value={useMemo(() => configOptions, Object.values(configOptions))}
```
go1dfish: I think you'll be fine here, but one thing to keep in mind with useMemo is that the second argument must have a consistent length.

If more/less config options could be added during runtime (unlikely) this could break.

No change necessary here, just mentioning it because you picked up this pattern from my PR.

fernandoabolafio: It's very unlikely the configuration will change during the runtime. If it happens then all of this component needs to be reviewed.

PR #1166, opened 2019-05-06
Support src/ imports

go1dfish:
This allows us to import from "src/....."

From: https://github.com/facebook/create-react-app/issues/5118#issuecomment-464368371


go1dfish:
.babelrc is deleted in this branch because it was ignored by create-react-app https://github.com/facebook/create-react-app/issues/1325

But isn't ignored by react-app-rewired.

PR #1169, opened 2019-05-06
Hide comment score

### LINKED ISSUE #1164: [cms] shows up/down vote point count
up and down voting is disabled for cms, but i still see "0 points" above comments.

please hide the point count display as the up/down vote div has been hidden or removed.

thi4go:
This is the quick fix to disable comment score on snew.
 
image

Currently shows a `[score hidden]` which might be unwanted, and if so, a simple PR on snew can change or remove this message.

closes #1164

### DIFF
--- a/src/components/snew/ThingComment.js
+++ b/src/components/snew/ThingComment.js
@@ -128,6 +128,7 @@ class ThingComment extends React.PureComponent {
             isNewComment,
             user: loggedInAsEmail,
             authorHref: `/user/${props.authorid}`,
+            score_hidden: props.isCMS,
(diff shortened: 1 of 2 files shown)


fernandoabolafio:
@thi4go I think modifying snew makes more sense for this. The message "score hidden" just adds noise to the view since it doesn't make any sense to the context. Can you tweak snew and then we use this PR to update the snew version?

thi4go:
Hey @fernandoabolafio just opened the PR on snew https://github.com/decred/snew-classic-ui/pull/30 👍

PR #1175, opened 2019-05-08
Fix "Community" typo

degeri:


alexlyp:
Mmmm "Managament"? dat double typo

PR #1183, opened 2019-05-09
Remove redirect when user session expires

### LINKED ISSUE #953: Don't redirect to logout page on session expiration
Per jy-p on Slack:

"something i find very annoying about pi's frontend is that when you get logged out after a period of inactivity, it sends all the open pages to the logged out page

e.g. i have N proposal pages open for reading, then i have to explicitly revisit those pages after a login expires and is auto-logged out"

thi4go:
Ended up leaving the redirect from admin routes, bc some of them contained sensitive information that would be best not to display them when a session is invalid, even if actions regarding that route are disabled. This PR removes the annoying redirect that happened when a session expires.

closes #953

### DIFF
--- a/src/selectors/app.js
+++ b/src/selectors/app.js
@@ -24,6 +24,7 @@ import {
   apiUserResponse,
   apiEditUserResponse,
   apiEditUserPayload,
+  loggedInAsEmail,
   isCMS,
   apiAdminInvoices
 } from "./api";
@@ -254,7 +255,11 @@ export const userHasPaid = state => {
   return getUserPaywallStatus(state) === PAYWALL_STATUS_PAID;
 };
 export const userCanExecuteActions = state => {
-  return userHasPaid(state) && !getKeyMismatch(state);
+  return (
+    userHasPaid(state) &&
+    !getKeyMismatch(state) &&
+    loggedInAsEmail(state) !== false
+  );
 };
 
 export const isProposalStatusApproved = state =>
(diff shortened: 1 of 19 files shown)


degeri:
This is only specific to admins?

thi4go:
Actually now the authenticated routes also don't redirect to logout page at expiration of a session. It'll only pop-up the login modal and disable actions that were once enabled with a valid session. Took my time to test this out, but sure could use more eyes on this @degeri :). Best way to test this is to set expiration time for 10s or such and see if the authenticated page actions gets disabled once expired.

thi4go:
Thanks for the thorough @fernandoabolafio, great feedback. I'll be working on this and hopefully will commit the changes soon.

src/selectors/app.js
```diff
@@ -254,7 +255,11 @@ export const userHasPaid = state => {
   return getUserPaywallStatus(state) === PAYWALL_STATUS_PAID;
 };
 export const userCanExecuteActions = state => {
-  return userHasPaid(state) && !getKeyMismatch(state);
+  return (
+    userHasPaid(state) &&
+    !getKeyMismatch(state) &&
+    loggedInAsEmail(state) !== false
```
fernandoabolafio: Please, make sure to that the result ove `loggedInAsEmail(state)` is evaluated as a boolean if you want to detect falsy values. For example: `"" !== false` or `null !== false` is evaluated to `true`.
```suggestion
    !!loggedInAsEmail(state) !== false
```

PR #1193, opened 2019-05-14
Add drafts for invoices and multiline description field

### LINKED ISSUE #1173: [cms] UI suggestion : larger description box, multiline for description and ability to save draft
1. larger box for description
2. allow entering multiline for description
3. ability to save draft and submit later - i have this error and not able to submit and need to reenter everything when i refresh / logout.

camus-code:
Closes #1173 

This commit adds the ability to save invoices as a draft. Similar to the proposal drafts, these are saved in local storage and NOT available across devices.

This commit also changes the description column cells to a textarea, allowing for a resizable, multiline input field.


camus-code:
Thanks a bunch for the detailed review @fernandoabolafio. You're totally right about splitting these up as well. I normally try and do that, I just got a tad caught up with trying to resolve #1173 in full. In the future I will definitely split up issues like that so I'm not trying to tackle too much at once.

To keep up with the theme of splitting this up, I will be addressing your comments in separate commits

camus-code:
@fernandoabolafio I've addressed the two issues you brought up with the multiline editor and persisting data on browser refresh.

The only remaining issue is the ability to save drafts without all the fields filled in. As I mentioned to you last night, this behavior is consistent with how our proposal drafts work as well. Obviously this is not ideal, but I think it's sufficient in the mean time.

Since this PR is already pretty large, I think the problem mentioned deserves its own issue. I believe it'll take a fair bit of reorganizing our HOC's because of how redux-form handles errors versus warnings.

fernandoabolafio:
Sonds good @camus-code. Can you rebase it?

camus-code:
@fernandoabolafio yeah of course. Just rebased it.

PR #1194, opened 2019-05-14
Lazy load proposals vote status

### LINKED ISSUE #1203: Vote count showing on incorrect tabs
Clicking through the tabs on testnet causes the vote count to stay painted on the screen. 

image

fernandoabolafio:
This PR implements the lazy loading of vote status for proposals while moving a few pieces in the state management to make it possible.
With these changes, the vote status will be fetched only for proposals which have been already loaded. **The `VoteStats` component** will handle the fetching of the vote status in isolation. Which means it **will have it's own loading/error state, independently of the rest of the proposal**. However, the redux state is still being used for caching the response. We need that to prevent future re-fetching of data where the vote status are unlikely (or impossible) to change which means that **the re-fetching will be automatically triggered only for 'active voting' proposals.**
closes #1194 
closes #1203


fernandoabolafio:
Changing to WIP in order to test some edge cases before releasing it for review.

Issue #1212, opened 2019-05-20
[cms] Resize PNGs on invoice details page


alexlyp:
Currently PNGs are rendering at whatever resolution they are uploaded which result in some huge images that can't be fully seen.

They should scale down to the screen width.

victorgcramos:
This diff from #1218 should fix it: https://github.com/decred/politeiagui/blob/0bf469d6ef0150abf46b45cd62274d00b57bfc5b/src/style/index.css#L234-L235

fernandoabolafio:
@victorgcramos no, it doesn't fix it. This issue needs to be addressed as soon as possible as large files are making the screen content larger than the viewport creating a horizontal scroll for the entire screen:
![May-29-2019 07-52-44](https://user-images.githubusercontent.com/14864439/58532832-0acac480-81e7-11e9-8fb0-cd1ff9ffcdf1.gif)

victorgcramos:
Now I guess It's fixed

https://github.com/decred/politeiagui/blob/4375c9e3663137dc112147129891628ff4b965fa/src/style/index.css#L428

> @victorgcramos no, it doesn't fix it. It introduced another issue in the proposal details page in fact. Please, see the review in the Pull request.

victorgcramos:
Sorry, I made a mistake... Now, this should fix:

https://github.com/decred/politeiagui/blob/4375c9e3663137dc112147129891628ff4b965fa/src/style/index.css#L428

PR #1218, opened 2019-05-21
Diff Improvements

### LINKED ISSUE #1012: Proposal version viewer - no scroll bar in modal window
The only way to scroll when viewing an old proposal version is mousewheel, which is cumbersome when the proposal is long. I can see the scroll bar of the full page moving in the background but cannot drag it.

### LINKED ISSUE #1149: Proposal text escapes diff view
When looking through old proposal versions, it seems some of the styled markdown escapes the container. We need to make sure all text wraps within the container. 

Screen Shot 2019-05-01 at 12 07 49 PM

victorgcramos:
Fixes #1012, #1149 and #1212

### Proposal version viewer - diff component
This commit improves the Diff component, now showing the diffs on raw text instead of rendered markdown.

The purpose of doing this is to mitigate future errors due to markdown limitations to show diffs.

By analyzing the way that other applications deal with diffs on markdown files, such as GitHub, it is possible to conclude that the smartest way to deal with it is to separate the diff from the rendered content.

Besides that, this version now allows us to see the changes that are not rendered by the `MarkdownRenderer` such as blank lines addition. Formerly, if you added one or more blank lines, a new version was created, but the diff modal showed a new version without any changes.

### Proposal version viewer - scrolling issue
By the diffs added on this PR, the scrolling issue reported on #1012 should get fixed.

### Overflowed images on CMS
This also resizes overflowed images, as reported on #1212.
> In some cases, images with a wide width end up getting very shrunken. To solve this, we can implement an image viewer that displays the image with its full resolution. However, I will not use this space to add that feature, since that's not the purpose of this PR.

### DIFF
--- a/src/components/Modal/ModalContentWrapper.js
+++ b/src/components/Modal/ModalContentWrapper.js
@@ -31,7 +31,7 @@ class ModalContentWrapper extends React.Component {
       style
     } = this.props;
     return (
-      <div className="modal-content" style={{ minWidth: "700px", ...style }}>
+      <div className="modal-content" style={{ minWidth: "803px", ...style }}>
         <div className="modal-content-header">
           <h2 style={{ fontSize: "18px", textTransform: "uppercase" }}>
             {title}
--- a/src/components/Modal/Modal.js
+++ b/src/components/Modal/Modal.js
@@ -27,9 +27,10 @@ class Modal extends React.Component {
           disabled={disableCloseOnClick}
         />
         <div className="modal">
-          <div className="modal-overlay" />
-          <div className="modal-container" style={style}>
-            {children}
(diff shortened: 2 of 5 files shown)


src/components/Modal/ModalContentWrapper.js
```diff
@@ -31,7 +31,7 @@ class ModalContentWrapper extends React.Component {
       style
     } = this.props;
     return (
-      <div className="modal-content" style={{ minWidth: "700px", ...style }}>
+      <div className="modal-content" style={{ minWidth: "803px", ...style }}>
```
fernandoabolafio: How did you get to this number (803)?

fernandoabolafio:
Is it ready for another review?

PR #1222, opened 2019-05-22
[cms] Improve invoice validation 

fernandoabolafio:
This PR improves the invoices validation by:
- Adding full policy validation for invoice fields
- Display better validation message (including allowed chars)

It also improves the validation code by introducing [Yup](https://github.com/jquense/yup). This lib has more 'declarative' approach for dealing with forms validation. This dep is going to be introduced in the redesign process so I figure would be nice to use it already for invoices validation. The datasheet component doesn't handle the line items validation anymore, all the validation, including line items, were lifted to the synchronous validation using redux form + yup.

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,8 @@
     "tweetnacl": "^1.0.1",
     "tweetnacl-util": "^0.15.0",
     "typescript": "^3.4.3",
-    "xss-filters": "^1.2.7"
+    "xss-filters": "^1.2.7",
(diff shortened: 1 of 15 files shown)


alexlyp:
Validation for the Contractor information seems to be working correctly.

Line Item validation seems to work for Required, and Min/Max Length.  But it didn't seem to catch bad characters entered into a description line item field.

PR #1235, opened 2019-05-28
UI Fixes for Invoice Submission Page

### LINKED ISSUE #1206: [cms] remove button on bottom
Currently the Remove button for the rows is on the top. If more rows are created then can be viewed on one page, then its not possible to see what is being removed when the button is pushed. If the button was on the bottom of the table, it would then be easy to see what was being removed as the button is pushed.

### LINKED ISSUE #1215: [cms] Line item column widths
Currently Domain, Subdomain and Description are all the same width.

Ideally, Domain and subdomain should be a bit more narrow and then description should be wider (at least 1.5x the others)

### LINKED ISSUE #1216: [cms] Proposal token cell overflowing
Currently, when populated the proposal token cell overflows on top of the labor and expense cells when viewing the invoice details as an admin or user.

camus-code:
Closes #1206 
Closes #1215 
Closes #1216 

This commit addresses & resolves three recent issues brought up for the invoice submission component 

- Add / remove row buttons moved to the bottom of the datasheet table
- The proposal token column has been adjusted to use the multiline editor, as well as wrapping the text to prevent overflow
- Line item column widths have been adjusted. Domain and subdomain have been narrowed, while the description column & proposal token column have been widened

### DIFF
--- a/src/style/index.css
+++ b/src/style/index.css
@@ -1899,14 +1899,6 @@ form div.usertext-body.may-blank-within.md-container {
   list-style-type: none;
 }
 
-.multiline-cell-value {
-	word-break: break-word;
-}
-
-.multiline-cell-editor {
-	resize: vertical;
-}
-
 /* On screens that are 768px or less, apply the following styles */
 @media screen and (max-width: 768px) {
   /* hide side bar */
--- a/src/style/index.css
+++ b/src/style/index.css
@@ -1899,13 +1899,14 @@ form div.usertext-body.may-blank-within.md-container {
   list-style-type: none;
 }
 
-.description.cell {
+.multiline-cell-value {
 	word-break: break-word;
 }
 
-.description-multiline {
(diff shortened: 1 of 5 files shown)


src/style/index.css
```diff
@@ -1899,13 +1899,14 @@ form div.usertext-body.may-blank-within.md-container {
   list-style-type: none;
 }
 
-.description.cell {
+.multiline-cell-value {
```
fernandoabolafio: Can you move the styles regarding the multiline editor or any other "datasheet related" to `src/components/InvoiceDatasheet/styles.css`? 

It's important we keep the styles closer to the components using it for maintainability reasons and also to ease the migration of components after the redesign process.

camus-code: Yeah of course! Just pushed those changes. Didn't realize we had began migrating stylesheets to their respective component directory

Issue #1205, opened 2019-05-17
Politeiagui redesign/refactoring coordination and checklist


fernandoabolafio:
This issue describes the work in progress for the [redesign](https://github.com/decred/dcrdesign/issues/77) and refactoring implementation of politeiagui. It includes an overview of the design as well a check-list so anyone can track what has been accomplished so far. 
There are basically 6 milestones to be accomplished (listed at the end of this document), I expect 6 majors PRs for each of those and 'on-demand' PRs to deal with minor changes, fixes, and adjustments.
 
### Developers
The politeiagui devs are split into 2 teams. The team redesign is composed of @tiagoalvesdulce and me. It is possible and expected that a few overlaps will happen along the way.

### Pi-UI library
The [pi-ui](https://github.com/tiagoalvesdulce/pi-ui) library is being built to accommodate the new UI components and layouts.  The lib doc is automatically deployed [here](https://compassionate-lalande-d3ef5d.netlify.com)

### General Goals
- Implement the new design layout
- Remove the politeiagui dependency on the `snew-classic-ui
` library
- Allow features and context-specific elements to be customized from an external configuration source as described on #1067

### Technical Goals
- Make the codebase structure more context-driven in order to have high order components closer to the integration code
- Reduce the code dependency on Redux by using new React technologies such as hooks and context API

### Work in progress

- [x] **User**
    - [x] Login 
       - [x] Layout
       - [x] Integration
    - [x] Signup
      - [x] Layout
      - [x] Integration
      - [x] Custom config: toggle admin invite
   - [x] Reset password
      - [x] Request Layout
      - [x] Reset Layout 
      - [x] Integration
  - [x] Verify account email 
      - [x] Request Layout
      - [x] Reset Layout 
      - [x] Integration
  - [x]  Privacy Policy
       - [x] Layout 
       - [x] Custom config: change content based on preset type (cms, proposals)
  - [x] Detail
      - [x] Layout
      - [x] List information
      - [x] Manage Identity 
      - [x] Change username
      - [x] Change password 
      - [x] Preferences
  - [x] Paywall
      - [x] Layout
      - [x] Integration
      - [x] Custom config: enable/disable
   - [x] e2e integrations
        - [x] Routes (public, authenticated and admin-authenticated)

- [x] **Proposal**
    -  [x] Public List
        - [x] Layout
        - [x] Integration
        - [x] Search proposal capability
    - [x] Admin List
        - [x] Layout 
        - [x] Integration
    - [x] New
[...]

Issue #771, opened 2018-10-23
Privacy policy on separate page


xaur:
Separate page is the common practice I observe on the web. It makes the policy much more accessible (can be static page without js requirement). Should be linked from the footer.

camus-code:
Not working on this anymore. Free for anyone else to pick up @xaur

marcopeereboom:
@RichardRed0x can you write some copy for this?

xaur:
Politeia admin knowledge is required, only admins know what user data is collected, how it is stored, processed, etc.

marcopeereboom:
pubkeys, email, IP address and payment information is all that is silently recorded. Everything else lives in the backend that is there for all to see. One can easily set up a pseudo-anonymous account using tor that would hide more things. We will add email-less signups in the future.

RichardRed0x:
Here's a small update to the current copy: 

> The proposals.decred.org database stores your account email address, user-name, cryptographic identity public key(s), IP addresses and payment transaction details - and associates this data with all of your proposals, comments and up/down votes. Your email address will be kept private and will not be shared with any third parties. In the interests of transparency, data associating your public key with content contributions will be published openly in the Decred-proposals [repository](https://github.com/decred-proposals/mainnet/).

Is there anything else it should cover?

marcopeereboom:
I think this covers it all. Just need someone to wire it up.

xaur:
@RichardRed0x looks right. Two nits: 1) "username", 2) I suggest to type out the address of decred-proposals repository instead of a link.

marcopeereboom:
@camus-code can you wire this up?

camus-code:
Yeah definitely @marcopeereboom. Finishing up my open PR’s / issues then I’ll wire this up.

camus-code:
@RichardRed0x are you around? I'm wondering about the copy for the privacy policy when in CMS mode. Would you mind updating it for that as well whenever you get a chance?

marcopeereboom:
Think it is about the same.

camus-code:
Ok thanks @marcopeereboom. Just wanted to make sure there wasn't anything else we wanted to add

RichardRed0x:
As @xaur pointed out user-name was wrong in the original copy. I'm not sure what is meant by " I suggest to type out the address of decred-proposals repository instead of a link".

xaur:
@RichardRed0x I mean to use

```
[...]

PR #1251, opened 2019-06-03
[Redesign] User authentication pages and base structure for further changes

fernandoabolafio:
This PR deals with the first milestone of the [new design implementation](https://github.com/decred/politeiagui/issues/1205) which is:

- User authentication: Login, Signup, Reset Password, Verify account email, Privacy Policy, e2e integrations. 
The checklist of this PR can be found in the issue https://github.com/decred/politeiagui/issues/1205, where each milestone is better described.

I tried to keep the changes for the new version isolated as possible from the current application so we don't have major conflicts and functionalities regressions. To accomplish that the `src/index.js` is loading a different 'App' for each case. I used the "v2" suffix in the related folder/files in order to differentiate what is being used only in the redesign process.

- To start version 2 run: `yarn start-v2`
- To build version 2 run: `yarn build-v2`

The preview of the current state of the application can be found here: https://pi-redesign.authit.co. Please, refer only to the pages related to the "User authentication" context as they are the ones which are expected to be fully complete and functional in this PR. Other pages, such as Public Proposals or Create Proposal, are there just to give a better sense of the website overall structure.


MariaPleshkova:
Well. As we're still working on redesigning Pi, there were some small tweaks made on these screens.
image
"Reset password" was moved left. And the same was made for Sign up ("Resend verification email" was moved left too).
image

**Now regarding more specific things:**

**1.** Log in/Create a new account titles need to be Regular font, not Semi-bold.
Also their size should be 28px.

**2.** The size of the normal text should be 16px (Email, password, etc.)
image

**3.** We also need to add hover effects on all clickable elements (blue/gray links). I suggest to change them to Semi-bolds.
image

**4.** Change the color of “Privacy policy” sign to _#8997A5_

**5.** Add hover effect to non-active header item (Log in/Sign up). Hover effect - _#091440_

**6.** Main white section (rectangle) should be on the center of the screen (up/down and left/right)
image

**7.** Notification text should starts with the capital letter. Its size has to be 13px.
The size of fields titles (that were moved up) - 13px as well.
image

**8.** Not sure if Privacy Policy screen is already finished, but just in case:
It has to be a pop up.
image

Basically, amends for the rest created screens are the same as for Log in screen above (fonts sizes/weights, color, etc.)

So these are the ones I've noticed.

lukebp:
One thing I noticed is that the "required" text causes the form entry fields below it to be pushed down the page.

fernandoabolafio:
@MariaPleshkova thanks for your comments, they are all pretty straight forward to get fixed. I have comments about #6 and #8:
- About #6 (Main white section (rectangle) should be on the center): I first placed the main area at the center of the page. For login it and other smaller contents such as "Resend verification email" it looks nice. However, for the Signup form, for instance, it causes the page content to overflow the viewport, thus requiring the user to scroll the page to fill up the form. This can be avoided just by having the Form placed closer to the top of the page. What do you think?
- About #8 (Not sure if Privacy Policy screen is already finished...): I chose to go with a separated page for privacy policy because of this: https://github.com/decred/politeiagui/issues/771. We recently decided to move the Privacy Policy from a pop up to a separated page, so I figure that would be better to stick with that.

MariaPleshkova:
@fernandoabolafio 

- About 6: so do you suggest to keep only long forms closer to the header? Or to leave all of them as you made at first?

- About 8: hm, but in this case we have to add at least Back button there. As when user opened Privacy policy and then he wants to go back, he needs to click on Log in/Sign up again. Also what will happen if user already filled the fields and then decided to click on Privacy Policy? Will he lose this entered data?
IMHO there are only 1-2 paragraphs there, so not sure that we need a separate page for this. 
@linnutee what do you think?

Regarding an updated link, I'll send this asap. Finishing with some last tweaks.

fernandoabolafio:
@MariaPleshkova thanks for the quick reply. 
- About 6: I think all forms for the Authentication pages (Signup, login, resend email etc) could be closer to the top. IMO it's easier to read and we keep the form position consistent across the different pages.
- About 8: I don't have strong feelings about this. I am okay with both approaches. However, I tend to like the modal better for the reasons you mentioned above.

linnutee:
Regarding 8, i'd say modal is more fitting as there's quite little content. If the privacy policy happens to increase in content size in the future or become more significant then there's some more reason to make a subpage for it. 

I'll also mention regarding graphic assets, logo for example – wherever possible use svg's for icons or any vector based assets.  Then there's less play around the @2x and @3x png's.

tiagoalvesdulce:
@linnutee @MariaPleshkova we are building a UI lib with live documentation ([pi-ui](https://github.com/tiagoalvesdulce/pi-ui)). Can you guys take a look and write issues there? I think it's more productive that way. Also, we are missing the UI example of some components like the Dropdown in the original Adobe XD document.

fernandoabolafio:
@linnutee @MariaPleshkova I addressed your main concerns regarding the font-size and the privacy policy. Now the privacy policy is displayed in a modal but there is also a "permalink" which redirects to the full route in case we need to link from other places.
Screen Shot 2019-06-11 at 21 43 29

It is missing the "i" icon from the specs because we need to add that to the pi-ui lib yet. I will make a point to address those minor details later if that is okay for you.

lukebp:
A few comments.  The first two are directed towards @linnutee @MariaPleshkova.

1. Anytime a form field fails the validation, error text such as "Required" or "Invalid Email" pops up.  When the error text appears, all of the elements below it get shifted down.  Is this intentional?

2. This is the text that appears on the screen when you sign up.

> Politeia will send you a link to verify your email address. You must open this link in the same browser. After verifying your email, Politeia will create your Politeia “identity”, which consists of a public/private cryptographic key pair and browser cookie. This is necessary to verify your identity and allow submission of proposals, commenting, voting, and other Politeia functions. After completing the signup process, you can export your identity (public/private keys) to another browser at any time.

tbh I didn't read any of it.  I just pushed the "I understand, sign me up" button.  A big block of text like 
that is pretty unappealing and I'm not sure very many people will actually read it.  My suggestion would be to keep it short and to the point:

"Politeia will send you a link to verify your email address. You must open this link in the same browser." 

And then save the user identity explanation for the user onboarding modal that will appear the first time the user logs in.

3. @fernandoabolafio: is the "Verify Email    [dev]" element suppose to be there?
Screen Shot 2019-06-11 at 4 40 11 PM

fernandoabolafio:
@lukebp I will answer you here again just for documentation purposes. Yes, this [dev] tag is intentional to wrap actions which will no be available in a production website. This is just to simplify the process of going through the screens without leaving the website.

src/componentsv2/Sidebar.jsx
```diff
@@ -0,0 +1,28 @@
+import React from "react";
+import { Sidebar as UISidebar, Button, Card } from "pi-ui";
+import { withRouter } from "react-router-dom";
+import LoggedInContent from "./LoggedInContent";
+import StaticMarkdow from "./StaticMarkdown";
```
tiagoalvesdulce: There's a typo here

src/componentsv2/Sidebar.jsx
```diff
@@ -0,0 +1,28 @@
...
+        <Button
+          className="margin-bottom-s"
+          onClick={() => history.push(`/proposals/new`)}
+          fullWidth
+        >
+          Submit Proposal
+        </Button>
+      </LoggedInContent>
+      <Card paddingSize="small">
+        <StaticMarkdow contentName={aboutContent} />
```
tiagoalvesdulce: Same as above

src/containers/User/Login/Form.jsx
```diff
@@ -0,0 +1,104 @@
...
+  return (
+    <>
+      <FormWrapper
+        initialValues={{
+          email: "",
+          password: ""
+        }}
+        loading={!validationSchema}
+        validationSchema={validationSchema}
+        onSubmit={async (
```
tiagoalvesdulce: I think these functions shouldn't be inline in the JSX, can you create a named function in the component instead?

src/containers/User/ResetPassword/RequestForm.jsx
```diff
@@ -0,0 +1,86 @@
...
+  const [email, setEmail] = useState(false);
+  const success = !!email;
+  return (
+    <>
+      <FormWrapper
+        initialValues={{
+          email: ""
+        }}
+        validationSchema={validationSchema}
+        onSubmit={async (values, { setSubmitting, setFieldError }) => {
```
tiagoalvesdulce: Same as above.

src/containers/User/ResetPassword/ResetForm.jsx
```diff
@@ -0,0 +1,94 @@
...
+  if (invalidParamsError) {
+    throw invalidParamsError;
+  }
+
+  return (
+    <FormWrapper
+      initialValues={initialValues}
+      validationSchema={validationSchema}
+      loading={!validationSchema}
+      onSubmit={async (values, { setSubmitting, setFieldError, resetForm }) => {
```
tiagoalvesdulce: Same.

src/containers/User/Signup/Form.jsx
```diff
@@ -0,0 +1,185 @@
...
+        title={"Before you sign up"}
+        confirmMessage="I understand, sign me up"
+        onClose={onModalCancel}
+        onConfirm={onModalConfirm}
+      />
+      <FormWrapper
+        initialValues={initialValues}
+        loading={!validationSchema}
+        validationSchema={validationSchema}
+        onSubmit={async (...args) => {
```
tiagoalvesdulce: Same.

src/containers/User/VerifyEmail/RequestForm.jsx
```diff
@@ -0,0 +1,107 @@
...
+        confirmMessage="I understand, continue"
+        onClose={onModalCancel}
+        onConfirm={onModalConfirm}
+      />
+      <FormWrapper
+        initialValues={{
+          email: ""
+        }}
+        validationSchema={validationSchema}
+        onSubmit={async (...args) => {
```
tiagoalvesdulce: Same.

src/containers/Header/hooks.js
```diff
@@ -0,0 +1,16 @@
+import * as sel from "src/selectors";
+import * as act from "src/actions";
+import { useRedux } from "src/redux";
+
+const mapStateToProps = {
```
tiagoalvesdulce: You should include the `loggedInAsUsername` selector here so the navbar keeps consistent with the username if it changes.

fernandoabolafio: Not sure I understand your comment. I am using the `loggedInAsUsername` selector

tiagoalvesdulce: I was in an outdated version. Sorry about that.

tiagoalvesdulce:
Two minor things that I noticed:

1. The space between the `Already have an account?` and `Log in!` is too big. Should be equal the space between the `Dont have an account?` and `Create here!`. See prints below:

space

Screen Shot 2019-06-18 at 11 57 36

2. The hover effect on links is moving elements near to it. Also, it doesn't have a smooth transition. See:

bug1

fernandoabolafio:
@tiagoalvesdulce fixed your comment about the span tags and the '1'. For 2 I created an issue in the pi-ui repo: https://github.com/tiagoalvesdulce/pi-ui/issues/56.

PR #1255, opened 2019-06-04
Adjust user layout to common layout rules

fernandoabolafio:
This PR tweak some aspects of the layout in the User Page. This page was using its own classes for wrapping the general content, causing a different behavior from other screens when the view is resized.  In summary what this PR accomplishes is:
- Use the same "content" class for the main container of the user's page.
- Remove useless markdown tags and css rules.
- Move user page related styles to it's own file and closed to the components which are using it.
- Remove the rule of "height: 100%" from the appContainer, what was causing the lazy loading of proposals to fail.


tiagoalvesdulce:
Monospace items are losing the left alignment on smaller screens.

Screen Shot 2019-06-05 at 10 31 01

PR #1265, opened 2019-06-10
Fixes proposal submit after image error

### LINKED ISSUE #1238: Cannot submit proposal after image error
I am not able to submit a proposal after I receive an "image exceeds the maximum size"  error.
There are multiple errors here:

How to reproduce:
- Go to the submit proposal route
- Attach a file larger enough to receive the error message. (Ps: at the moment the image is selected the client validation occurs successfully displaying an error message  at the top of the attachment area but the 'submit' button is not disabled)
- Submit the proposal
- The client validation occurs again but now displaying the error message at the top of the page. 
- Now the 'submit' button is disabled and even if I remove the errored file it doesn't return to the enabled state.

We need to make sure:
1. Client-side validation occurs at the form filling time, no need to wait for the user pressing the 'submit' button.
2. Errors are not duplicated in two areas of the screen.

victorgcramos:
Closes #1238

### DIFF
--- a/src/components/Form/Fields/FilesField.js
+++ b/src/components/Form/Fields/FilesField.js
@@ -18,23 +18,6 @@ export class FilesField extends React.Component {
     };
   }
 
-  onChange = ({ ...props }) => {
-    const {
-      input,
-      meta: { dispatch },
-      policy
-    } = this.props;
-    const newFilesList = props.hasOwnProperty("remove")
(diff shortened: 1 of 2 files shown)


victorgcramos:
Oh, my bad, I've made a wrong interpretation about the issue. Now I think we're good to go.

PR #1280, opened 2019-06-19
Fixes form crashing when typing invalid chars

### LINKED ISSUE #1278: [pi/cms] Submit form crash when inputting invalid characters
The form is crashing when inputting invalid characters in a sequence. Happens both in cms and pi.

crash

victorgcramos:
Closes #1278 

The problem occurred when trying to compare a null string returned from Regex.


thi4go:
Fixed the issue, thanks for the PR @victorgcramos. LGTM

PR #1291, opened 2019-06-26
Improve proposal details page efficiency

### LINKED ISSUE #1230: Proposal details page inefficiencies
I'm seeing some inefficiencies on the proposal details page that could be improved.

1. There are duplicate policy requests.
2. The policy, vote status, and comments requests only get sent after the proposal details request has returned (and it looks like its also waiting for some post processing to be completed as well).  Is there a reason for this?  You could cut the total load time in half if you fired off these requests concurrently.

Screen Shot 2019-05-24 at 4 07 19 PM

camus-code:
Resolves #1230

This commit updates the the policy, vote status, and comments requests so that they fire concurrently on the proposal details page.

This commit also fixes the bug where duplicate policy requests were being called.

### DIFF
--- a/src/actions/api.js
+++ b/src/actions/api.js
@@ -1079,7 +1079,12 @@ export const onFetchProposalVoteStatus = token => dispatch => {
     })
     .catch(error => {
       dispatch(act.RECEIVE_PROPOSAL_VOTE_STATUS(null, error));
-      throw error;
+      // Discard error when proposal is not public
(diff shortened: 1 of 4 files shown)


fernandoabolafio:
@camus-code thanks for this PR. However, we're halting the updates for the old GUI code unless they are critical fixes. Any other optimization or UI tweak we will not be moving forward. Thus, I am closing it.

PR #1292, opened 2019-06-26
Remove dcrdata bestblock request

### LINKED ISSUE #1288: Remove dcrdata best block request
Privacy Badger has been flagging dcrdata as a tracker.  decred/politeia#928 added the best block to the vote status response so that the gui doesn't have to pull the best block from dcrdata.

camus-code:
Closes #1288 

This commit removes the usage of dcrdata for fetching the last block height. The new block height used by the GUI comes from the vote status response, per https://github.com/decred/politeia/pull/928


fernandoabolafio:
@camus-code thanks for this PR. However, we're halting the updates for the old GUI code unless they are critical fixes. Any other optimization or UI tweak we will not be moving forward. Thus I am closing it.

PR #1295, opened 2019-06-27
[Redesign] User Details Page

tiagoalvesdulce:
Not implemented:

* Rescan (depends on https://github.com/decred/dcrdesign/issues/159)
* Preferences Tab (depends on https://github.com/decred/dcrdesign/issues/159)
* Manage Identity (depends on https://github.com/decred/dcrdesign/issues/159)

**Edit** To run the new version use: `yarn && yarn start-v2`


lukebp:
Noticed a few things.

**Change password form**
- Try clicking the elements with the input field name (ex. Current Password).  It doesn't let you.  You have to select outside of it to start typing.
- When a response for the change password request comes back, the loading wheel icon switches back to the "Change Password" button while the form is in the middle of closing.
- Is this the correct spacing between the error message and the next field item title?  It seems squashed and almost overlapping to me.
image


**Change username form**
The change user form as all of the same issues as the change password form.

tiagoalvesdulce:
* Try clicking the elements with the input field name (ex. Current Password). It doesn't let you. You have to select outside of it to start typing. **Happens in the auth routes as well, captured as an issue here: https://github.com/tiagoalvesdulce/pi-ui/issues/65**.
* When a response for the change password request comes back, the loading wheel icon switches back to the "Change Password" button while the form is in the middle of closing. **Fixed**.
* Is this the correct spacing between the error message and the next field item title? It seems squashed and almost overlapping to me. **This is a problem that also happens in the auth routes after https://github.com/tiagoalvesdulce/pi-ui/pull/60. Maybe @MariaPleshkova can help us to improve the margin here**.
* Changing the username reflects only in the name on the top bar but not in the user account title. **Fixed**.
* Clicking on "change username" causes a quick flickering of the modal opening and closing. Steps to reproduce. **Couldn't reproduce**.
* Layout missing a few things from the specs. **I am improving this, I was on an old version of the design specs. Depends on https://github.com/tiagoalvesdulce/pi-ui/issues/66**.

MariaPleshkova:
@tiagoalvesdulce created an example with appropriate spacings. Please, apply to all similar pages with fields (register, login, etc.)
https://xd.adobe.com/spec/e8dc6d15-3d08-4926-5d74-ed56adff6631-9eed/

tiagoalvesdulce:
* Try clicking the elements with the input field name (ex. Current Password). It doesn't let you. You have to select outside of it to start typing. **Fixed**
* Is this the correct spacing between the error message and the next field item title? It seems squashed and almost overlapping to me. **Fixed**

src/hooks/useMediaQuery.js
```diff
@@ -0,0 +1,75 @@
+import { useEffect, useState } from "react";
+
+// This variable will be true once the server-side hydration is completed.
+let hydrationCompleted = false;
+
+function deepEqual(a, b) {
+  return a.length === b.length && a.every((item, index) => item === b[index]);
+}
+
+export function useMediaQuery(queryInput, options = {}) {
```
fernandoabolafio: I like this hook a lot. In fact, I think we can build something even more abstract on top of it to create a better layout consistency. This new hook would come with some handy values to verify and build the layout against it. Something like:
```js
const { 
   size // extra-small/small/regular/etc
} = useResponsiveLayout();
```
Don't need to add it to this PR but it is something we can add during the next changes.

src/lib/queryString.js
```diff
@@ -42,3 +43,34 @@ export const removeQueryStringsFromUrl = (url, parameter, parameter2) => {
     .replace(new RegExp("([?&])" + parameter2 + "=[^&]*&"), "$1");
   window.history.pushState({ path: newurl }, "", newurl);
 };
+
+export function useQueryString(key, initialValue) {
```
fernandoabolafio: Perhaps we should move this to the hooks folder?

src/lib/queryString.js
```diff
@@ -42,3 +43,34 @@ export const removeQueryStringsFromUrl = (url, parameter, parameter2) => {
...
+  const [value, setValue] = useState(getQueryStringValue(key) || initialValue);
+  function onSetValue(newValue) {
+    setValue(newValue);
+    setQueryStringValue(key, newValue);
+  }
+
+  return [value, onSetValue];
+}
+
+export function useQueryStringWithIndexValue(key, initialIndex, values) {
```
fernandoabolafio: Same as above

MariaPleshkova:
**User details pages modifications** 
Realized the way grouping is presently done has issues, as there are a lot of not related items mashed together. So we decided to make 4 high level groups as subpages and populating those with the relevant items (Identity, Account, Credits, Preferences).

**Manage identity modifications**
Reduced one action, so now both buttons (Create new identity and Import identity) could be accessed here already, and there will be no need to double click (first opening modal “Manage identity”, then the action).

+Also notice, please, that active tab's yellow color was changed to cyan one (#2ED8A3) everywhere.

Preview: https://xd.adobe.com/view/d89e71a8-6b74-42f0-6434-93830ca703d3-aeaf/
Specs: https://xd.adobe.com/spec/94a4faec-1167-4b98-79b0-94f73092e342-e295/

tiagoalvesdulce:
OK Maria, this was merged before the newest design changes. Tab color was fixed on pi-ui and new issues should be created for the user details and manage identity modifications.

PR #1299, opened 2019-07-01
Changes login to use username instead of email

### LINKED ISSUE #1297: Change login to use username
decred/politeia#940 updates the login to use username instead of email.  This change will been to be reflected in the gui as well.

thi4go:
closes #1297


camus-code:
LGTM @thi4go

PR #1302, opened 2019-07-03
[cms] Add pay invoices admin button

alexlyp:
This adds a button on the generate payouts admin screen that uses the pay invoices endpoints.

This typically will only be used in the event of the address watcher not properly updating an invoice to paid.


src/components/GeneratePayouts/GeneratePayoutsPage.js
```diff
@@ -60,6 +64,20 @@ const GeneratePayoutsPage = ({
...
+            className="inverse"
+            onClick={e =>
+              confirmWithModal(modalTypes.CONFIRM_ACTION, {
+                message:
+                  "Are you sure you want to set all of these approved invoices to paid?"
+              }).then(confirm => confirm && onPayApprovedInvoices()) &&
+              e.preventDefault()
+            }
+            text="Set invoices to paid"
+            isLoading={false}
```
fernandoabolafio: The loading should not default to `false`.  It needs to be the correct value indicating whether the paid action is loading or not.

src/components/GeneratePayouts/GeneratePayoutsPage.js
```diff
@@ -60,6 +64,20 @@ const GeneratePayoutsPage = ({
           {payouts && <PayoutsTable payouts={payouts} />}
         </div>
         <div id="csv-hidden-div" style={{ display: "none" }} />
+        <div className=" payouts-btn-pay-invoices">
+          <ButtonWithLoadingIcon
```
fernandoabolafio: As far as I can tell, this button will be visible and clickable even after all the invoices were marked as paid. Shouldn't it be rendered or enabled only if the `numOfApprovedInvoices > 0` ?

src/components/GeneratePayouts/GeneratePayoutsPage.js
alexlyp: Well there really isn't any loading going on, as it should be handled immediately.  Should I just change which Button I'm currently using in that case?

src/components/GeneratePayouts/GeneratePayoutsPage.js
alexlyp: Good point, I'll add that check.

src/components/GeneratePayouts/GeneratePayoutsPage.js
fernandoabolafio: The click of this button will trigger a modal, which after the confirmation will be closed and fire a request against the API, right? Normally, what we do is to use the button loading to display that the request is ongoing to give feedback for the user. This also prevents the button to be clickable while the request hasn't finished yet.

PR #1304, opened 2019-07-09
[Redesign] Add website paywall

tiagoalvesdulce:
To run: `yarn && yarn start-v2`

Depends on: https://github.com/tiagoalvesdulce/pi-ui/pull/79


fernandoabolafio:
@tiagoalvesdulce the paywall is working great and the modal is looking good for mobile. I've just noticed a few things for the hamburger bar:

1. It needs a shadow on the right side of the container. See here:
Screen Shot 2019-07-12 at 01 12 24

2. The hamburger has it's own 'header'. By using the general header as also the hamburger header, there is a problem that when the page is scrolled down, you cannot close the menu.
Jul-12-2019 01-16-25

fernandoabolafio:
@tiagoalvesdulce I am finally done with the 'truncate' PR (https://github.com/tiagoalvesdulce/pi-ui/pull/79). With that, a few things need to be modified in this PR:
1. Every `pi-ui` component, except for `Link`, which exposes a `truncate` prop, needs an ID whenever `truncate` is `true`. There is a prototype checking for that which will throw an alert in the console otherwise. Link components have a simplified truncate version which doesn't require the ID to be specified.
2.  The `CopyableText` used for displaying an address exposes a `truncate` API now in case you want to use it. However, for addresses, I recommend you to not truncate the text and let the element display it in the default mode, which will break the address in the next line in case it doesn't fit. I believe is a better UX to always display the full address to where send payments.
3. Add the `truncate` property for the faucet link.

lukebp:
When you click on the copy address button, the user is not seeing any type of feedback to indicate that the address was successfully copied.  Should some type of "copied to clipboard" message appear?

tiagoalvesdulce:
@lukebp it is captured as an issue here: https://github.com/tiagoalvesdulce/pi-ui/issues/83

lukebp:
Ok, nice.  Guess I should start checking the pi-ui issues first.

src/componentsv2/HeaderNav/HeaderNav.jsx
```diff
@@ -0,0 +1,46 @@
...
+            to="/user/signup"
+          >
+            <Text className={styles.navLinkText}>Sign up</Text>
+          </NavLink>
+        </nav>
+      )
+  );
+};
+
+HeaderNav.propTypes = {
```
fernandoabolafio: Remove this if you're not adding prop types checking here

src/componentsv2/NavigationDrawer/NavigationDrawer.jsx
```diff
@@ -0,0 +1,37 @@
+import { classNames } from "pi-ui";
+import PropTypes from "prop-types";
+import React from "react";
+import HeaderNav from "../HeaderNav";
+import SidebarContent from "../SidebarContent";
+import styles from "./NavigationDrawer.module.css";
+
+const NavigationDrawer = ({ show, fullScreen }) => {
+  if (show) {
+    window.scrollTo(0, 0);
```
fernandoabolafio: Can you encapsulate this functionality of blocking/unblocking the scroll it into a hook or just a utility function?

lukebp:
@MariaPleshkova I have one nit with the wording of "Waiting for approval" when a payment is pending.  "approval" gives the impression that the payment needs to be approved by someone.  This is reinforced by the fact that we use the "approve" terminology when we're talking about a proposal that needs to be approved by an admin in order to be made public.  A more accurate way to phrase this would be "Waiting for confirmations".  Can we change the wording of this?

MariaPleshkova:
@lukepb yeah, agree! I think we can.

MariaPleshkova:
**1.** Add "No ... tickets" text to the pages with no tickets to show
image

Example (16px, Regular, #596D81): 
image


**2.** Could we move it closer to the right edge?
image



**3.** Yes/No should be semibold
image

Spec:
image

Also, will we replace categories tab with dropdowns on tablet and mobile later?
The rest seems ok for me. Good job!

tiagoalvesdulce:
1. It's not related to this PR, will be fixed in the future. Caputred as an issue: https://github.com/decred/politeiagui/issues/1307
2. Fixed to match the Header padding:
Screen Shot 2019-07-17 at 13 35 03

3. It's not related to this PR, will be fixed in the future. Captured as an issue: https://github.com/tiagoalvesdulce/pi-ui/issues/93

@MariaPleshkova Yes, we will be replacing the tabs with dropdowns on tablet and mobile soon. There's an issue on pi-ui to create the Dropdown component https://github.com/tiagoalvesdulce/pi-ui/issues/89 and @camus-code is already taking care of it.

PR #1309, opened 2019-07-22
[redesign] Implement comments

fernandoabolafio:
This PR implements the Comment feature following the redesign specs and refactoring the code accordingly to the new codebase structure for v2.
The list o items covered are:
- Posting new comments
- Listing comments
- Single thread mode (permalink)
- Sorting comments
- Replying to comments
- Blocking comment actions for specific proposals states (voting finished, abandoned, etc)
- Improve loading placeholders (comments and proposals)

Known issues:
 - Clicking on the comments link on a proposal doesn't automatically scroll to the comments list (will be addressed before this PR get merged)
 - Download comments bundle not working: I will suggest a design modification so I am putting it on hold until there.


lukebp:
1. @MariaPleshkova I wanted to give some feedback on the comment design.  It could be that I'm just used to reddit comments, but the comment upvote/downvote buttons and score being on the far right side stuck out to me.  In order to see the comment score for a specific comment you have to move your eyes to a different part of the screen, which seemed a little odd.  It can also be difficult to tell which comment score corresponds to which comment when there are a large number of comments on the screen.

image


The below comments are directed at @fernandoabolafio.

2.  If you're logged out, hovering the cursor over the comment shows the not allowed symbol.  I like how Reddit handles this situation.  If you try to upvote the comment while logged out, it lets you click the upvote button and then the login modal pops up.  What do you think?  This is probably out of the scope of this PR, but if you agree let's open an issue for is so that we don't forget.

3. When you click the comment permalink button, it actually loads the comment permalink.  This seems odd to me.  The behavior I was expecting was for it to copy the link to the clipboard.

4. I believe the below error is caused by an identity issue.  The error message should be better.

image

fernandoabolafio:
@lukebp Addressed your comments with the exception of 1 which I created an issue #1311, once the design team has reviewed it, I will implement the missing bits.
With the latest changes, you won't be able to submit proposals or comments if you have an identity invalid or missing the paywall payment.
I also added the persistence of comments into session storage, so you can refresh the page and the text you're typing will still be there.

MariaPleshkova:
@lukebp regarding upvote/downvote, maybe you're right. I created some new options of it and sent it to @linnutee for review

fernandoabolafio:
@MariaPleshkova I think I covered all of your comments. Just two things are changed/missing:

- The comment sorter dropdown I had to use it a bit larger to better accommodate the contents. 
- The "formatting help" is missing the design specs (not the link, but the content to be shown after). That's why it is not included there yet. This  is how it currently looks:
Screen Shot 2019-07-30 at 19 49 57

linnutee:
Re: comments @lukebp 
I’ve asked for those to be redesigned a while back and indeed one of the drivers was to iterate these to be more fitting for Pi's context instead of reddit's karma hunting.

Positioning: 
- Being on the right side does become difficult to track only when there’s no content populating the comment area. In that example there’s only 1 word comments, so that won't give a good result indeed. Right now the `name + date` `up/down thumbs` `replies` and `link` are positioned in four corners with the intent of framing each comment. This way there’s no need to add any separator line elements, as an invisible frame is drawn. 
- In actual Pi release, majority of comments have at least few sentences of length, so this should be enough to populate the comment area and not have those rating + link elements on the right side drift away.
- If the drift does happen to become an issue with actual content, I would recommend first trying to finess the spacing between the comments.  
- Not sure if the screenshot is from the implementation, but the design is a bit off there. The intention was to show both up and down ratings for the comments to but a slight more focus on the opinion and steer clear from just score hunting as it goes with reddit. So these should show both. 

Screenshot 2019-08-01 at 16 24 46

fernandoabolafio:
@linnutee we're not showing both up and down votes right now due to a technical restriction. Once https://github.com/decred/politeia/issues/957 is done we will tweak that in the gui side.

MariaPleshkova:
@fernandoabolafio I think it can be smth like this
image

MariaPleshkova:
@fernandoabolafio well, I guess these are the last ones for this PR (but overall looks pretty good):

1. make the distance between "comments" and a number 10px
image

2. found this issue on designs also. Change the color of the text inside the field to #091440, please. Bc it looks like it's not active at that moment, but it is
image

3. when Preview is active, I think, we can remove this field's frame and, therefore, move it a little bit to the left (to the left container border). And in this case change the color to #596D81. (sorry for that again)
image

4. add the line (#E6EAED) in between and move number closer to the icons
image

how it has to be:
image

PR #1313, opened 2019-07-26
[Redesign] Add proposal credits flow

tiagoalvesdulce:
This PR addresses the proposal credits flow.

Still missing the proposal credits payment polling because in the "old" pi it's being done in more than one place inside an IntervalComponent. I'm thinking of dropping this component moving this polling to the application state level. What do you think @fernandoabolafio ?


fernandoabolafio:
Sounds good to me moving the polling to the state level @tiagoalvesdulce. 

I haven't fully reviewed your PR as it is still a 'wip'. However, while running I found some layout mismatches with the specs. Check it out so you can have it fixed before finishing it up:

Screen Shot 2019-07-26 at 09 16 31

fernandoabolafio:
@tiagoalvesdulce I managed to get the polling working for proposal credits again. I figured it was better to keep the polling in the component level so we can reuse most of what we already had from the previous code. Although I rewrote most of it to use only hooks and functional components. The pollings will be replaced by WebSockets so we shouldn't waste too much time refactoring those chunks of code.

There are a few problems I encountered and aren't solved yet:
- The "Paid"/"Not Paid" message is being shown based on the data from the current logged in user. This info should be used from the user/:id route and not for the current user.
- The same occurs for the proposal credits. The number of credits being show is from the current user data and not from the user fetched in the user details page. 
- The Credits tab needs to be either disabled or hidden when the user accessing it is not the owner or an admin.
- We need to add the pending transaction into the payments history table.

fernandoabolafio:
@tiagoalvesdulce all the issues stated above should be fixed now.

fernandoabolafio:
@lukebp  @MariaPleshkova This PR is ready for a review. 

We changed a few things from the design specs which we think results on a better view flow for mobile screen:

- For mobile screens ( width < 560px) we displayed the content vertically instead, which makes the text copy less squeezed:
Screen Shot 2019-08-01 at 13 01 37

- Also for mobile, we used a vertical table which makes all the content visible, without the need to be scrolling the table:
Screen Shot 2019-08-01 at 13 01 54

lukebp:
@tiagoalvesdulce found a couple of issues.

1. If you're on the user details page and click the "+" proposal credits button the entire page will reload instead of just changing tabs.  Is this the desired behavior?  There is also a weird flickering of the screen when you do this.  Some elements flicker on and off for a second.  This only happens the first time that you click the "+" button.  After that it loads fine.  You may need to open an incognito window to reproduce it.

2. The registration fee says 0.4 on the credits tab but says 0.1 on "Complete your registration" modal.

3. When you click the button to copy the paywall address there is no "Copied to clipboard" tooltip.

4. Sent transaction hash goes off screen when on mobile.

image

5. It looks like the you're polling dcrdata for the registration fee confirmations.  We want to eliminate any dcrdata requests from the gui because plugins like Privacy Badger treat it as a tracker and block the requests.  You can get the registration payment info from the `Me` endpoint.  Can we poll that endpoint instead?

6. When I went to purchase my second round of proposal credits, it took 30 seconds for the pending purchase to show up in the credit history list.  Is that the intended behavior?

I also thought that showing the full "Thu, 01 Aug 2019 21:57:56 GMT" for the date was a bit excessive.  Seems like something like just "01 Aug 2019" would be cleaner.  That's up to design though.

MariaPleshkova:
@fernandoabolafio @tiagoalvesdulce 
1. As these buttons are smaller then our main ones, the font size need to be 13px instead of 16px
image

2. The new button appeared? (Rescan button)
I'd better put it next to the "Purchase more button", not under it. In a horizontal line I mean

3. The color of the titles should be #596D81
image

4. I think that when there's only 1 page, no need to show this page switcher at all
image

5. the header of the page has been changed as well
your implementation:
image

spec:
image

6. add the icon before the title, change an input field, change Back button
image

spec:
image

7. move Plus icon to the center of the button
image

8. add "See the status of purchasing" button
image

Now replying to @fernandoabolafio 's comment above
1. Nice idea to display the content vertically on mobile screens. Looks good. The only thing is that I'd better move this gray description text straight under the title, so the order will be title-description-status(paid/not paid or the number of proposals)

2. I like the idea as well, but it's hard to see where the next credit starts. I'd better put each and every credit into the "box" (also the color of the lines should be #F3F5F6)
Like this:
image

And the same for mobile view.

XD file to check the spacings, etc.
[my account_tablet.zip](https://github.com/decred/politeiagui/files/3461887/my.account_tablet.zip)

fernandoabolafio:
@MariaPleshkova thanks for your comments. We will be addressing them asap. Just about the #5 we need to clarify a few things. In the new Specs, it says "My Account" instead of the username. However, this route is mean to be used for two purposes, one is for account management for the current user. Another one is for visualizing the details of another user. So, in summary, we have, 3 major purposes in this account section:
1. A logged-in user can view and manage his account (change username, change password, buy proposals credits, manage identity)
2. An admin can view and execute a few actions for another user account (expire verification tokens, rescan proposal credits, deactivate account)
3. A random person who access another user's details only for visualization purposes. In this case, the tab of the credits is not available and maybe the account tab shouldn't be there either.

That said, using "My Account" as the title doesn't make sense for when the user who is accessing the page is != from the given user details. Let me know what are your thoughts. I am even thinking we could have two different routes:
- One which is only for account management (the one from the specs).
- Another one which is much simpler just for visualization of the user details.

tiagoalvesdulce:
@MariaPleshkova what's your position about Fernando's comment? What should we do?

tiagoalvesdulce:
@MariaPleshkova What's the function of the add "See the status of purchasing" button? The table showing the statuses are right below.

tiagoalvesdulce:
@lukebp 

1. Fixed
2. Fixed
3. Fixed
4. Fixed
5. We are polling from `api/v1/proposals/paywallpayment` for proposal credits.
6. Yeah, it happens because there's is a delay between the moment the payment is done and the detection of this payment by the polling.
7. Fixed date format.

@MariaPleshkova 

1. Fixed
2. Fixed
3. Fixed
6. Button solved. Icon not added to pi-ui yet. We can create an issue to solve it later. Input: https://github.com/tiagoalvesdulce/pi-ui/issues/107

The "My Account" discussion can be done later and we can create an issue to implement the changes required.
 
Regarding the Table comments: https://github.com/tiagoalvesdulce/pi-ui/issues/101

fernandoabolafio:
@tiagoalvesdulce everything is looking good but I found a few things which were not addressed from the comments:
- Invert the order of the content in the account section for mobile view. As per Maria comment:
> [...] The only thing is that I'd better move this gray description text straight under the title, so the order will be title-description-status(paid/not paid or the number of proposals)
- Move Plus icon to the center of the button
- Faucet tx is truncated needlessly and yet it is still overflowing on mobile screens:
on desktop:
Screen Shot 2019-08-11 at 18 39 28
on mobile:
Screen Shot 2019-08-11 at 18 40 47

MariaPleshkova:
@fernandoabolafio @tiagoalvesdulce my idea was smth like this:

1. Personal account of logged-in user
image

2. Admin view of another user's account
image

3. Random person's view of another user's account
image

MariaPleshkova:
@tiagoalvesdulce if I got @linnutee right, this button ("See the status of purchasing") should open a modal with the last purchasing and from there user is able to cancel it
image

tiagoalvesdulce:
@MariaPleshkova the User Page will be updated in a future PR. Also, I don't think we need a "See the status of purchasing" button 'cause I think it's impossible to cancel a transaction once it's been placed.

MariaPleshkova:
@tiagoalvesdulce ok, let's remove it then

PR #1314, opened 2019-07-29
[cms] Add Invoices tab to user details for admin

alexlyp:
~Requires https://github.com/decred/politeia/pull/959~

This adds the invoices tab to a user details page.  This tab
should only be accessible to administrators.


alexlyp:
1) Hrm ok need to figure out how to display that error.
2) Cool, I'll add some checks to ensure that the request is only done as an admin

alexlyp:
So I'm not quite sure how to handle the 404's, how does pigui typically deal with them?

fernandoabolafio:
> So I'm not quite sure how to handle the 404's, how does pigui typically deal with them?

The problem here is not the 404 itself. The issue is because it is not displaying any error returned from the API call. You can do it just like any other part of the application, use the appropriate selector to get the error for the desired API response and show it instead of the content in case of error.

In your specific case it would look like:

- create a selector: 
```js
const userInvoicesError = getApiError("userInvoices")
```

- add it to your connector:
```js
const mapStateToProps = {
...
userInvoicesError: sel. userInvoicesError
}
```

- and then, display the error on your component:
```jsx
const UserInvoicesComp = ({ userInvoicesError, ...  })  => {
  return userInvoicesError ? <Message type="error" header="Failed to fetch user invoices" body={userInvoicesError} /> : <ListInvoices />
}
```

alexlyp:
Added requested error check

PR #1323, opened 2019-08-07
[redesign] Add attach files feature to proposals

### LINKED ISSUE #1312: [redesign] Attach files to proposal
Allow for attached files in a proposal when running pgui redesign

thi4go:
This PR adds the hability to add/remove attached files to proposals, and exhibit their thumbnails.

To be faithful to the [design specs](https://xd.adobe.com/spec/25f9f648-8d13-446f-75aa-242b21116f87-53c8/screen/573ae56d-a921-4fd7-a684-71c65bc60089/Create-proposal-file-attached-invalid-mime-type), we need to wait for [this component](https://github.com/tiagoalvesdulce/pi-ui/issues/90) to get implemented. In the meantime, this PR exhibits the error messages similar to how the input box text does.

closes #1312


src/componentsv2/Files/Thumbnail.jsx
```diff
@@ -0,0 +1,57 @@
...
+						<P className={styles.fileErrorText}>
+							{Object.values(err)[0]}
+						</P>
+					)
+				})}
+			</>
+		) : null }
+		<Row className={styles.thumbnailWrapper} justify="left" topMarginSize={errors ? "s" : "m"}>
+			{files.map(f => f.mime.includes("image") ? (
+				<div className={styles.thumbnailWrapper}>
```
fernandoabolafio: I think the `Row` wrapping the thumbnails should not have the same class as the thumbnails. This is semantic confusing and also can cause a lot of issues since changing a style for the children will affect the parent style.

MariaPleshkova:
@fernandoabolafio I thought we can leave the same container for all images and fit images in there by width + show full size img (like a pop up) on a click

fernandoabolafio:
Okay @MariaPleshkova, I agree. Let's go with that and iterate. If we need we improve it later.

thi4go:
Hey, just pushed the commit that addresses the review, @fernandoabolafio. Thanks for the suggestions, code looks much better now. One thing I had to modify in this commit was the placement of the "x" icon on the attached files; it needs to be out of the image thumbnail, otherwise it disappears depending on the color the image has.

I didn't implement the pop up on image click that shows its full size, though, as commented above. Should this be done in this PR also? If so, should this pop up be a modal?

src/componentsv2/Files/Thumbnail.jsx
```diff
@@ -0,0 +1,78 @@
...
+import { classNames, P } from "pi-ui";
+
+
+const ThumbnailGrid = ({ value, onClick, errors = [], viewOnly = false }) => {
+  const files = value.filter(f => f.name !== "index.md");
+  const filesWithErrors = files.map((file, key) => errors.files && errors.files[key] ?
+    { ...file, error: true } : { ...file, error: false }
+  );
+
+	return (
```
fernandoabolafio: needs proper indentation here

src/componentsv2/ProposalForm/validation.js
```diff
@@ -22,5 +31,27 @@ export const proposalValidationSchema = ({
       )
       .matches(...yupFieldMatcher("Name", proposalnamesupportedchars))
       .required("Required"),
-    description: Yup.string().required("Required")
+    description: Yup.string().required("Required"),
+    files: Yup.array().of(
+      Yup.object().shape({
+        mime: Yup.string()
+          .required()
```
fernandoabolafio: remove required

src/componentsv2/ProposalForm/validation.js
```diff
@@ -22,5 +31,27 @@ export const proposalValidationSchema = ({
...
+          .oneOf(validmimetypes, validMimeTypesMessage(validmimetypes)),
+        name: Yup.string().required(),
```
fernandoabolafio: remove required

src/componentsv2/ProposalForm/validation.js
```diff
@@ -22,5 +31,27 @@ export const proposalValidationSchema = ({
...
+        payload: Yup.string().required(),
+        size: Yup.number()
+          .required()
```
fernandoabolafio: remove required

thi4go:
Hey @fernandoabolafio, forgot to test the editing with attachments, so I didn't run by those errors, apologize about that.

Fixed the validations and compile warnings now. Your suggestions were good for validating max files, but it is important to separate between image and text files on max files validation, since that is how the policies on backend are structured. If they change values, will be good to have the gui working without needing big changes. Another thing is, since we are running all validations on Yup, it would be hard to do that dynamics of showing the error the first time, and changing that behaviour on the second time. We also can't validate the length of files array without considering each file's mime type, as explained above.

The way I dealt with this is: when the user tries to upload the sixth image, we let them, but it renders the max count error and disables the upload button. User can only proceed from there if he removes one of the exceeding images. Alternatively, if you think is best and necessary, I can try to do that on Yup of changing to last in first out after rendering error message for the first time, although I'm not sure that's possible to do without too much bloated code.

Regarding the `Uncaught TypeError: Cannot read property 'executeCommand' ...`, this wasn't introduced in this PR, it also happens when the button is clicked on master. Haven't dig into this yet, but I'll try to figure it out for this PR, since they are related. I believe this is the only issue still standing.

thi4go:
Hey @fernandoabolafio, fixed the disabling of buttons, and bottom margin for the error texts. I couldn't reproduce those errors you said mate, topics 2 and 3. The component only lets you attach 6 images. I tested it again here, and the red border is rendering correctly on the images that have error, as per my last commits.

MariaPleshkova:
Hey. Just wanted to review the changes, but idk why can upload any images. I can click on the button, pop up with my computer files opens, I click on smth, but nothing happens.

Maybe that's because of this error?
image
But this in fact restricts me from posting, not from writing smth or ataching images...

thi4go:
Think this is close to a final version, at least for this PR.

@fernandoabolafio changed files count validation to Files component, and left other validations on Yup. This gave us flexibility to handle which files are going in or not. Thanks for the suggestions. Still figuring out the best way to do things with the new code patterns, so I appreciate the reviews :sweat_smile: 

@MariaPleshkova keep in mind that these error renders are temporary, until we have the [Flaoting Badge component](https://github.com/tiagoalvesdulce/pi-ui/pull/104) to match the design specs. Please let me know if there is any improvement to be done on this PR :+1: also, not sure what is causing that issue. Try creating a fresh verified account and see if it does the trick.

I'm attaching a short gif of how errors are being rendered and dealt with currently:
test-attached

fernandoabolafio:
This PR looks good to me in terms of functionalities, I will leave the design compliance for @MariaPleshkova now. The changes implemented by this PR are up for review on pi-redesign and you should be able to access it.

> Hey. Just wanted to review the changes, but idk why can upload any images. I can click on the button, pop up with my computer files opens, I click on smth, but nothing happens.
It wasn't deployed by that time, now should be all good.

MariaPleshkova:
1. I know that this is not for this PR, but just in case:
Proposal credits number is missing
image
image

2.  I think that when there's smth wrong with the file, we don't have to show the file itself, only notification. Because in that case it looks like the file is uploaded
image

3. I'd move the cross icon a little bit up and closer to an image
Like this:
image

Instead of this:
image

4. As far as I remember, we discussed that there should be an ability to open an image as a pop up. There's hover effect on it right now, but it's not clickable

5. I think we have to relocate this notification. It's not like in the view area. I didn't even see it at first. 
image

Guess we can make it like this one:
image
It is closer to uploaded images and strikes the eye more

Actually on my designs notifications looked like this (see below). But tbh I think that this orange text below the frame works better in that case
image

fernandoabolafio:
@thi4go just noticed that the specs include a modal when the user clicks to attach files:
Screen Shot 2019-08-22 at 13 30 20

https://xd.adobe.com/spec/2f0ecbf8-6367-48ea-6b71-1e56191dcd06-593f/screen/5931fea0-31ad-4e2e-99df-2a23b8b60f7e/Create-proposal-attach-file

thi4go:
@fernandoabolafio good catch mate. Implementing it

thi4go:
Just updated the PR. Regarding @MariaPleshkova 's review:

1. Fixed in other PR
2. Fixed
3. Fixed
4. Implemented
5. Fixed

@fernandoabolafio the attach file modal is also done. Also, I stumbled on a bug that was quite hard to figure out; we were using [`react-file-reader`](https://github.com/pgonzalez-santiago/react-file-reader), a fork from the deprecated library. When selecting multiple files, it sometimes changed the indexes of the base64 payload array, which made images get formatted with wrong payload on `getFormattedFiles`. Image below illustrates this:

Screenshot from 2019-08-23 17-15-39

I then installed [`react-file-reader-input`](https://github.com/ngokevin/react-file-reader-input.git), which was the most minimalist implementation I could find with most stars. If there is any other suggestion of lib, or even if our own implementation of a file selector is preferred, let me know!

This PR is ready for another review @MariaPleshkova @fernandoabolafio. Thanks!

src/componentsv2/AttachFileButton/hooks.js
```diff
@@ -0,0 +1,16 @@
+import { useState } from "react";
+
+export function useAttachFileModal() {
```
fernandoabolafio: No need for a custom hook here. Just use the hook `useBooleanState` for controlling modal states. See comments below.

PR #1326, opened 2019-08-08
[wip][redesign] Add admin pages and remaining proposal features

fernandoabolafio:
This PR adds:
- [x] Admin listing of proposals
- [ ] User listing of proposals
- [ ] List pagination for admin and user listing
- [x] Proposal unvetted actions (censor, publish)
- [x] Proposal public actions (abandon, start vote, revoke vote, authorize vote)


lukebp:
1. When you censor a proposal, it removes it from the "Unreviewed" list but it doesn't add it to the "Censored" list until you refresh.

2. When I click on the saved draft proposal title I get a 404.  Clicking on the edit icon works as expected but not when you click the title.

MariaPleshkova:
Well, these are what I found:

1. We can switch to an admin view by clicking on "admin", but when we are already there we still see "admin" in drop down. I think we have to change it to "Public proposals" or just "Proposals" or smth like that and going back to Public proposals page by clicking on it
image

spec:
image


2. It seems to me that the name of a user is closer to the arrow on my designs. Or proposals credits and a plus are farther from it. I mean right now it looks like there are 3 different elements:
image
But it has to be like this: 
image
Hope I was clear:)

3. We have to add a hover effect to the arrow (that is next to the name in header). Let's make it our main blue color on hover #2970FF ( we already have the same pattern for cross icons on pop ups) 

4. Need to move this closer to the left side
image

spec:
image

5. I just also thought, maybe it would be better to add one more pop up to appear after clicking on Confirm button in Confirm/Report as spam pop ups? A pop up with "Confirmed"/"Reported as spam" text to give users some feedback that the action has been done successfully. What do you think?

fernandoabolafio:
@MariaPleshkova thanks for reviewing! I addressed all of your comments, except by the `#4`. I checked and all error messages have left padding of 20px as from the specs. 
About 5, I came up with this to give feedback on proposal actions:
uifeedback

All changes are deployed to pi-redesign.

MariaPleshkova:
@fernandoabolafio this is what I've ended up with (the copy was just created by me, so we can change it, if it's not ok)
image

spec: https://xd.adobe.com/spec/4e7eb732-a445-46b0-4f7e-4152097a10c8-384f/

fernandoabolafio:
@MariaPleshkova updated the actions feedback accordingly to your specs.

MariaPleshkova:
1. when you try to change smth here, the background and the pop up move a little bit (on one of my computers. on another-everything is ok)
image
2. Need to change up/down arrows there. Make these rectangle areas around them invisible and make the color of the arrows themselves #8997A5 

3. Start Abandoned from the capital letter
image
Btw, success modals look cool!

I guess these are the last tweaks from my side.

fernandoabolafio:
@MariaPleshkova thanks for your comments. Answering to each one specifically:

1. I wasn't able to reproduce it both on chrome or Firefox on my computer. Can you write an issue providing more details about how this happened? Like which browser you're using, which machine, etc. Some gifs/video of the issue would be nice too.

2. The number input component is currently being improved on pi-ui and should be updated in this repo soon enough: https://github.com/tiagoalvesdulce/pi-ui/pull/130.

3. Fixed ✅

MariaPleshkova:
@fernandoabolafio @tiagoalvesdulce 
it's chrome on windows, 1920x1080

PR #1349, opened 2019-08-28
Fixes save drafts for proposals and invoices

### LINKED ISSUE #1347: Hyperlink to Draft Proposal is undefined
When you save a proposal as a draft the href attached to the Title links to '/proposals/undefined'

[Screenshot](https://imgur.com/a/a27sWu3)

thi4go:
This is a minor fix for a save draft action crash that was happening on master.

This was intended to fix #1347, but this has been fixed already, it's just not deployed to production yet.

### DIFF
--- a/src/actions/app.js
+++ b/src/actions/app.js
@@ -228,22 +228,25 @@ export const onSaveDraftInvoice = ({
   lineitems,
   files,
   draftId
-}) => {
+}) => dispatch => {
   resetNewInvoiceData();
   const id = draftId || uniqueID("draft");
-  return act.SAVE_DRAFT_INVOICE({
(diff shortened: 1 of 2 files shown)


fernandoabolafio:
@thi4go I'm a bit confused from the PR description. If #1347 is already fixed, what is this PR fixing anyway?

thi4go:
Hey @fernandoabolafio, this line `const { payload } = this.props.onSaveProposalDraft(...args);` was breaking things because `onSaveProposalDraft` returns only the proposal id, so it would save the draft properly, but the redirect would cause an error. Can you test it on master to see if you are getting the same?

PR #1350, opened 2019-08-29
[redesign] Add search user page

### LINKED ISSUE #1341: [redesign] Implement search users page
Implement the Search users page for admins as per design specs

fernandoabolafio:
closes #1341

### DIFF
--- a/src/componentsv2/HeaderNav/HeaderNav.jsx
+++ b/src/componentsv2/HeaderNav/HeaderNav.jsx
@@ -19,7 +19,11 @@ const HeaderNav = ({ history, location }) => {
   function goToPublicProposals() {
     history.push("/");
   }
+  function goToSearchUsers() {
+    history.push("/user/search");
+  }
   const isOnUnvettedRoute = location.pathname === `/proposals/unvetted`;
+  const isOnSearchUsersRoute = location.pathname === `/user/search`;
   return username ? (
     <div className={styles.loggedInContainer}>
       <ProposalCreditsIndicator />
@@ -29,6 +33,11 @@ const HeaderNav = ({ history, location }) => {
         )}
         {isOnUnvettedRoute && (
           <DropdownItem onClick={goToPublicProposals}>Proposals</DropdownItem>
+        )}
+        {user.isadmin && !isOnSearchUsersRoute && (
+          <DropdownItem onClick={goToSearchUsers}>
+            Search for users
+          </DropdownItem>
         )}
         <DropdownItem onClick={goToUserProposals}>My Proposals</DropdownItem>
         <DropdownItem onClick={goToUserAccount}>Account</DropdownItem>
--- a/src/componentsv2/HeaderNav/HeaderNav.jsx
+++ b/src/componentsv2/HeaderNav/HeaderNav.jsx
@@ -27,7 +27,7 @@ const HeaderNav = ({ history, location }) => {
   return username ? (
     <div className={styles.loggedInContainer}>
       <ProposalCreditsIndicator />
-      <Dropdown title={username}>
+      <Dropdown itemsListClassName={styles.dropdownList} title={username}>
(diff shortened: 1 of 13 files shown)


fernandoabolafio:
@MariaPleshkova I had to add a "Search By" field also which wasn't present in the design specs. How does that look to you?
Screen Shot 2019-08-29 at 15 47 41

MariaPleshkova:
@fernandoabolafio I remember that there was such field earlier, but I decided to remove it, because we just dublicated and added extra actions for users. Why do we need it if there's a text inside an input field "user email or username"? I mean, can't the system just filter input info? If there's a @ - it is an email

fernandoabolafio:
@MariaPleshkova this is a technical restriction because of the way the API currently works. We cannot parse the input to infer it is an email or a username because any text search is valid for both cases (like "anyuser" can be used to search an email like "anyuser@email.com"). Also, if you provide an email and a username for the API, it will try to match both. Anyway, I get what you're saying and I like it too but it is just not possible right now. We need to specify for the API what the admin is searching for.

MariaPleshkova:
Ok.I'd do it this way then
image

spec link to the common admin file: https://xd.adobe.com/spec/4e7eb732-a445-46b0-4f7e-4152097a10c8-384f/

fernandoabolafio:
@MariaPleshkova updated: 
Screen Shot 2019-08-30 at 12 54 35

The style for the Radio Buttons will be updated in pi-ui soon enough: https://github.com/tiagoalvesdulce/pi-ui/issues/133.

MariaPleshkova:
@fernandoabolafio increase the distance here, please (30px)
image

fernandoabolafio:
@MariaPleshkova updated:
Screen Shot 2019-08-30 at 14 42 16

src/componentsv2/HeaderNav/HeaderNav.jsx
```diff
@@ -19,17 +19,26 @@ const HeaderNav = ({ history, location }) => {
   function goToPublicProposals() {
     history.push("/");
   }
+  function goToSearchUsers() {
+    history.push("/user/search");
+  }
   const isOnUnvettedRoute = location.pathname === `/proposals/unvetted`;
+  const isOnSearchUsersRoute = location.pathname === `/user/search`;
```
tiagoalvesdulce: There's no need for this to be a template literal. Should be a string literal instead.

src/componentsv2/layout/layouts.module.css
```diff
@@ -15,7 +15,13 @@
     display: flex;
     flex-direction: column;
     justify-content: space-between;
-    max-height: 14rem;
+    /* max-height: 14rem; */
```
tiagoalvesdulce: This comment can be cleaned up

src/containers/User/Search/Search.jsx
```diff
@@ -0,0 +1,134 @@
...
+            }}
+            onSubmit={onSubmit}
+            validationSchema={Yup.object().shape({
+              searchTerm: Yup.string().required("Required")
+            })}
+          >
+            {({
+              values,
+              handleChange,
+              // handleBlur,
```
tiagoalvesdulce: Parameters not used can be erased

src/containers/User/Search/Search.jsx
```diff
@@ -0,0 +1,134 @@
...
+              handleSubmit,
+              isSubmitting,
+              setFieldValue,
+              // errors,
```
tiagoalvesdulce: Same as above

src/containers/User/Search/Search.jsx
```diff
@@ -0,0 +1,134 @@
...
+              isValid
+              // touched
```
tiagoalvesdulce: Same as above

PR #1356, opened 2019-08-29
[redesign] Highlight new comments since last visit

### LINKED ISSUE #1348: [redesign] Highlight of new comments since last visit
Implement the highlight of new comments since the last visit.

fernandoabolafio:
closes #1348 

@MariaPleshkova I needed to add some padding to the comments frames so the gray background looks good on it. With that, the top-level comments are no longer aligned to the title, I didn't think of any other way of doing that. It looks good to me though. Let me know what you think:
Screen Shot 2019-08-29 at 18 10 24

### DIFF
--- a/src/componentsv2/CommentForm/CommentForm.jsx
+++ b/src/componentsv2/CommentForm/CommentForm.jsx
@@ -11,7 +11,8 @@ const CommentForm = ({
   onSubmit,
   onCommentSubmitted,
   disableSubmit,
-  persistKey
+  persistKey,
+  className
 }) => {
   async function handleSubmit(
     values,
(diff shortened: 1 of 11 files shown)


fernandoabolafio:
@MariaPleshkova I adjusted the comments to allow it to grow a bit wider than the container:
Screen Shot 2019-08-30 at 13 28 39

For mobile, I am hiding the  "new" and keeping only the background. I also found better to have the comments container to occupy the full width: 
Screen Shot 2019-08-30 at 13 28 19

MariaPleshkova:
move it closer to the edge
image

MariaPleshkova:
@thi4go hm, yes, it makes sense.
@fernandoabolafio I think we can try to add smth like this:
image

fernandoabolafio:
@thi4go @MariaPleshkova updated! The results is shown in this [video](https://gofile.io/?c=OBMZmp).

thi4go:
Looking great! @fernandoabolafio LGTM

PR #1360, opened 2019-08-30
[redesign] Add censor comment action for admins

### LINKED ISSUE #1343: [redesign] Censor of comments
Implement the censor of comments as per design specs.
Needs design specs from @MariaPleshkova .

thi4go:
closes #1343

### DIFF
--- a/src/containers/Comments/Comment/CommentWrapper.jsx
+++ b/src/containers/Comments/Comment/CommentWrapper.jsx
@@ -1,6 +1,9 @@
 import React, { useState } from "react";
 import styles from "./Comment.module.css";
+import { Text } from "pi-ui";
+import ModalConfirmWithReason from "src/componentsv2/ModalConfirmWithReason";
 import CommentForm from "src/componentsv2/CommentForm";
+import useBooleanState from "src/hooks/utils/useBooleanState";
 import { useComment } from "../hooks";
 import Comment from "./Comment";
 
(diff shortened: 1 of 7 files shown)


src/containers/Comments/Comment/CommentWrapper.jsx
```diff
@@ -104,6 +123,20 @@ const CommentWrapper = ({ comment, children, numOfReplies, ...props }) => {
       {showReplies && (
         <div className={styles.childrenContainer}>{children}</div>
       )}
+      <ModalConfirmWithReason
```
fernandoabolafio: This modal needs to be lifted up to the `Comments` component. With your approach, we will end up with a Modal rendered for each comment since the modals are just "hidden". 
You can add the Modal to the outermost component (eg. Comments) and pass the methods to trigger the modal through the `CommentContext.Provider` so they are accessible from every comment in that list.

fernandoabolafio:
@thi4go one more thing, the  "Censor" button should not appear if the comment is already censored.

thi4go:
Addressed the review comments @fernandoabolafio 👍

MariaPleshkova:
@fernandoabolafio I suggest to do it this way on mobile
image
It's better not to put Censor button close to Reply

fernandoabolafio:
ok @MariaPleshkova. I will create an issue for that since this is merged already.

PR #1364, opened 2019-09-04
[redesign] Add Identity management

### LINKED ISSUE #1340: [redesign] User account: identity tab
Implement the identity tab for the user account as per design specs

tiagoalvesdulce:
Closes #1340

### DIFF
--- a/src/componentsv2/ModalImportIdentity.jsx
+++ b/src/componentsv2/ModalImportIdentity.jsx
@@ -1,6 +1,6 @@
-import { Button, Modal, Text, TextInput } from "pi-ui";
+import { Button, getThemeProperty, Icon, Modal, Text, TextInput, useTheme } from "pi-ui";
 import PropTypes from "prop-types";
-import React, { useState } from "react";
+import React, { useEffect, useState } from "react";
 import ReactFileReader from "react-file-reader";
 import FormWrapper from "src/componentsv2/FormWrapper";
 import { INVALID_FILE, INVALID_KEY_PAIR, LOAD_KEY_FAILED, PUBLIC_KEY_MISMATCH } from "src/constants";
@@ -58,10 +58,13 @@ const onSubmitFiles = (onIdentityImported, userPubkey, loggedInAsEmail, json, se
 const ModalImportIdentity = ({
   show,
   onClose,
-  validationSchema
+  validationSchema,
+  successMessage,
+  successTitle
 }) => {
-  const { onIdentityImported, userPubkey, loggedInAsEmail } = useUserIdentity();
+  const { onIdentityImported, userPubkey, loggedInAsEmail, keyMismatchAction } = useUserIdentity();
   const [fileError, setFileError] = useState(null);
+  const [success, setSuccess] = useState(null);
   const onSubmitNewIdentity = (
     values,
     { resetForm, setSubmitting, setFieldError }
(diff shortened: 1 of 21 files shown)


tiagoalvesdulce:
1. WIll fix, opened de PR to get help with the verificationtoken bit.
2. OK
3. No need to worry about it, was just testing solutions for the verificationtoken problem.
4. Will not fix in this PR, will open another one after this. Will focus on Identity Mngmt.
5. Couldn't find it in the specs.
6. IMO we should hide it with 1, people can see the user id in the URL anyways.
7. It was confusing in the specs, I understood we should show some os the keys and them show all of them in the Modal, but I'm OK with removing it.
8. OK

src/pages/User/VerifyKey.jsx
```diff
@@ -0,0 +1,79 @@
...
+  } = useVerifyKey();
+
+  useEffect(() => {
+    const { verificationtoken } = qs.parse(location.search);
+    if (
+      isEmpty(location.search) ||
+      !verificationtoken ||
+      typeof verificationtoken !== "string"
+    ) {
+      history.push("/user/login");
```
fernandoabolafio: Better throw an error here instead of redirecting to login without further explanation.

tiagoalvesdulce: Yeah, I agree.

tiagoalvesdulce:
@MariaPleshkova can you help me with the fifth item of Fernando's list? Missing text copy for Past Public Keys and User ID. Thanks!

MariaPleshkova:
@tiagoalvesdulce yeah, there's no text copy in the specs. I think we need to request it

tiagoalvesdulce:
OK, everything but the copies is done.

src/pages/User/VerifyKey.jsx
```diff
@@ -0,0 +1,69 @@
...
+  useEffect(() => {
+    if (loggedInAsEmail && !keyUpdated) {
+      const { verificationtoken } = qs.parse(location.search);
+      onVerifyUserKey(loggedInAsEmail, verificationtoken);
+    }
+  }, [loggedInAsEmail, location, onVerifyUserKey, keyUpdated]);
+
+  useEffect(() => {
+    if (
+      verifyUserKey &&
```
fernandoabolafio: Can we extract this big if statement to a more meaningful constant? Also, it is very confusing that we have two effects in this component and one is calling `onVerifyUserKey` and the other one is calling  `verifyUserKey`. Perhaps we can at least give better names to the functions to become more clear what they are doing. 
If I am not missing anything here you could also simplify whats happening here by calling the `onVerifyUserKey` using an async function and if it succeeds you can immediately verify the new key and update the local storage.
Something like:

```jsx
const VerifyKey = ({}) => {
   ...
   const verifyAndUpdateLocalKey = useCallback(async () => {
       setLoading(true);
       try {
         await onVerify(...)
         verifyUserPubkey(...)
         pki.myPubKeyHex(...)
         setLoading(false)
       } catch(e) {
          setLoading(false)
          throw e;
       }
    },[...])

    useEffect(() => {
       if(...) {
        verifyAndUpdateLocalKey(...)
      }
    },[...])


   return (...);
}
```

This removes the need for two effects on this small component and makes the code more readable. Also, notice that I am adding a flag for loading which is currently lacking.

src/componentsv2/ModalImportIdentity.jsx
```diff
@@ -0,0 +1,161 @@
...
+      onSubmitFiles(onIdentityImported, userPubkey, loggedInAsEmail, values, setFileError);
+      resetForm();
+      setSubmitting(false);
+    } catch (e) {
+      setSubmitting(false);
+      setFieldError("global", e);
+    }
+  };
+  return (
+    <Modal title="Import Identity" show={show} onClose={onClose}>
```
fernandoabolafio: Please, set a fixed width for the Modal so it doesn't expand when an error is presented.

src/pages/User/VerifyKey.jsx
```diff
@@ -0,0 +1,69 @@
...
+    verifyUserKey,
+    updateMe,
+    verifyUserKeyError,
+    onVerifyUserKey
+  } = useVerifyKey();
+
+  const [keyUpdated, setKeyUpdated] = useState(false);
+
+  useEffect(() => {
+    if (loggedInAsEmail && !keyUpdated) {
```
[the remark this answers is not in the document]
tiagoalvesdulce: Really? They are not related. `keyUpdated` is just a check to not fetch an already updated key and the loggedInAsEmail check is to not pass an `undefined` value to `onVerifyUserKey`. I can not think of a good name for this constant, to be honest. Although, I'm free to listen to suggestions if you have one.

src/componentsv2/ModalImportIdentity.jsx
tiagoalvesdulce: Are you sure? Setting a fixed width for the Modal doesn't sound very responsive.

tiagoalvesdulce: I tested and responsiveness works just fine

src/pages/User/VerifyKey.jsx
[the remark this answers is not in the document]
fernandoabolafio: Ok, let's keep it like that.

MariaPleshkova:
@fernandoabolafio @tiagoalvesdulce added this as a new issue #1377

PR #1374, opened 2019-09-10
[redesign] Use batch vote summary endpoint

### LINKED ISSUE #1357: [redesign] Use batch vote summary route
The `AllVoteStatus` and `VoteStatus` routes are being deprecated and replaced by the `BatchVoteSummary` route (decred/politeia#967).

The gui will need to replace any `AllVoteStatus` and `VoteStatus` calls with a `BatchVoteSummary` call.

amassarwi:
- closes #1357


src/actions/api.js
```diff
@@ -1191,6 +1186,29 @@ export const onFetchProposalsVoteStatusByTokens = tokens => async dispatch => {
   }
 };
 
+export const onFetchBatchProposalsVoteStatus = (csrf, tokens) => dispatch => {
```
fernandoabolafio: Rename the action and don't pass the csrf as a parameter, use `withCsrf` internally as other functions currently do.
```suggestion
export const onFetchBatchProposalsVoteSummary = (tokens) => dispatch => {
```

src/actions/api.js
```diff
@@ -354,18 +353,14 @@ export const onFetchProposalsBatch = (tokens, fetchVoteStatus = true) =>
     dispatch(act.REQUEST_PROPOSALS_BATCH(tokens));
     try {
       let promises = [api.proposalsBatch(csrf, tokens)];
-
       if (fetchVoteStatus) {
```
fernandoabolafio: We don't need an array of promises here anymore since we're always dispatching two promises. Instead, do something like:
```js
...
const proposals = await api.proposalsBatch(csrf, tokens);
if (fetchVoteStatus) {
   await dispatch(onFetchBatchProposalsVoteStatus(tokens));
}
...
```

src/containers/Proposal/Detail/hooks.js
```diff
@@ -21,13 +21,15 @@ const mapStateToProps = {
...
-  loading: or(sel.proposalIsRequesting, sel.isApiRequestingPropVoteStatus)
+  loading: or(
+    sel.isApiRequestingProposalsVoteSummary,
+    sel.isApiRequestingProposalsBatch
+  )
 };
 
 const mapDispatchToProps = {
   onFetchUser: act.onFetchUser,
-  onFetchProposal: act.onFetchProposal,
```
fernandoabolafio: In this hook, keep `onFetchProposal` to bring the proposal data. Only replace the vote status call with the vote summary one.
Note that the batch call for proposals doesn't bring the entire proposal data (the files aren't included to make it lighter) so in order to fetch the full proposal you need to use onFetchProposal.

src/containers/Proposal/Detail/hooks.js
```diff
@@ -39,22 +41,22 @@ const isEqualProposalToken = (proposal, token) => {
 };
 
 const proposalWithFilesOrNothing = proposal => {
-  return proposal && proposal.files && !!proposal.files.length
-    ? proposal
-    : null;
+  // return proposal && proposal.files && !!proposal.files.length
```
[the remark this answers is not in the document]
fernandoabolafio: See comment above

src/reducers/api/api.js
```diff
@@ -273,7 +273,11 @@ const api = (state = DEFAULT_STATE, action) =>
     [act.REQUEST_PROPOSALS_VOTE_STATUS]: () =>
       request("proposalsVoteStatus", state, action),
     [act.RECEIVE_PROPOSALS_VOTE_STATUS]: () =>
-      onReceiveProposalsVoteStatus(state, action),
+      onReceiveProposalsVoteSummary(state, action),
```
amassarwi: Should we revert this change and keep the older handler `onReceiveProposalsVoteStatus` for `RECEIVE_PROPOSALS_VOTE_STATUS` action

amassarwi:
Question:
https://github.com/decred/politeiagui/blob/master/src/actions/api.js#L1160 

We use `onFetchUserProposalsWithVoteStatus` action in `src/containers/Proposal/User/hooks.js`, which calls `onFetchProposalsVoteStatusByTokens` [here](https://github.com/decred/politeiagui/blob/master/src/actions/api.js#L1172), shouldn't we use the vote summary new action ?

amassarwi:
> Question:
> https://github.com/decred/politeiagui/blob/master/src/actions/api.js#L1160
> 
> We use `onFetchUserProposalsWithVoteStatus` action in `src/containers/Proposal/User/hooks.js`, which calls `onFetchProposalsVoteStatusByTokens` [here](https://github.com/decred/politeiagui/blob/master/src/actions/api.js#L1172), shouldn't we use the vote summary new action ?

Double checked the user's proposals page and it didn't show up because of the vote status requests, updated it to use the new vote summary route

fernandoabolafio:
Yes @amassarwi, you're right, we need to use the vote summary batched call on wherever we were calling the vote status before.

src/reducers/api/api.js
fernandoabolafio: Yeah, let's keep the old handler for this action for now. So, do use the new handler for RECEIVE_PROPOSALS_VOTE_SUMMARY and keep the old one for the vote status. We will remove it altogether once we release the v2 into production.

amassarwi:
> @amassarwi just revert the change you pointed out for the old reducer handler and this is good to go. Nice job!

Done

PR #1395, opened 2019-09-11
Close Login modal when going to Privacy Policy screen

### LINKED ISSUE #1380: Close Login modal when going to Privacy Policy screen
This pop up should close when going to another screen
pi_login

victorgcramos:
Closes #1380


victorgcramos:
Following @fernandoabolafio's suggestion, I created a new hook for route changes in order to apply the "Privacy Policy" (#1380) fix for all route redirecting.

src/componentsv2/Router/RouterProvider.jsx
```diff
@@ -12,6 +12,27 @@ const routerCtx = createContext();
...
 
+/**
+ * Watch for route changes and triggers the callbackFn when
+ * route changes.
+ *
+ * @param {Boolean} active - if hook is active or not
+ * @param {Function} callbackFn - function that will be
+ * triggered when route changes
+ */
+export const useRouteChanges = (active, callbackFn) => {
```
fernandoabolafio: Please, let's move this hook to its own file under `src/hooks/utils`. And what do you think about renaming it to `useOnRouteChange`?

src/componentsv2/ModalLogin.jsx
```diff
@@ -1,8 +1,10 @@
 import React from "react";
 import { Modal } from "pi-ui";
 import LoginForm from "src/containers/User/Login/Form";
+import { useRouteChanges } from "src/componentsv2/Router";
 
 const ModalLogin = ({ title = "Login", onLoggedIn, onClose, ...props }) => {
+  useRouteChanges(props.show, onClose);
```
fernandoabolafio: Why do you need to pass an "active" arg to this function? Couldn't it be something like:

```jsx
const ModalLogin = ({...}) => {
   useOnRouteChange(onClose);
   ...
}
```

I tested the approach above with the following hook and it worked  just fine:

```js
import { useEffect } from "react";
import { useRouter } from "src/componentsv2/Router";

export default function(callbackFn) {
  const { location } = useRouter();
  useEffect(() => {
    callbackFn();
  }, [location.pathname, callbackFn]);
}
```

src/componentsv2/Router/RouterProvider.jsx
victorgcramos: Good idea! Working on this!

PR #1400, opened 2019-09-12
[redesing] Identity page explanatory texts

### LINKED ISSUE #1373: [redesign] Include text copy for identity tab
The identity tab is missing the text copy under "Past public keys"  and "user id". 

Screen Shot 2019-09-10 at 12 32 59

thi4go:
closes #1373

### DIFF
--- a/src/containers/User/Detail/Identity/Identity.jsx
+++ b/src/containers/User/Detail/Identity/Identity.jsx
@@ -108,6 +108,10 @@ const Identity = ({ history, loadingKey, pubkey, id: userID, identities }) => {
                   "margin-top-s"
                 )}
               >
+                <P className="margin-bottom-s">
+                The active private key is used to sign proposals and comments. 
+                The public key is used to verify the identity.
+                </P>
                 <Text backgroundColor="blueLighter" monospace>
                   {pubkey || userPubkey}
                 </Text>
@@ -136,6 +140,9 @@ const Identity = ({ history, loadingKey, pubkey, id: userID, identities }) => {
       >
         Past public keys
       </Text>
+      <P>
+        List of inactive public keys your account has had in the past.
+      </P>
       <Button size="sm" onClick={openShowAllModal}>
         Show all
       </Button>
@@ -150,6 +157,10 @@ const Identity = ({ history, loadingKey, pubkey, id: userID, identities }) => {
       >
         User ID
       </Text>
+      <P className="margin-bottom-s">
+        Unique 16-byte UUID, as defined in <a href="http://tools.ietf.org/html/rfc4122">
+        RFC 4122</a>, used to identify your user.
+      </P>
       <Text backgroundColor="blueLighter" monospace>
         {userID}
       </Text>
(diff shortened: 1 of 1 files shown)


src/containers/User/Detail/Identity/Identity.jsx
```diff
@@ -108,6 +108,10 @@ const Identity = ({ history, loadingKey, pubkey, id: userID, identities }) => {
                   "margin-top-s"
                 )}
               >
+                <P>
+                The active private key is used to sign proposals and comments. 
+                The public key is used to verify the identity.
+                </P> <br />
```
tiagoalvesdulce: This `<br>` tag isn't needed, try using css properties instead.

src/containers/User/Detail/Identity/Identity.jsx
```diff
@@ -150,6 +157,10 @@ const Identity = ({ history, loadingKey, pubkey, id: userID, identities }) => {
       >
         User ID
       </Text>
+      <P>
+        Unique 16-byte UUID, as defined in <a href="http://tools.ietf.org/html/rfc4122">
+        RFC 4122</a>, used to identify your user.
+      </P> <br />
```
tiagoalvesdulce: Same as above.

thi4go: Fixed mate

src/containers/User/Detail/Identity/Identity.jsx
```diff
@@ -136,6 +140,9 @@ const Identity = ({ history, loadingKey, pubkey, id: userID, identities }) => {
       >
         Past public keys
       </Text>
+      <P>
+        List of inactive public keys your account has had in the past.
```
fernandoabolafio: The last active pub key is also included in that list, so either we filter it out or we properly update this message. I like the former best.

src/containers/User/Detail/Identity/Identity.jsx
```diff
@@ -108,6 +108,10 @@ const Identity = ({ history, loadingKey, pubkey, id: userID, identities }) => {
                   "margin-top-s"
                 )}
               >
+                <P className="margin-bottom-s">
+                The active private key is used to sign proposals and comments. 
+                The public key is used to verify the identity.
```
fernandoabolafio: I think we can expand this a bit.  I thought about something like:

Your public and private keys constitute your identity. The private key is used to sign proposals, comments and any other relevant action taken by you on Politeia. You can have only one identity active at a time. This is the public key stored in your browser:

@RichardRed0x can you weight on this, please?

@thi4go please make sure that this message won't conflict with the message displayed when there is a key mismatch between server and client.

RichardRed0x: Looks good. Some tweaks/expansion:

Your public and private keys constitute your identity. The private key is used to sign your proposals, comments and any up/down votes on Politeia. You can have only one identity active at a time. Your keys are stored in your browser by default, so if you use Politeia on multiple machines you will need to import your keys before you can participate. This is the public key stored in your browser:

thi4go:
Addressed the comments mates. Ready for another review @fernandoabolafio @RichardRed0x

PR #1409, opened 2019-09-16
Add cancel button to the ProposalForm component.

### LINKED ISSUE #1391: [redesign] Add Cancel button to the Edit proposal page
If I choose Edit proposal, I should have an ability to cancel editing, not only to submit it. So we have to add one more button in there.
I suggest smth like this:
image

camus-code:
Closes #1391 

This commit adds a cancel button to the ProposalForm component. This button redirects the user to the last visited page on Politeia. If none exists, this button redirects the user to the Politeia homepage.

### DIFF
--- a/src/componentsv2/ProposalForm/ProposalForm.jsx
+++ b/src/componentsv2/ProposalForm/ProposalForm.jsx
@@ -70,7 +70,7 @@ const ProposalForm = React.memo(function ProposalForm({
   		setFieldValue("name", values.name);
       setFieldValue("description", values.description);
     },
-    [setFieldValue]
+    [setFieldValue, values.name, values.description]
   );
 
   return (
--- a/src/componentsv2/ProposalForm/ProposalForm.jsx
+++ b/src/componentsv2/ProposalForm/ProposalForm.jsx
@@ -12,7 +12,6 @@ import ModalMDGuide from "src/componentsv2/ModalMDGuide";
 import DraftSaver from "./DraftSaver";
 import { useProposalForm, useFullImageModal } from "./hooks";
 import useBooleanState from "src/hooks/utils/useBooleanState";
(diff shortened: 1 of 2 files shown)


src/componentsv2/ProposalForm/ProposalForm.jsx
```diff
@@ -65,6 +66,16 @@ const ProposalForm = React.memo(function ProposalForm({
 
   const textAreaProps = useMemo(() => ({ tabIndex: 2 }), []);
 
+  const { pastLocations, history } = useRouter();
+  const previousLocation = pastLocations[1];
+  const returnToPreviousLocation = useCallback(() => history.goBack(), [
+    history
+  ]);
+  const returnHome = useCallback(() => history.push("/"), [
```
fernandoabolafio: Instead of returning home, go to the proposal detail route

src/componentsv2/ProposalForm/ProposalForm.jsx
```diff
@@ -65,6 +66,15 @@ const ProposalForm = React.memo(function ProposalForm({
 
   const textAreaProps = useMemo(() => ({ tabIndex: 2 }), []);
 
+  const { pastLocations, history } = useRouter();
```
fernandoabolafio: You're not using the pastLocations here. Tbh, let's simplify it, just go to the proposal detail always on cancel.

tiagoalvesdulce:
Shouldn't we just cancel the edits made and show the proposal state before them?

tiagoalvesdulce:
Adding to my comment above, if editing a proposal I think the Ideal is to cancel the edits made to the proposal and not redirect the user anywhere else. If creating a new proposal, the cancel button should redirect to the previous location.

PR #1433, opened 2019-09-20
[redesign] Do not collapse comments by default

### LINKED ISSUE #1426: [redesign] Do not collapse comments by default
Make all comments expanded. We may want to change that later. 
For example, if a comment has too many negative votes it could be collapsed, as Reddit does.

thi4go:
Should we keep the `isThreadParent` / `topLevelComment` props for some reason?

closes #1426


tiagoalvesdulce:
No, they are no longer necessary. You can go ahead and remove them.

PR #1437, opened 2019-09-23
[redesign] Update page title when accessing a proposal details page

### LINKED ISSUE #1416: [redesign] Change tab title when acessing a proposal
When accessing a proposal, we need to change the tab title to the proposal title.

amassarwi:
- closes #1416

### DIFF
--- a/src/containers/Proposal/Detail/hooks.js
+++ b/src/containers/Proposal/Detail/hooks.js
@@ -49,11 +49,12 @@ const getCurrentHTMLTitle = () => document.title;
 const setHTMLTitle = title => (document.title = title);
 
 export function usePageTitle(title) {
+  title = title || "";
   const [originalTitle, setOriginalTitle] = useState();
   useEffect(() => {
     const originalTitle = getCurrentHTMLTitle();
     setOriginalTitle(originalTitle);
-    setHTMLTitle(title || "");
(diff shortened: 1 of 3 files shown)


src/containers/Proposal/Detail/hooks.js
```diff
@@ -44,6 +44,21 @@ const proposalWithFilesOrNothing = proposal => {
     : null;
 };
 
+const getCurrentHTMLTitle = () => document.title;
+
+const setHTMLTitle = title => (document.title = title);
+
+export function usePageTitle(title) {
```
fernandoabolafio: This needs to be moved under its own file (`src/hooks/utils/useDocumentTitle`). Another thing is that this hook is triggering an infinite update:
 - Your effect reacts, to originalTitle changes
- The effect mutates the state for `originalTitle` by applying the current title
- Your effect reacts again to the `originalTitle` changes... see where I am going here?
ps:Just put a log inside the effect and you will see it.

I thought about it for a while and I think we will need to use a different approach. Instead of storing the original title, we may just use the default title from the app config (`const { title } = useConfig()`). And apply this value once the component unmounts. What do you think?

amassarwi: Make sense, adding it

PR #1442, opened 2019-09-23
[wip][redesign] Hide download of comments on single thread view

### LINKED ISSUE #1424: [redesign] Download Comments Bundle link disappears
In the single comment thread page, the Download Comments Bundle link is being shown for a couple of seconds and then it disappears. I think we shouldn't show it at all.

bug2

fernandoabolafio:
closes #1424

### DIFF
--- a/src/componentsv2/Proposal/Proposal.jsx
+++ b/src/componentsv2/Proposal/Proposal.jsx
@@ -229,7 +229,7 @@ const Proposal = ({ proposal, extended, collapseBodyContent }) => {
                     className="margin-right-s"
                     label="Download Proposal Bundle"
                   />
-                  {isPublic && !!numcomments && (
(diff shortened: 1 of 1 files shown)


thi4go:
Hey mate, got a weird behavior on the download link when going back from single comment thread page to proposal page. See below:

Peek 2019-09-23 17-55

tiagoalvesdulce:
It's working fine here.

fernandoabolafio:
I see the issue on my end @thi4go, nice catch! It's a bit tricky because we reset and fetch comments when switching views mode ( single thread or not). I will se if I can come up with a better cache solution for this situation.

fernandoabolafio:
Will be fixed on https://github.com/decred/politeiagui/pull/1448

PR #1446, opened 2019-09-23
build: replace travis-ci with ci via github actions.

dajohi:


dajohi:
@fernandoabolafio it will run properly once in master.

PR #1448, opened 2019-09-24
[redesign] Cache comment data in its own state branch

### LINKED ISSUE #1423: [redesign] Comments voting not working on the comment single thread page
When I go to the comment single thread page by clicking the date, I cannot up/downvote the comments of that thread. Works fine if I refresh the page though.

bug

### LINKED ISSUE #1424: [redesign] Download Comments Bundle link disappears

fernandoabolafio:
This PR moves the handling of comments data to its own reducer as explained here: https://github.com/decred/politeiagui/pull/1448#issuecomment-535465091.

closes #1423 
closes #1424


tiagoalvesdulce:
The behavior is pretty inconsistent. Sometimes I can vote and sometimes not. Also, it's looking really weird when I can place a vote. Take a look below:

Can't vote:

bugg2


Can vote but it's messed up (the upvote is downgrading the score and the downvote is sometimes subtracting 2 from the score):

bugg

fernandoabolafio:
@tiagoalvesdulce is there any specific steps you followed to not be able to vote? I am not able to reproduce it here.
Regarding the "vote messed up". This is the correct behavior. The comment you were testing had "-1" firsly, so:
- you vote up -> (1 + -1) result is 0
- you vote up again -> (0 + -1) result is -1 
- you vote down -> (-1 + -1) result is -2

So if you vote up and then down, the result would jump from 0 to -2, as expected.

tiagoalvesdulce:
Oh, OK. We just need to fix the "can't vote" bug then.

fernandoabolafio:
@tiagoalvesdulce @thi4go @victorgcramos tagging you here because I think this a important decision to be taken. 
This PR issue is related to what's causing #1442. 
We are always fetching and resetting the proposal comments when loading a Proposal detail page. This happens because our redux structure is very unflexible and is driven by the API calls instead of being driven by the real needs of the application. 

**This is how we currently apply the fetching of comments:**
- Load proposal page -> fetch comments -> move away from proposal page -> reset proposal comments

**How we should be dealing with this:**
- Load proposal page -> fetch comments and store comments by the proposal token -> move away from proposal page -> comments are still stored in the app state

This way we avoid the override of state in the store. So, for instance, if the user, just moved away from a proposal page and return to that, we just pull the data from the store instead of re-fetching it.

The same kind of situation is true for other data fetched while in the proposal route page, such as comments likes. 

Other consequences of this state structure driven by API calls is that we may hurt the single-source-of-truth. For example, when we access a proposal detail page, we fetch the proposal data and keep it in its own state branch instead of merging it with the other proposals fetched for the public list or unvetted list.

**Solution:** 

What I have in mind is to introduce a normalized data state using another reducer and progressively move our data to this normalized state. The outline of the design would be:

-  The "api" state holds only the control values such as:
    - isRequesting: either this API call is ongoing or not
    - payload: params used to create this API call (e.g: "token", "userid") 

- The "model" (normalized) state holds the normalized data processed from the api call result.
     - For example `commentsByToken`, `voteSummaryByToken`, `proposals`, etc.

**Example for the fetching of comments:**

| action                    | action payload                | api state                                | model state                                                    |
|---------------------------|-------------------------------|------------------------------------------|----------------------------------------------------------------|
| REQUEST_PROPOSAL_COMMENTS | { token }                     | {isRequesting: true, params: { token } } | nothing changes                                                |
| RECEIVE_PROPOSAL_COMMENTS | {comments, accesstime, token} | { isRequesting: false}                   | {commentsByToken: { "xtoken": { comments, accessTime, token}}} |
| RECEIVE_PROPOSAL_COMMENTS | { error }                     | { isRequesting: false, error}            | nothing changes                                                |

Initially, we won't remove the data stored in the api state as it can cause compatibilities issues with the old version and for CMS so I suggest we leave it as is and once we get the "model" state very solid we start the cleanup process.

My last commit has the skeleton of this structure for the fetching of comments. Please, take a look and let me know what do you think.

tiagoalvesdulce:
Hey Fernando, I like the approach you suggested. Your last commit is looking good to me.

fernandoabolafio:
ok @tiagoalvesdulce this PR is ready for a final review.

thi4go:
Hey fernando, also liked the normalized state model approach you did in this PR. Will be nice to gradually transfer our state to this, and decouple from our old redux api model.

victorgcramos:
Great approach @fernandoabolafio! And I agree with @thi4go. I really like the cache for comments, since we need to fetch them every time a user access a proposal. Good job!

tiagoalvesdulce:
LGTM. I have just one question: should the Download Comments Bundle link be shown in the single thread comment page?

fernandoabolafio:
It looks fine for me @tiagoalvesdulce. We need to fetch all comments in the single thread view anyway

PR #1449, opened 2019-09-24
ci: only run on pull requests

dajohi:


fernandoabolafio:
You may want to address this as well: `yaml: line 19: mapping values are not allowed in this context`
This error happened in the last master push.

PR #1463, opened 2019-10-02
[redesign] Fix loading errors from account page

### LINKED ISSUE #1461: [redesign] Credits tab keeps showing the spinner indefinitely
When I go to the "credits" tab under a user account it keeps showing the spinner indefinitely.

thi4go:
This PR fixes some issues on the account page regarding loading data with the spinner, where it would render indefinitely. It also fixes the flickering on credits tab, when switched from the identity tab.

closes #1461

### DIFF
--- a/src/containers/User/Detail/Credits/Credits.jsx
+++ b/src/containers/User/Detail/Credits/Credits.jsx
@@ -19,6 +19,7 @@ const Credits = () => {
     user,
     userMe,
     isApiRequestingUserProposalCredits,
+    shouldFetchProposalCredits,
     proposalCreditPurchases,
     loggedInAsUserId,
     proposalPaywallAddress,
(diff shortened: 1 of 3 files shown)


fernandoabolafio:
Is it still relevant if we have https://github.com/decred/politeiagui/pull/1465/?
I am saying that because the loading spinner is showing based on the length of the proposal credits array which will always be a falsy value when the user has no prop credits (because the default value is an empty array). However, with #1465, if we store the prop credits per user id, we can rely on the fact that there are no prop credits data at all for a user by checking under the user key (ex: `proposalCredits.byId[userid]`).  Anyway, I rather have this fixed from the root on #1465. Can we do that?

thi4go:
We can mate, I'm finishing #1465 up and will update the status on this PR. 👍

thi4go:
This will be fixed in #1465

PR #1465, opened 2019-10-02
[redesign] Cache user proposal credits data in its own state branch

### LINKED ISSUE #1457: [redesign] Proposal credits found by a rescan need a page refresh to get added
Had an account that was missing two proposal credits; when I hit rescan, it found those two credits, but it wasn't added to my header field, only when I ran a page refresh. It needs to update the state accordingly.

thi4go:
This PR normalizes user proposal credits data in its own state branch, as we are doing per https://github.com/decred/politeiagui/pull/1448.

Named this state branch as `user` with the intention of putting the logged in user data inside it gradually. If its deemed best, we can change this nomenclature for something like `me`, or even, `users`.`me`, where we'd put data from other users on the same branch `users`. Please let me know what yall think @fernandoabolafio @tiagoalvesdulce @victorgcramos 

closes #1457 and #1461


src/selectors/models/user.js
```diff
@@ -0,0 +1,27 @@
+import get from "lodash/fp/get";
+
+export const proposalCreditsV2 = get(["user", "proposalCredits", "credits"]);
+
+export const proposalCreditsPurchasesV2 = state => {
```
fernandoabolafio: It's very hard to follow what is happening in this selector. We need a better naming and maybe break it down a bit or add some inline comments in the key parts.
Also, let's use the methods from `reselect` to create memoized selectors. If you follow my suggestions of storing the proposal credits per user ID. This selector would turn into something like this:
```js
const makeGetUserCreditsGroupedByTx = (userid) => { 
 const spentUserCredits =  makeGetSpentUserCredits(userid);
 const unspentUserCredits = makeGetUnspentUserCredits(userid);
 return createSelector(
    spentUserCredits,
    unspentUserCredits,
    (spentCredits, unspentCredits) => [...spentCredits, ...unspentCredits].reduce((result, credit) => { ... }, []), // returns the grouped credits
    groupedCredits => groupedCredits.sort(...) // sort the grouped credits
 )
}

```


The reducer code needs some work of course and other details you'll have to figure out but I think from that example you can grasp the idea.

thi4go: Thanks for the suggestions mate, now the selectors are memoized and composable, and much more clear on what is happening.

src/reducers/models/user.js
```diff
@@ -0,0 +1,40 @@
...
+
+const DEFAULT_STATE = {
+  proposalCredits: {
+    purchases: {
+      unspentcredits: null,
+      spentcredits: null
+    },
+    credits: null
+  }
+};
```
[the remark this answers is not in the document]
thi4go: Fixed the general structure of the reducer mate, will comment on the PR page on how it is right now 👍

thi4go:
Hey mate this is ready for another review. I opted for the following structure on our credit reducer, please let me know if its satisfactory:

image

Did this way so we wouldn't get duplicated user ids for spent and unspent credits.

thi4go:
Hey mate, updated the code with bug fixes and general improvements. Also tested every path the user can make regarding these changes. Think this is good to go now 👍

PR #1468, opened 2019-10-03
Improve layout for large screens and remove the sidebar from some pages

### LINKED ISSUE #1425: [redesign] Layout issues
As pointed by @behindtext we have room for some improvements in the overall layout of the website:

> what i think definitely needs improvement - vertical size of header is too large, it pushes content down and just eats up space
the div that contains "Proposal Details" in particular - it just wastes space
the spacing between that proposal details div and the content div below it also eats up space
the left and right margins eat up space that should be used for content
the right div that has the About Politeia info really needs to go away - it just wastes space
that div has some utility and doesn't look so out of place on the landing page that lists proposals
on the proposal details page, it is a total waste, imo
i think the font size could come down a bit without causing legibility problems

Another thing pointed out by @davecgh is:
>  The other thing that stands out to me is the lack of a strong visual indicator between the entries to clearly delineate them (e.g. a drop shadow like the old one has is enough).

Another important thing is that people are claiming that they would rather use the old website than the new one for the desktop. One of the main reasons is the centralization of content. Looks like that having the content always aligned to left margin is a common preference.

@MariaPleshkova @linnutee

tiagoalvesdulce:
1. Remove the "About Politeia" sidebar from the single proposal page (details, edit, new, etc) and from the Search Users page.
2. Remove the title from the Proposal Details page
3. Fix left and right margins for screens larger than 1400px and stretches the content to fill the window

Closes #1425


tiagoalvesdulce:
@MariaPleshkova I think the "About Politeia" sidebar is out of context in the Account page. Can we get rid of this?

fernandoabolafio:
Credits tab doesn't look good as well  in the stretched layout:
Screen Shot 2019-10-05 at 10 54 55

PR #1475, opened 2019-10-07
[redesign] Rewording some proposal status

### LINKED ISSUE #1413: [redesign] "Waiting for approval" proposal status is not clear
image

I assume it means that the proposal owner has authorized voting to start? "Waiting for approval" reads more like waiting for admin approval to be displayed, to me.

The others are labelled "Hasn't authorized yet", which is also not that clear.

How about: "Waiting for owner authorization for voting" and "Waiting for admin authorization for voting"? They're a bit longer but it's hard to explain this fairly unusual status in a few words.

thi4go:
Hey guys @RichardRed0x @MariaPleshkova , wanted to get your final word on this one. So this PR changes the following status:

`Hasn't authorized yet` -> `Waiting for author authorization`
`Waiting for approval` -> `Waiting for admin authorization`

We also have `Active` for ongoing proposal voting, `Finished` for finished proposal voting, and `Abandoned`.

Is there any status you think needs some change? imo I like how succinct they are after a proposal is active, and each one seems clear to me after the changes you proposed, but of course, my opinion might be biased.

closes #1413

### DIFF
--- a/src/componentsv2/Proposal/helpers.js
+++ b/src/componentsv2/Proposal/helpers.js
@@ -30,9 +30,15 @@ export const getProposalStatusTagProps = (proposal, voteSummary) => {
   if (isPublicProposal(proposal) && !!voteSummary) {
     switch (voteSummary.status) {
       case PROPOSAL_VOTING_NOT_AUTHORIZED:
-        return { type: "blackTime", text: "Hasn't authorized yet" };
+        return {
+          type: "blackTime",
+          text: "Waiting for author to authorize voting"
+        };
       case PROPOSAL_VOTING_AUTHORIZED:
-        return { type: "yellowTime", text: "Waiting for approval" };
+        return {
+          type: "yellowTime",
+          text: "Waiting for admin to start voting"
+        };
       case PROPOSAL_VOTING_ACTIVE:
         return { type: "bluePending", text: "Active" };
       case PROPOSAL_VOTING_FINISHED:


RichardRed0x:
> Waiting for author authorization

That one reads a bit odd because of author author, slightly better would be be "Waiting for authorization from author" or "Waiting for author to authorize voting" - extra word but it could be a good one to add, because without "voting" people who are not familiar with Politeia will not understand what the authorization is for.

MariaPleshkova:
Agree with @RichardRed0x .

thi4go:
Good points @RichardRed0x, pushing in the changes.

src/componentsv2/Proposal/helpers.js
```diff
@@ -30,9 +30,12 @@ export const getProposalStatusTagProps = (proposal, voteSummary) => {
...
     switch (voteSummary.status) {
       case PROPOSAL_VOTING_NOT_AUTHORIZED:
-        return { type: "blackTime", text: "Hasn't authorized yet" };
+        return {
+          type: "blackTime",
+          text: "Waiting for author to authorize voting"
+        };
       case PROPOSAL_VOTING_AUTHORIZED:
-        return { type: "yellowTime", text: "Waiting for approval" };
+        return { type: "yellowTime", text: "Waiting for admin authorization" };
```
fernandoabolafio: Shouldn't it be "Waiting for admin to start voting" for the same reason pointed by @RichardRed0x?

thi4go: True mate, I agree this one should change also, and explicit voting on it. :+1:

PR #1476, opened 2019-10-08
[redesign] Loading indicator when fetching proposal tokens

### LINKED ISSUE #1308: [redesign] Display a loading indicator before starting to fetch the proposals
We need a loading indicator while the proposals lazy loading hasn't started yet (e.g while fetching the token inventory).

This issue is a continuation of a discussion in #1296. I will quote@MariaPleshkova suggestion here:

> @lukebp I think it can be solved by adding this effect of blinking gray cards once the page starts showing any content on it. I mean we can just add a few (eg 3 or 4) of these gray rectangles even before the system counts the numper of proposals. What do you think @fernandoabolafio ?

_Originally posted by @MariaPleshkova in https://github.com/decred/politeiagui/pull/1296#issuecomment-512731934_

thi4go:
Decided to go with the spinner to indicate loading status to be consistent with other loading areas we have across the system.

closes #1308

### DIFF
--- a/src/containers/Proposal/Public/Public.jsx
+++ b/src/containers/Proposal/Public/Public.jsx
@@ -1,4 +1,5 @@
 import React, { useCallback, useMemo } from "react";
+import { Spinner } from "pi-ui";
 import styles from "./PublicProposals.module.css";
 import { tabValues, mapProposalsTokensByTab } from "./helpers";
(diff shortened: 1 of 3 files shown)


victorgcramos:
Tested here and it LGTM! Good job!

Issue #1471, opened 2019-10-04
[cms] Update invoice submission for SubContractors


alexlyp:
https://github.com/decred/politeia/pull/1023

This recent politeia PR adds the ability to denote a given line item as a billed subcontractor hours.

It also adds a new request so users may return a list of subcontractors that are using them as a supervisor. This will allow us to populate a dropdown (when in the new type 4) with their list and they can simple select their subcontractor for that line item.

victorgcramos:
Just waiting for https://github.com/decred/politeia/pull/1023 to be merged to jump into this!

Issue #1470, opened 2019-10-04
[cms] Admin CMS Manage User Tab


alexlyp:
Administrators need easy access to be able to update user's Contractor Type, Domain and Supervisor User ID.  This can be accomplished by adding a CMS only UserDetails tab.

victorgcramos:
okay @alexlyp!

PR #1479, opened 2019-10-09
Markdown refactor

victorgcramos:
I’ve been working on this over the past days, and I came up with this solution for our Markdown diff issues.

## The main problem
Before going through the idea of this PR, I would like to discuss what are the issues presented on our current markdown diff.

The diff process consists of identifying which elements have changed and displaying them in the correct position regarding its raw text.

### Elements Identification
Our current markdown renderer (react-markdown) doesn’t have a compiler that can parse all elements without declaring a component. Because of that, we are limited to tag diffs directly on the raw text.

Previously, the diff process was applied to the raw text like the example above:

**Old raw Text:**
```
Hi, this is the first version
```

**New raw Text:**
```
Hi, this is the second version
```

**Diff on raw texts**
```
Hi, this is the <span class=“removed”>first</span> <span class=“added”>second</span> version
```

This example works fine because it’s not complicated. But the tagging process on our markdown diff caused some issues:
- The renderer breaks for new line skips
- Dealing with allowed tags on our renderer is not an easy task since the user can write its custom tags to mess up with the system
	> I can add this `<span>` tag by myself, and the renderer would process it anyway.

- Elements that are not paragraphs such as headers, links, **bold** and _italic_ can’t display tags inside it, so we can’t do the diff for those elements on the raw text

### Element displaying
Not every change in the raw text leads to a change in the rendered markdown. For example, you can press `enter` N times, but the rendered markdown will only show one line skip, you can change the link URL and keep the same alt text.

```
> Raw text 1:

# Header

Paragraph

> Raw Text 2:

# Header



Paragraph
```
will both result on:
****
# Header

Paragraph
****

The following image is an example of how important it is to keep the position of the elements correct:
Screen Shot 2019-10-09 at 00 40 13

How to deal with the mentioned invisible changes, with the correct positioning, and display the information for the user without affecting the UX?

## The solution

After discussing with @fernandoabolafio, I came up with the idea of changing our markdown render to a new one that has this `compiler` feature. Also, it requires that the new lib allows custom renderers so that we can keep our XSS filters, external link modal, etc.

I found this one: [Markdown-to-jsx](https://github.com/probablyup/markdown-to-jsx).

This is an excellent lib that renders markdown strings to jsx elements, allows custom renderers, and has the compiler so we can work with the **output**, not the **input** anymore. All main requirements are satisfied.

Another good point of working with the rendered markdown is that we can see exactly which elements (not string anymore) have changed, and what were the changes (props, children, type)...

I think this diagram can illustrate my point:

MarkdownDiffDiagrams
 
In my opinion, dealing with the rendered content is way better for both positioning and identification.

# Markdown Diff

| Version 1 | Version 2|
|-|-|
|Screen Shot 2019-10-09 at 21 55 35|Screen Shot 2019-10-09 at 21 56 06|

The table below allows us to compare the diff that we have today (**React-markdown**) with the new diff that this PR aims to implement, using the **Markdown-to-jsx** lib.

|React-markdown|Markdown-to-jsx|
|-|-|
| Screen Shot 2019-10-09 at 22 05 57 | react-to-markdown-diff|

Another example:

|React-markdown|Markdown-to-jsx|
|-|-|
| react-markdown-diff-links| markdown-to-jsx-diff-links|


victorgcramos:
The description is not done yet, but I opened this PR so we can discuss the issues regarding the current Diff version that we have today.

tiagoalvesdulce:
Hey @victorgcramos, you raised a good point. Our markdown diff solution is very rudimentary right now. I have some questions about this `markdown-to-jsx` lib:

1. Is it well maintained? Who maintains it? It looks like a really small library in comparison with the `react-markdown` lib.
2. How hard it is for us to customize it as we please?

victorgcramos:
Hi @tiagoalvesdulce!

1. I agree that this lib is smaller, but I think it is enough for what we need. The contributors' community is active (PR opened 3 days ago) for what I've seen. The project is maintained by [volunteers](https://github.com/probablyup/markdown-to-jsx#contributors) and used by almost 8k projects.

    I think it worths the shot based on what we do need. What about you?

2. This is the best feature that this lib allows us to. The markdown compiler allows a custom renderer method that describes what should happen when some element is rendered. It takes `type`, `props` and `children` as arguments, and returns a React element. Check it out [here](https://github.com/probablyup/markdown-to-jsx#optionscreateelement---custom-reactcreateelement-behavior) for more info regarding the customization.

tiagoalvesdulce:
Thanks, @victorgcramos. I think you have some good points. Let's give it a try!

fernandoabolafio:
Looking pretty good to me!  Let's give it a try. Once this PR is ready for final review I can push this branch to pi-redesign so we can test the diff with "real data" from proposals.decred.org.

degeri:
So I did a XSS test review and also tried all the old markdown related issues. The XSS filters and protections we have put in place still seem to be working. LGTM

victorgcramos:
Okay, @degeri and @fernandoabolafio thanks for your feedback! The new markdown is working, but the diff doesn't work well yet, I can see. So, I will fix those errors that Fernando mentioned, and push a fix for this. Thanks again!

tiagoalvesdulce:
@victorgcramos what's the status of this PR? Are you still working on it?

victorgcramos:
@tiagoalvesdulce sure I am! I took some time to work on #1470 and #1471, and then I will come back to this PR.

victorgcramos:
Holding this on until DCC is done since DCC needs higher priority.

tiagoalvesdulce:
I'll be closing this PR because it's been on hold for some time. Feel free to open a new one when you have this functionality more polished @victorgcramos

PR #1481, opened 2019-10-09
Fix fetching of proposal details based on current user access level

fernandoabolafio:
This PR solves an issue where the proposal detail was loading indefinitely when the prop fetched from the API didn't have files. It can happen when a user without the appropriate access level tries to access an unreviewed or censored proposal.

### DIFF
--- a/src/containers/Proposal/Detail/hooks.js
+++ b/src/containers/Proposal/Detail/hooks.js
@@ -137,7 +137,7 @@ export function useProposal(ownProps) {
         setProposal(prop);
       }
     },
-    [getProposalFromCache, proposal]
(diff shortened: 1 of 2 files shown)


src/containers/Proposal/Detail/hooks.js
```diff
@@ -114,7 +132,7 @@ export function useProposal(ownProps) {
 
   useEffect(
     function handleProposalChanged() {
-      const prop = proposalWithFilesOrNothing(getProposalFromCache());
+      const prop = getProposal();
```
victorgcramos: This is causing this warning:
```
React Hook useEffect has a missing dependency: 'getProposal'. Either include it or remove the dependency array
```

PR #1482, opened 2019-10-09
Remove the huge container for the 'go back' link

fernandoabolafio:
If you access a proposal directly without being authenticated, there won't be a  "go back" link, neither a "new proposal" button. Resulting in a huge white container on top of the page:
Screen Shot 2019-10-09 at 16 31 34

The  "New Proposal" button isn't really needed at this page, so I simplified it all by removing the white container at the top of the page and keeping only the  "go back" link when the user is coming from the proposals list:

Screen Shot 2019-10-09 at 16 22 11

### DIFF
--- a/src/containers/Proposal/Detail/Detail.jsx
+++ b/src/containers/Proposal/Detail/Detail.jsx
@@ -1,5 +1,5 @@
 import React, { useMemo, useCallback } from "react";
-import { Link, PageDetails } from "pi-ui";
+import { Link } from "pi-ui";
 import { withRouter } from "react-router-dom";
 import Proposal from "src/componentsv2/Proposal";
 import styles from "./Detail.module.css";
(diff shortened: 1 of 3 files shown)


fernandoabolafio:
@MariaPleshkova is that ok for you?

victorgcramos:
LGTM! It is working as expected.

MariaPleshkova:
@fernandoabolafio in general, looks good. But don't you think it's a little bit confusing that the same "go back" link can be both inside this white container and without it at all? I mean, the same page, the same link, but different interface

fernandoabolafio:
@MariaPleshkova
> "go back" link can be both inside this white container and without it at all?
There won't be a "white container" on this page anymore with the changes applied here.  So either the link will appear by itself or nothing will be displayed on top of the proposal content.

MariaPleshkova:
@fernandoabolafio got it. So, you removed "New proposal" button from all the screens except Public proposals, right?

MariaPleshkova:
@fernandoabolafio ok then

PR #1483, opened 2019-10-09
[redesign] Proposal paywall polling polishment

### LINKED ISSUE #1472: [redesign] Proposal paywall polling needs polishment
There are a few things we need to improve on the proposal credits paywall:
- The paywall polling should be triggered only when the current user accesses its own account credits. Right now the polling is fired even when we access another user's account credits.
- The paywall is triggered every time we access the credits tab, despite the user having an active purchase or not. This should be adjusted to:
   - The user generates a new payment that triggers the polling.
   - When the user accesses the credits tab, we ask the backend once (or a limited number of times) if there is any pending transaction, if there is we keep polling, otherwise we stop it.

amassarwi:
According to #1472 the following improvements are needed regarding proposals paywall payment polling: 
 - [x] The paywall polling should be triggered only when the current user accesses its own account credits.
- [x] The paywall is triggered every time we access the credits tab, despite the user having an active purchase or not. This should be adjusted to:
  - The user generates a new payment that triggers the polling.
  - When the user accesses the credits tab, we ask the backend once (or a limited number of times) if there is any pending transaction, if there is we keep polling, otherwise we stop it.

closes #1472


src/actions/api.js
```diff
@@ -88,7 +88,7 @@ export const setPollingPointer = paymentpolling => {
   globalpollingpointer = paymentpolling;
 };
 
-const POLL_INTERVAL = 10 * 1000;
+const POLL_INTERVAL = 3 * 1000;
```
amassarwi: Just for debugging, will revert it when to merge

Issue #1490, opened 2019-10-14
Redux state normalization


fernandoabolafio:
As previously discussed here: https://github.com/decred/politeiagui/pull/1448#issuecomment-535465091 we have technical debt in the way we currently cache data fetched from the API. Our caching works per endpoint versus per data entity as it should be. The biggest issues caused by this approach are:
 - Violates the "single-source-of-truth" principle, as the same data can be cached in different places.
 - Hard to give maintenance as, for instance, all state modifiers are in the same reducer/file.
 - Slow down the overall scalability.

## Functional comparison

**Current flow:**

![Untitled Diagram (3)](https://user-images.githubusercontent.com/14864439/66762813-f3932780-eea6-11e9-8418-44ae5b8a7dc2.png)

**Flow suggested:**

![Untitled Diagram (4)](https://user-images.githubusercontent.com/14864439/66762877-16bdd700-eea7-11e9-9d59-a8c00914bde1.png)

As shown in the flowcharts above, the new approach adds a second reducer that reacts to "RECEIVE_DATA" actions and properly creates/updates a normalized state accordingly to each entity's needs and structure.  The API reducer will be kept only for managing "control values" such as the params used for the request, "isRequesting" and "hasError". 

## Structural comparison

![Untitled Diagram (5)](https://user-images.githubusercontent.com/14864439/66763278-dc086e80-eea7-11e9-8739-9bff7b38814b.png)

The major structural change will be uncoupling those huge API modules into smaller and better manageable modules.

## Execution

These will be a progressive work and will be done by parts. The checklist of entities to be moved into its own state branch are:
- [x] Comments
- [x] Proposal Credits
- [x] Proposals
- [x] Proposals votes
- [x] User
- [ ] Paywall
- [x] Invoices

thi4go:
I'm thrilled with this ongoing refactor on our state management system. Many thanks for pulling this out mate. Decoupling our state from the gigantic `api` module will have a major impact on the code's maintainability and scalability. By having the state separated by contexts (proposals, user, votes...), we will have each part of the state management control living close to each other, that is, the actions, reducers and selectors. This will result in a much more comprehensive flow of data, instead of having to go through enormous files to understand whats going on, or to add new features.

I'd like to help out on this process. Thought of normalizing the user data now. Any considerations @fernandoabolafio? As soon as you ok this I can get started on it.

victorgcramos:
[...]

PR #1492, opened 2019-10-14
Create normalized state for caching proposals data

fernandoabolafio:
This PR deals with the normalization of the state for proposals state. It includes both the proposals state and the proposals vote state which are two different branches.
This is part of https://github.com/decred/politeiagui/issues/1490.


thi4go:
Think I found something. When I approve a proposal when viewing it, everything is ok. But when I approve a proposal directly from the Admin page, I get this error:

image

The proposal does change its status to approved, but it redirects to the error page.

fernandoabolafio:
@thi4go thanks for testing it and nice catch! It's fixed now, ready for another review.

Issue #1504, opened 2019-10-21
[redesign] First click on submit in proposal form doesn't fire action


thi4go:
This is a minor bug on proposal form; when I edit the body of a proposal, submit works properly when first clicked. When I edit only the title of the proposal, I have to click submit button two times for it to fire up.

![submit_bug](https://user-images.githubusercontent.com/2729122/67209973-a9ea9580-f3ee-11e9-9c2b-ae919fdfb60d.gif)

amassarwi:
@thi4go I can have a look, if you are not working on it @thi4go :thinking: ?

thi4go:
Go ahead mate, I'm not working on it currently 👍

amassarwi:
@fernandoabolafio can be closed, was fixed with #1508

PR #1508, opened 2019-10-21
[redesign] Fix edit proposal name `double click` issue

amassarwi:
This PR closes the issue mentioned in #1504.

The onBlur handler was blocking the submit action if name text field was active(focused) - click on submit was calling the input's onBlur handler instead of submit button (e.g user was still able to submit by pressing the enter key).

### DIFF
--- a/src/componentsv2/ProposalForm/ProposalForm.jsx
+++ b/src/componentsv2/ProposalForm/ProposalForm.jsx
@@ -17,18 +17,15 @@ import useBooleanState from "src/hooks/utils/useBooleanState";
 const ProposalForm = React.memo(function ProposalForm({
   values,
   handleChange,
-  handleBlur,
   handleSubmit,
   isSubmitting,
-  touched,
   setFieldValue,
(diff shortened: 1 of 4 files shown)


src/componentsv2/ProposalForm/ProposalForm.jsx
```diff
@@ -98,21 +95,18 @@ const ProposalForm = React.memo(function ProposalForm({
...
       <MarkdownEditor
         name="description"
         className="margin-top-s"
         value={values.description}
         textAreaProps={textAreaProps}
         onChange={handleDescriptionChange}
-        onBlur={handleBlur}
         placeholder={"Write your proposal"}
-        error={touched.description && errors.description}
+        error={errors.description}
```
amassarwi: The check touched[key] is redundant in both cases - it's an edit form so user can start only from valid start which may become invalid only if user `touches` one of the user (incorrectly), this why onBlur is not needed as well here.

PR #1517, opened 2019-10-23
[redesign] Show user public key on identity tab

### LINKED ISSUE #1515: [redesign] Current user pub key is not shown to other users
The current user public key should be shown under the user account -> identity tab for every user who accesses it. This is how it is currently shown:
Screen Shot 2019-10-23 at 18 02 26

thi4go:
closes #1515


thi4go:
Fixed the selector that was causing this bug mate! @fernandoabolafio

victorgcramos:
LGTM

PR #1549, opened 2019-11-01
[cms] Admin CMS Manage User Tab

### LINKED ISSUE #1470: [cms] Admin CMS Manage User Tab

victorgcramos:
Closes #1470


alexlyp:
So overall the UI/UX looks good for a first pass.

Only issue I'm seeing is it would be nice to show a more clear error but not sure it's possible.  Getting a 1050 on the backend "invalid supervisor id".  It shows the cause on the backend (user id not a supervisor) but nothing on the front end.  Believe that would require a larger change though.  So not really a requirement for this PR.

Overall great job.

Next I'd like to add a dropdown instead of a text box for selecting available supervisor user id's.  Once this is merged we can discuss that.

src/helpers.js
```diff
@@ -50,6 +50,7 @@ export const getTextFromJsonToCsv = file => {
 
 export const getHumanReadableError = (errorCode, errorContext = []) => {
   const genericContactMsg = "please contact Politeia administrators";
+  console.log(errorContext);
```
tiagoalvesdulce: You left a log here.

victorgcramos: oh sorry! removing it right now

PR #1553, opened 2019-11-02
Fix proposal details page back icon text

### LINKED ISSUE #1544: "Go back" issues
The "go back to" link on a proposal appears to be misbehaving

1. Navigate to any proposal from the home page. Notice the text says "Go back to public proposals".
2. Click the author's name.
3. Press the browser's native back button to return to proposal. Notice the text now says "Go back to user account". Click that link.
5. User is not taken to the author's account.

amassarwi:
closes #1544

### DIFF
--- a/src/containers/Proposal/Detail/Detail.jsx
+++ b/src/containers/Proposal/Detail/Detail.jsx
@@ -41,18 +41,15 @@ const ProposalDetail = ({ Main, match }) => {
   const goBackLinkFromPreviousLocation = useMemo(() => {
     if (!previousLocation) return null;
     let message = "";
-
     switch (previousLocation.pathname) {
-      case "/":
-        message = "Go back to public proposals";
-        break;
       case "/proposals/unvetted":
         message = "Go back to unvetted proposals";
         break;
+      case (previousLocation.pathname.match("/user/*?tab=proposals")):
+        message = "Go back to user account";
+        break;
       default:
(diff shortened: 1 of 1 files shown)


src/containers/Proposal/Detail/Detail.jsx
```diff
@@ -41,18 +41,15 @@ const ProposalDetail = ({ Main, match }) => {
...
     let message = "";
-
     switch (previousLocation.pathname) {
-      case "/":
-        message = "Go back to public proposals";
-        break;
       case "/proposals/unvetted":
         message = "Go back to unvetted proposals";
         break;
+      case (previousLocation.pathname.match("/user/*?tab=proposals")):
```
amassarwi: More strict pattern to handle case when navigating to proposal details form user's proposals tab.

amassarwi:
I still get unexpected behavior in some cases following these steps:
1. Navigate to user page by clicking on one of the proposals author's name
2. Proposals tab
3. Click on one of the proposals 
4. Proposal details showing `go back to public proposals`

I'd expect it to show `back to user account` tho.
The `search` property of `previousLocation` object is null and this why `previousLocation.pathname.match("/user/*") && previousLocation.search.includes("tab=proposals")` is false, therefore `go back to public proposals` text is showing, is there a way to force forwarding the query params (the search) of the previous route ? :thinking:

Issue #1301, opened 2019-07-02
[cms] Subtotal resetting after editing Labor field


victorgcramos:
This is minor, but when you edit the Labor column of any invoice row (after publishing it), the Subtotal column gets reset.

To reproduce this:
1. Create an invoice
2. Edit the Labor of any row
3. You will get this:
![Screen Shot 2019-07-02 at 03 33 01](https://user-images.githubusercontent.com/22639213/60489616-24927800-9c7b-11e9-81e7-f28ef9f68d2f.png)

victorgcramos:
@fernandoabolafio this was fixed by #1566

PR #1566, opened 2019-11-06
[cms] Allow invoice submission for sub contractors

### LINKED ISSUE #1471: [cms] Update invoice submission for SubContractors

victorgcramos:
Closes #1471.

One quick observation: since there is no `contractor rate` field, the subtotal will not take the `Labor (hours)` line item into consideration when calculating the `Subtotal`. Therefore, the only way to modify the `Subtotal` value is to put the subcontractor's payout amount into the `Expense` field.


fernandoabolafio:
@victorgcramos it's working nicely. Two things I noticed:

1. This is minor but it took me a while to figure out: trim the user ids in the Manage User tab when adding supervisors to a user.

2. Since we don't have the contractor rate info yet, I think we should block the "hours" field and use the expenses only. At least for now. What do you think @alexlyp?
Also, when I set some value in the hours' field, it changes the subtotal to 0, even if I had an expense placed on.

alexlyp:
Hrm yah I suppose that would be a reasonable fix for the time being.  Let me get this contractor rate added to the line item so it can be updated properly.  Let's put a quick hold on this PR in that case.

alexlyp:
OK i've completed work on the required politeia PR.  https://github.com/decred/politeia/pull/1044

You'll need to update your local DB by doing: "ALTER TABLE line_items ADD contractor_rate int;"

victorgcramos:
Just found out that this also Closes #1301

alexlyp:
This needs a rebase

alexlyp:
Hrm getting an error when running it: 

./src/components/DCC/Details/Details.jsx
Module not found: Can't resolve 'snew-classic-ui' in '/home/user/git/src/github.com/decred/politeiagui/src/components/DCC/Details'

PR #1575, opened 2019-11-13
[cms] Submit New DCC

### LINKED ISSUE #1539: [cms] New DCC
The new/submit DCC page is also going to be very similar to "submit invoice."  Though there only needs to be a single text area for the sponsor statement instead of the line items and what not.

There should be a draft option.   There will be no edit opportunities as well.

victorgcramos:
Closes #1539 

This diff adds:
- Submit new DCC page and form
- Save drafts

> This commit is following the old code pattern (the one that we currently use on CMS). As discussed with @fernandoabolafio, this diff can be easily adapted to the new pattern (the one on GUI v2).


alexlyp:
This overall looks really good!
Couple things: 
- Upon successful creation of a new dcc (in which it shows a success message).  Can you hide the form below and maybe add a button that says make a new one, or it can show a button to the future DCC listings page.  Or maybe upon adding the DCC listings page, just go there on success.
-  I'm trying to figure out a better way for users to enter the userid.  What is there now is totally acceptable for MVP, but I'm thinking we could do a dropdown for all nominee cmsusers, for an issuance.  Then they can simply select the user based on email address.  For a revocation, a dropdown of usernames would probably be better. Though I guess we need to discuss privacy concerns for both of these approaches.
- I'm thinking when revocation is selected, contractor type and domain can be hidden.  Domain should be populated from the given cmsuser's user details.  And I think we can just set the contractor type to Revoked for a revocation request.  I need to review the backend code to confirm whether this is acceptable.

victorgcramos:
Good review @alexlyp. Thanks a lot! I'll address those comments soon. I started working on the whole DCC tree, and I'll push those changes together with the rest of the tree, okay? Thank you!

victorgcramos:
New DCC and other DCC stuff are on #1611

Issue #1577, opened 2019-11-15
High CPU usage after upvote


xaur:
- I click upvote comment button
- After 3 seconds the upvote is added (which is too long on its own, but that's a separate issue)
- Then the CPU usage spikes to 100% for ~10 seconds and Firefox offers me to kill the offending script: https://proposals.decred.org/static/js/16.ea6869a6.chunk.js:1

This wasn't the case before the redesign. My browser setup didn't change.

xaur:
When posting a comment I had to cancel two "Kill this script?" popups. I wonder what numbers is it crunching so heavily?

fernandoabolafio:
@xaur in what proposal did it happen?

xaur:
The upvote CPU drain happened in both [Spanish Marketing](https://proposals.decred.org/proposals/5af0ce1cd325be6be39109c2750f34095c4e8feeea962ede058a1e4f4a61473e) and [TinyDecred budget](https://proposals.decred.org/proposals/ad0f9688b3467734e2581604914b2cc32c6eb7991dff460eff41d21f66d88451) proposals.

The post-comment one happened in Spanish Marketing proposal but not in TinyDecred budget.

xaur:
Was making a few comment upvotes today in the Latam proposal and there is no more freezing of Firefox. The CPU spike around 10 seconds after the upvote is down from 100% to 20% and doesn't last long. And to be fair it's hard to be sure that other stuff is not happening in the system.

fernandoabolafio:
Ok @xaur I will consider this as closed because of https://github.com/decred/politeiagui/pull/1578. If you find any other issues in this regard, feel free to open more issues.

PR #1578, opened 2019-11-15
Improve general performance on Proposal Detail page

### LINKED ISSUE #1583: Comment voting not working properly
Per @RichardRed0x in decred/politeia#1053.

"For the last day or so, most of the time when I have tried to vote on comments by clicking the thumbs up/down it has not registered my vote."

#1578 is suppose to fix the issue.  This needs to be confirmed before this issue can be closed.

fernandoabolafio:
This PR adds many points of performance improvements for the Proposal Detail page. This intended to fix the high CPU usage reported on https://github.com/decred/politeiagui/issues/1577. While I couldn't reproduce an extreme case as reported in the issue where the browser offer to "kill the tab" I did notice that, especially for Firefox, the response to actions such as voting on a comment or posting a comment was much slower than Chrome.
After profiling what was happening I found many opportunities to prevent needless repaintings (or re-renders). 
By using the React profiling tool I could reduce the most costly rendering step from ~163ms to ~9ms. In the screenshots below it is possible to see how the number of components affected for an action reduced drastically:
**Before:**
Screenshot 2019-11-18 at 12 42 03

**After:**
Screenshot 2019-11-18 at 12 43 06

There is still room for improvement among other components. For that, I will create an issue which shall be done after the competition of #1490 which will probably modify the selectors' structures which affect directly the changes required for improving performance during state updates.

closes #1583

### DIFF
--- a/src/Appv2/Loader/hooks.js
+++ b/src/Appv2/Loader/hooks.js
@@ -1,38 +1,24 @@
 import * as sel from "src/selectors";
 import * as act from "src/actions";
-// import { or } from "src/lib/fp";
-import { useRedux } from "src/redux";
+import { useSelector, useAction } from "src/redux";
 
-const mapStateToProps = {
-  user: sel.currentUser,
-  userPubkey: sel.currentUserPublicKey,
-  currentUserEmail: sel.currentUserEmail,
-  currentUserID: sel.currentUserID,
-  lastLoginTime: sel.currentUserLastLoginTime,
-  keyMismatch: sel.getKeyMismatch,
-  apiError: sel.apiError,
-  userCanExecuteActions: sel.userCanExecuteActions,
-  onboardViewed: sel.onboardViewed,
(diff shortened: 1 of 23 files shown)


xaur:
> I did notice that, especially for Firefox, the response to actions such as voting on a comment or posting a comment was much slower than Chrome.

@fernandoabolafio Chrome is doing a ton of optimizations to javascript to run it at almost native speed. Chrome itself is heavily optimized and afaik critical parts are even written in assembly. Optimizations are good for users in general because things work faster and the battery is saved. Isn't it great when stuff works fast? Not always. When your shiny new CPU is churning myriads of instructions with the help of Chrome's optimizer, it's easy to stop seeing slow code and let the bloat creep in.

If you want to never miss slow front end code, here's my simple recipe: get an older laptop, get Firefox and disable all js optimizations. If that's not enough you can always throttle your CPU ;)

PR #1581, opened 2019-11-18
Correctly select which proposals have to be fetched next

### LINKED ISSUE #1580: Missing proposals on public list
If a user access the website directly into a proposal detail page and then go to the public list page. Depending on which position the proposal accessed firstly has in the list, it can cause the lazy loading to skip the proposal before it in the list.

fernandoabolafio:
closes #1580


fernandoabolafio:
Changing it to wip b/c I just found I need to test a few more edge cases.

PR #1585, opened 2019-11-20
[cms] User proper userid field for request

alexlyp:
Previously user.id was undefined and thus causing any manageCmsUser request to fail.


victorgcramos:
LGTM

PR #1590, opened 2019-11-23
Use dcrdata.decred.org for block explorer

### LINKED ISSUE #1589: Update links to voting charts
dcrdata.decred.org should be used instead of mainnet.dcrdata.org for all block explorer links.

amassarwi:
Closes #1589

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -6,7 +6,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node-version: [12.x, 13.x]
+        node-version: [12.x]
 
     steps:
     - uses: actions/checkout@v1


src/componentsv2/DcrTransactionLink.jsx
```diff
@@ -4,7 +4,7 @@ import { useLoaderContext } from "src/Appv2/Loader";
 
 const DcrTransactionLink = ({ txID }) => {
   const { apiInfo } = useLoaderContext();
-  const hostName = apiInfo.testnet ? "testnet" : "explorer";
+  const hostName = apiInfo.testnet ? "testnet" : "dcrdata";
```
amassarwi: This isn't good enough, with this change, url gonna look like `https://dcrdata.dcrdata.org/tx/${txID}` which is wrong, fixing it

PR #1597, opened 2019-12-01
set button type to button.

dajohi:
The default button type is submit, but we are using onClick.  By
setting the type to button, the button does nothing if js is
disabled.

Firefox recommends explicitly setting the button type.


tiagoalvesdulce:
LGTM but depends on https://github.com/decred/pi-ui/pull/196

Issue #1601, opened 2019-12-02
Duplicated comments 


fernandoabolafio:
In this thread, the "plusbit" comment was replicated 3 times: https://proposals.decred.org/proposals/e559188b0febcab29c49c1f7dd5c66645e31be00894a150ef7d0b8ceb6486605/comments/4. 
From the [raw data](https://github.com/decred-proposals/mainnet/blob/master/e559188b0febcab29c49c1f7dd5c66645e31be00894a150ef7d0b8ceb6486605/1/plugins/decred/comments.journal
) we can notice a few things:
- 2 out of the 3 comments have exactly the same body text, thus the same signature.
- 1 out of the 3 comments is slightly different because it has a blank space at the end of the body text.

It is likely that multiple requests are being fired from the GUI when submitting a new comment but I couldn't reproduce this behavior so far. 

One thing that is clear we need to do is to trim comments to make sure it doesn't make to the API with blank spaces at the end of the text. This should be addressed by: https://github.com/decred/politeiagui/pull/1600.

lukebp:
Another instance of this happening.

https://matrix.to/#/!VFRvyndKpzcLrVslQD:decred.org/$157581204658910QrbZF:decred.org?via=decred.org&via=matrix.org&via=zettaport.com

lukebp:
@xaur posted a detailed bug report here with more details.

https://github.com/decred/politeia/issues/931#issuecomment-562850323

cc: @amassarwi

xaur:
Just had a situation very similar to#931: long "loading" after pressing "Add comment" that ended with "Server error". I figured that my comment got stuck somewhere and went ahead to post other comments. After several minutes (like 10?) my first comment [appeared](https://proposals.decred.org/proposals/012b4e335f25704e28ef196d650316dca421f730225d39e37b31b3c646eb8497/comments/81).

lukebp:
hmm so it sounds like the politeia backend was busy doing something, your request timed out, and then the backend added the comment successfully after the backend finished doing whatever it was doing.  I'd have to look into it more, but there really only a couple of things that could hold politeia up like that.

[...]

PR #1617, opened 2019-12-09
First comment issue fix

amassarwi:
While investigating #1601 I've found another issue:
Submitting new comment on a post with no comments didn't work.

I've added a small fix.

### DIFF
--- a/src/containers/Comments/Comment/CommentWrapper.jsx
+++ b/src/containers/Comments/Comment/CommentWrapper.jsx
@@ -52,7 +52,7 @@ const CommentWrapper = ({ comment, children, numOfReplies, ...props }) => {
   }, [showReplies]);
 
   const handleSubmitComment = useCallback(
-    async comment => {
+    comment => {
       return onSubmitComment({
         comment,
         token,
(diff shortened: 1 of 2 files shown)


src/containers/Comments/Comment/CommentWrapper.jsx
```diff
@@ -52,7 +52,7 @@ const CommentWrapper = ({ comment, children, numOfReplies, ...props }) => {
   }, [showReplies]);
 
   const handleSubmitComment = useCallback(
-    async comment => {
+    comment => {
```
amassarwi: @fernandoabolafio 
Could this be the cause of #1601 ? :open_mouth: :open_mouth:

fernandoabolafio: I don't see why it would be the cause. Did you notice anything different after your change?

amassarwi: I couldn't reproduce the issue locally, so not sure if it changed anything in the context of 1601

fernandoabolafio: right, so why this change is needed then?

amassarwi: it's an async function without any await calls & linter complained about it.

PR #1619, opened 2019-12-11
Improve image components for easier use accross component

fernandoabolafio:
This PR reorganize and improve the code for files presentational components (Thumbail, ThumbnailGrid, etc). This makes easier to integrate these components when developing features.


src/componentsv2/Files/ThumbnailGrid.jsx
```diff
@@ -0,0 +1,87 @@
...
+    : [fileErrors];
+  return uniqueErrors.map(
+    err =>
+      err && (
+        <p key={err} className={styles.fileErrorText}>
+          {err}
+        </p>
+      )
+  );
+};
```
tiagoalvesdulce: Can we put it in another file?

fernandoabolafio: done

PR #1656, opened 2020-01-10
[cms] Properly handle January on SubmitInvoice

### LINKED ISSUE #1652: [cms] Properly handle January on SubmitInvoice
Due to the current month handling in SubmitPageInvoice, when the current month is January and the user makes no changes, it ends up submitting the invoice month/year as 0/2020 (for example now).  

I have added PR for validation to no accept those month/year combination any more but this corner case should be fixed on the GUI side as well.

victorgcramos:
Closes #1652

### DIFF
--- a/src/components/snew/SubmitPage/SubmitPageInvoice.js
+++ b/src/components/snew/SubmitPage/SubmitPageInvoice.js
@@ -1,4 +1,4 @@
-import React, { useState, useEffect, useCallback, useMemo } from "react";
+import React, { useState, useEffect, useCallback } from "react";
 import ReactBody from "react-body";
 import ErrorField from "../../Form/Fields/ErrorField";
 import SelectField from "../../Form/Fields/SelectField";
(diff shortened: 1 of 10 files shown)


src/lib/editors_content_backup.js
```diff
@@ -185,6 +185,12 @@ export const resetNewInvoiceData = () => {
 };
 
 export const getNewInvoiceData = () => {
+  const isFirstMonth = getCurrentMonth() === 1;
+  const defaultYearValue = isFirstMonth
+    ? getCurrentYear() - 1
+    : getCurrentYear();
+  const defaultMonthValue = isFirstMonth ? 12 : getCurrentMonth();
+
```
tiagoalvesdulce: Here would be nice if you create a function like: `getDefaultDate()` and returned both the month and the year. This will improve readability.

victorgcramos: Good suggestion!

src/components/snew/SubmitPage/SubmitPageInvoice.js
```diff
@@ -46,20 +45,20 @@ const InvoiceSubmit = (props) => {
...
   const [contractorRate, setContractorRate] = useState(0);
 
-  const yearOptions =
-    getCurrentMonth() === 1
-      ? YEAR_OPTIONS.slice(0, YEAR_OPTIONS.length - 1)
-      : YEAR_OPTIONS;
+  const { yearOptions, year: currentYear } = useMemo(
+    () => getCurrentYear(),
+    []
+  );
```
tiagoalvesdulce: why do we need useMemo here? it's not an expensive computation and it has no dependencies at all.

PR #1666, opened 2020-01-13
[cms] DCC comments

victorgcramos:
This diff adds the comments module for the DCCs


src/actions/api.js
```diff
@@ -847,7 +855,8 @@ export const onSubmitComment = (
   comment,
   parentid,
   commentid,
-  isCMS = false
+  isCMS = false,
+  isDCC = false
```
tiagoalvesdulce: Ideally, this action shouldn't have those conditions (`isCMS`, `isDCC`, ...) because it goes against modularization (every new entity should require changing code inside the action). What we should do instead is to create 3 different functions (1 for CMS, 1 for DCC and 1 for pi). This way, to add support to a new entity the developer can create an entirely new chunk of functionality (function) instead of a conditional into existing code. Then, in the hooks/connectors the config object can be accessed to pass the correct function to the component level. This may or may not be done in this PR (it's up to you). I'm just leaving my thoughts here. If you think it's out of the scope of this PR, please create an issue for us to keep track of it if you agree with me.

src/connectors/proposal.js
```diff
@@ -64,23 +65,30 @@ class Wrapper extends React.PureComponent {
 
   handleViewAllClick = (e) => {
     e && e.preventDefault() && e.stopPropagation();
+    const isDCC = this.props.location.pathname.includes("/dcc/");
```
tiagoalvesdulce: Can we bubble this verification up? I feel like it should be up in the component tree or maybe in the config object.

src/connectors/reply.js
```diff
@@ -87,7 +88,8 @@ class Wrapper extends React.PureComponent {
           onToggleMarkdownHelp: this.onToggleMarkdownHelp.bind(this),
           showCommentForm: this.state.showCommentForm,
           toggleCommentForm: this.toggleCommentForm,
-          onCloseCommentForm: this.onCloseCommentForm
+          onCloseCommentForm: this.onCloseCommentForm,
+          isDCC: this.props.location.pathname.includes("/dcc/")
```
tiagoalvesdulce: With my previous suggestion, this would go out

src/actions/api.js
victorgcramos: I agree with you, @tiagoalvesdulce! Makes total sense! But I think it might be better to solve it on another issue. Thank you!

src/components/DCC/Comments/hooks.js
```diff
@@ -0,0 +1,28 @@
+import { useEffect } from "react";
+// import { useDCC } from "../DCCProvider";
```
tiagoalvesdulce: old hook?

src/components/DCC/Comments/styles.css
```diff
@@ -0,0 +1,29 @@
+.dccComment {
+  background-color: white;
+  padding: 10px;
+  border-bottom: solid #eee 1px;
+  margin: 0px 1em;
+}
+
+.dccCommentHeader {
```
tiagoalvesdulce: empty class

src/components/DCC/DCCProvider.jsx
```diff
@@ -0,0 +1,23 @@
...
+  const [isDCC, setIsDCC] = useState();
+
+  useEffect(() => {
+    setIsDCC(location.pathname.includes("/dcc/"));
+  }, [location]);
+
+  return <dccCtx.Provider value={{ isDCC }}>{children}</dccCtx.Provider>;
+};
+
+export default withRouter(DCCProvider);
```
tiagoalvesdulce: well done!

src/components/snew/Subreddit.js
```diff
@@ -26,19 +27,20 @@ class CustomSubreddit extends Component {
...
-        <Route path="/user/signup" component={noSidebar(this.props)} />
-        <Route path="/user/login" component={noSidebar(this.props)} />
-        <Route
-          path="*"
-          component={
-            this.props.isCMS ? noSidebar(this.props) : withSidebar(this.props)
-          }
-        />
-      </Switch>
+      <DCCProvider>
```
tiagoalvesdulce: Great. Now we can use this DCC context to put everything all DCC routes should have access.

src/components/DCC/Comments/hooks.js
victorgcramos: yes hehe, sorry!

PR #1695, opened 2020-01-16
Sync cmsv2 with master tip

thi4go:


fernandoabolafio:
I don't if that is the case but to avoid messing up with the commit history, prefer to do `git merge master` in the branch you want to sync rather than `git rebase`.

thi4go:
Understood it better now mate, will be correcting this @fernandoabolafio 👍

PR #1701, opened 2020-01-17
[cms-v2] Generate Payouts

amassarwi:
This PR adds `Generate Payouts` action to cms admin page - which redirects to payouts page.
Admins will be able to export payouts as .csv & mark invoices as paid.


src/containers/Invoice/Admin/List.module.css
```diff
@@ -4,3 +4,7 @@
   justify-content: center;
   align-items: center;
 }
+
+.actionsWrapper {
+  width: 69%;
```
tiagoalvesdulce: why 69%? How did you come up with this number?

src/containers/Invoice/Payouts/PayoutsList.jsx
```diff
@@ -0,0 +1,74 @@
...
+import { Row } from "src/componentsv2/layout";
+import styles from "./PayoutsList.module.css";
+import { useAdminInvoiceActions } from "../Actions";
+
+
+const PayoutsList = ({ TopBanner, PageDetails, Main }) => {
+  const { loading, payouts } = useAdminPayouts();
+  const { onPay } = useAdminInvoiceActions();
+  const actions = !loading && payouts && payouts.length ? (
+    <Row justify="space-between" className={styles.actionsWrapper}>
```
tiagoalvesdulce: Can you create a variable for this boolean expression`!loading && payouts && payouts.length`?

src/containers/Invoice/Payouts/PayoutsList.jsx
```diff
@@ -0,0 +1,74 @@
...
+                approvedtime: new Date(approvedtime * 1000).toUTCString(),
+                year,
+                month,
+                contractorname,
+                contractorrate: contractorrate / 100,
+                labortotal: labortotal / 100,
+                expensetotal: expensetotal / 100,
+                total: total / 100,
+                exchangerate: exchangerate / 100,
+                dcrtotal: dcrtotal / 100000000,
```
tiagoalvesdulce: we have a utility function to convert atoms to DCR called `convertAtomsToDcr` can you use it so it's more declarative?

src/containers/Invoice/Payouts/PayoutsList.jsx
```diff
@@ -0,0 +1,74 @@
...
+        >
+        </PageDetails>
+      </TopBanner>
+      <Main fillScreen>
+        {loading && (
+          <div className={styles.spinnerWrapper}>
+            <Spinner invert />
+          </div>
+        )}
+        {!loading && payouts && payouts.length ? (
```
tiagoalvesdulce: here, instead of `{condition ? <Component /> : null}` you can go with `{condition && <Component />}`

src/containers/Invoice/Payouts/PayoutsList.module.css
```diff
@@ -0,0 +1,19 @@
+.spinnerWrapper {
+  display: flex;
+  height: 50%;
+  justify-content: center;
+  align-items: center;
+}
+
+.actionsWrapper {
+  width: 69%;
```
tiagoalvesdulce: This number is being repeated. Isn't that the case for a variable?

src/containers/Invoice/Payouts/PayoutsList.jsx
```diff
@@ -0,0 +1,74 @@
+import React from "react";
+import { Link, Spinner, Table } from "pi-ui";
+import { useAdminPayouts } from "./hooks";
+import ExportToCsv from "src/componentsv2/ExportToCsv";
+import { Row } from "src/componentsv2/layout";
+import styles from "./PayoutsList.module.css";
+import { useAdminInvoiceActions } from "../Actions";
+
+
+const PayoutsList = ({ TopBanner, PageDetails, Main }) => {
```
tiagoalvesdulce: Let's apply prop validation to new components. Please use propTypes for this.

src/containers/Invoice/Admin/List.module.css
amassarwi: You're right, will try achieve the same result using fixed `rem` width

tiagoalvesdulce: I mean, it's fine to use percentages but looks like an arbitrary number. It would be nice if you comment on the code explaining how did you come up with the number if there are some calculations behind it.

amassarwi:
@tiagoalvesdulce good notes, were addressed!

amassarwi:
@tiagoalvesdulce all points addressed.

PR #1706, opened 2020-01-20
[cms] Show usernames instead of count for support/oppose

alexlyp:
Instead of showing just the counts of support/opposing users, we should show the usernames.

~This requires https://github.com/decred/politeia/pull/1102~

### DIFF
--- a/src/components/DCC/Details/Details.jsx
+++ b/src/components/DCC/Details/Details.jsx
@@ -46,7 +46,6 @@ const DCCDetail = props => {
         confirm && onChangeDCCStatus(status, reason)
     ) && e.preventDefault();
 
-
   return (
     <div className="content" role="main">
       <div className="dcc-info">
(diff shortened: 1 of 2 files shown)


src/components/DCC/Details/Details.jsx
```diff
@@ -104,8 +103,8 @@ const DCCDetail = props => {
             <p className="dcc-info">{dcc.dccpayload.statement}</p>
 
             <h2>Votes</h2>
-            <DCCInfo label="Support">{dcc.supportuserids.length}</DCCInfo>
-            <DCCInfo label="Against">{dcc.againstuserids.length}</DCCInfo>
+            <DCCInfo label="Support">{dcc.supportusernames}</DCCInfo>
+            <DCCInfo label="Against">{dcc.againstusernames}</DCCInfo>
```
victorgcramos:
```javascript
<DCCInfo label="Support">{getVotesUsernameList(dcc.supportusernames)}</DCCInfo>
<DCCInfo label="Against">{getVotesUsernameList(dcc.againstusernames)}</DCCInfo>
```

and the `getVotesUsernameList` function should be declared on `../helpers` in order to avoid this shrinkage mentioned before.

```javascript
// on src/components/DCC/helpers.js
...

const separateName = length => (name, index) => index < length - 1 ? name + ", " : name;

export const getVotesUsernameList = (names) => names.map(separateName(names.length));

...
```

PR #1714, opened 2020-01-23
[cmsv2] Fix date picker

### LINKED ISSUE #1711: [cms-v2] Opening the datepicker when the exchange rate is being fetched locks the picker
bug_select_date

tiagoalvesdulce:
Fixes #1711 

Replace useEffect hook with useLayoutEffect hook and remove the prop dependency list


src/componentsv2/MonthPicker/MonthPickerField.jsx
```diff
@@ -26,7 +26,7 @@ const pickerLang = {
   to: "To"
 };
 
-const makeText = m => {
+const makeText = (m) => {
   if (m && m.year && m.month)
     return pickerLang.months[m.month - 1] + " " + m.year;
```
amassarwi: We could use here template string instead

PR #1720, opened 2020-01-24
[cms-v2] Draft Invoices

### LINKED ISSUE #1673: [cms-v2] Invoice form draft
Implement the capability of saving an invoice draft.

thi4go:
This PR adds the capability to create draft invoices.

closes #1673


thi4go:
Addressed the comments mate @tiagoalvesdulce ! :+1: 

Played around a bit and thought the button would look better like this, so that we don't bulk up the UI with a full button on the invoice card:

Screenshot from 2020-01-27 15-05-44

src/componentsv2/Invoice/DraftInvoice.jsx
```diff
@@ -29,13 +30,14 @@ const DraftInvoice = ({ draft, onDelete }) => {
                 </Title>
               }
               status={
-                <Status disableMobileView>
-                  <button
+                <Status>
+                  <Link
+                    href="#"
```
tiagoalvesdulce: The problem of being a link with `href="#"` is that it will redirect the user to the top of the page when deleting a draft in a long list. Also, it will modify the URL. Please use a Button and style it to look like a Link :)

thi4go: Oh that's right, my bad. Will fix this up :)

PR #1721, opened 2020-01-25
[cms-v2] Huge cleanup of legacy files/deps and zero-reference code

fernandoabolafio:
This PR does a lot of cleanup in the codebase for the alternative branch cmsv2. Among them are:

- Removal of legacy code. Basically the old `src/components`, `src/hocs` and other unused directories of files.
- Removal of many actions/selectors and other modules not being currently used.
- Because of the change described above many tests were also removed.
- A temporary approach to fix the tests not passing in the cmsv2 was added.
- Remove the dep `snew-classic-ui`.


src/reducers/api/api.js
```diff
@@ -293,73 +242,44 @@ const api = (state = DEFAULT_STATE, action) =>
...
-      [act.REQUEST_PROPOSALS_VOTE_STATUS]: () =>
-        request("proposalsVoteStatus", state, action),
-      [act.RECEIVE_PROPOSALS_VOTE_STATUS]: () =>
-        onReceiveProposalsVoteStatus(state, action),
       [act.REQUEST_PROPOSALS_VOTE_SUMMARY]: () =>
         request("proposalsVoteSummary", state, action),
       [act.RECEIVE_PROPOSALS_VOTE_SUMMARY]: () =>
-        onReceiveProposalsVoteSummary(state, action),
+        receive("proposalsVoteSummary", state, {}),
       [act.REQUEST_PROPOSAL_VOTE_STATUS]: () =>
```
amassarwi: We could remove this action AFAIR, we aren't calling vote status route anymore.

PR #1752, opened 2020-02-06
Allow user to close NavDrawer when the page scrolls down

tiagoalvesdulce:


src/componentsv2/NavigationDrawer/NavigationDrawer.module.css
```diff
@@ -13,6 +13,19 @@
   box-shadow: 0.5rem 2rem 2rem 0px var(--color-shadow-light);
 }
 
+.navigationDrawerHeader {
+  width: 400px;
```
amassarwi: I would suggest working with `rem` also here, what do you think ?

src/componentsv2/NavigationDrawer/NavigationDrawer.module.css
```diff
@@ -13,6 +13,19 @@
   box-shadow: 0.5rem 2rem 2rem 0px var(--color-shadow-light);
```
amassarwi:
```suggestion
  box-shadow: 0.5rem 2rem 2rem 0 var(--color-shadow-light);
```

PR #1754, opened 2020-02-06
[cms-v2] Invoice auto save

### LINKED ISSUE #1672: [cms-v2] Invoice form auto save
Implement the capability of auto-saving the invoice being edited so the changes aren't lost if the user accidentally leaves the page. 
This isn't the same as a draft. The auto-saving saves the changes just for the session and aren't persisted once the tab is closed or the invoice is submitted.

amassarwi:
closes #1672

### DIFF
--- a/src/hooks/utils/useSessionStorage.js
+++ b/src/hooks/utils/useSessionStorage.js
@@ -36,23 +36,20 @@ export default function useSessionStorage(key, initialValue) {
     [key, setError]
   );
 
-  const onStorageChange = useCallback(
-    (event) => {
-      if (event.key !== key) {
-        return;
-      }
+  const onStorageChange = (event) => {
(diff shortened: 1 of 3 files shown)


src/hooks/utils/useSessionStorage.js
```diff
@@ -0,0 +1,61 @@
...
+        setSessionStorageValue(key, newValue);
+        setStoredValue(newValue);
+      } catch (e) {
+        setError(e);
+      }
+    },
+    [key, setError]
+  );
+
+  const onStorageChange = useCallback(
```
tiagoalvesdulce: The `useCallback` hook isn't necessary because we are not passing the `onStorageChange` function down to a child component and the memoized callback isn't optimizing anything. Since the code is more readable without it, I think we can take it out.

PR #1758, opened 2020-02-08
[cms-v2] Make proposal token line item field into a dropdown

### LINKED ISSUE #1627: [cms-v2] Make proposal token line item field into a dropdown
The proposal token line item field should be using approved proposals to populate a dropdown for users to select from.

While testing may be a challenge (due to XSS requests being blocked or something), this should be done with relative ease by using the existing proposals api and a new selector.

amassarwi:
fixes #1627

### DIFF
--- a/src/lib/api.js
+++ b/src/lib/api.js
@@ -27,6 +27,7 @@ const STATUS_ERR = {
 };
 
 const apiBase = "/api";
+const apiBaseProposals = "/proposals-api";
 const getUrl = (path, version = "v1") => `${apiBase}/${version}${path}`;
 const getResponse = get("response");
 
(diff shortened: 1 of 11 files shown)


src/setupProxy.js
```diff
@@ -1,6 +1,15 @@
 const proxy = require("http-proxy-middleware");
 
 module.exports = function(app) {
+  app.use(
```
amassarwi: I needed to run another `politeiawww` instance without the `--mode=cmswww` flag on `4445` port  (i.e `politeiawww --listen=:4445`) & add this second API proxy to be able to call both APIs.
How is this gonna work on `testnet` & `mainnet` ? would it be possible to call `https://cms.decred.org/api/v1/proposals/tokeninventory` from cms gui without adding second proxy/api ?

src/lib/api.js
```diff
@@ -250,8 +251,10 @@ export const parseResponse = (response) =>
     return { response: json, csrfToken: response.headers.get("X-Csrf-Token") };
   });
 
-const GET = (path) =>
-  fetch(apiBase + path, { credentials: "include" }).then(parseResponse);
+const GET = (path, isMocked = false) =>
```
amassarwi: See other comment: https://github.com/decred/politeiagui/pull/1758/files#r376728238

src/setupProxy.js
tiagoalvesdulce: Honestly, I don't know. I have to take a closer look. Any clue @alexlyp ?

PR #1765, opened 2020-02-12
[cms-v2] Chore/Fix build and address issues with the fetching of proposals

fernandoabolafio:
This PR fixes the build for the cmsv2 branch and add some code improvements. There is one issue when populating the proposal tokens field to be figured out as we may need to fetch the approved proposals from either test-proposals.decred.org or proposals.decred.org instead of the cmswww API which doesn't provide that data.


fernandoabolafio:
The issue with the invoice form population will be a fix for https://github.com/decred/politeiagui/issues/1766.

amassarwi:
Looks good, shouldn't we update invoice edit component as well ? it calls the same hook, should be calling the new one as well

PR #1769, opened 2020-02-12
[cms-v2] Master sync

amassarwi:


amassarwi:
@thi4go 
@thi4go 
I needed to resolve conflicts while adding commits one by one, I committed the conflicts resolving code changes also for commit by commit instead once for all, which generated new hashes, I don't think it's an issue though, all 20 commits were added, and all changes were included.

thi4go:
@amassarwi makes sense mate, looks correct then.

Issue #1776, opened 2020-02-14
Update StartVote route to v2.


lukebp:
decred/politeia#1088 deprecates the v1 StartVote route and adds a v2 StartVote route.

See the [v2 api spec](https://github.com/decred/politeia/blob/master/politeiawww/api/www/v2/v2.go) for details on the changes and let me know if you have any questions.

PR #1780, opened 2020-02-15
Add start vote v2

tiagoalvesdulce:
We should update production with the dark theme changes before getting this PR in.

Closes https://github.com/decred/politeiagui/issues/1776


src/containers/Proposal/Actions/hooks.js
```diff
@@ -89,7 +89,8 @@ export function usePublicActions() {
         proposal.censorshiprecord.token,
         duration,
         quorumPercentage,
-        passPercentage
+        passPercentage,
+        proposal.version
```
amassarwi: can we destruct `version` & `token` as well ?

PR #1781, opened 2020-02-15
Comments `flat` mode

### LINKED ISSUE #1209: Add "NEW FLAT" comment display mode
Motivation: I need a way to quickly identify all new comments that I did not read, and to never miss a comment.

NEW and OLD sort modes change ordering of comments within their corresponding nesting level. For example, in the NEW mode recent top level comments move to top, but if a recent comment was nested under an older top level comment it will stay down below.

The only way I imagine to show all recent comments on top regardless of their nesting is to remove the nesting and show a flat list. This is altering not only the "sort mode" but also the "display mode", the latter in the title.

Flat list with removed tree structure has a downside of losing context for nested comments which are replies to other comments - can be addressed with "see in context" link button. The upside is that it doesn't require an API change and can be done on UI only (perhaps including server side rendering).

The idea originated in #1021 which has the same motivation but requires API changes:

> Ordering by "New" only helps to spot new top level comments. For nested replies it will only sort them under their parent which might be old. Therefore sorting the tree is not too helpful to quickly identify comments that I did not read.

> Right, so why can't the sorting be improved to make it possible to view all comments by "New" and not just top level comments.

amassarwi:
This PR adds comments `flat` display mode, as described in #1209 - in order to make it easier for users to read all new comments without the struggle caused by the comments nested/tree structure.

Closes #1209


src/containers/Comments/Comment/Comment.jsx
```diff
@@ -110,19 +114,18 @@ const Comment = ({
...
             </LoggedInContent>
           )}
-          {numOfReplies > 0 && (
+          {numOfReplies > 0 && !isFlatMode &&  (
             <span className={styles.showReplies} onClick={onClickShowReplies}>
               {showReplies ? "-" : `+${numOfReplies}`}
             </span>
           )}
-          {numOfNewHiddenReplies > 0 && !showReplies && (
+          {numOfNewHiddenReplies > 0 && !showReplies && !isFlatMode && (
```
tiagoalvesdulce: Can we improve this component to use more meaningful boolean variables? This one, for example, can be: `const showNewReplies: numOfNewHiddenReplies > 0 && !showReplies && !isFlatMode;`

src/containers/Comments/Comment/CommentWrapper.jsx
```diff
@@ -115,14 +117,19 @@ const CommentWrapper = ({ comment, children, numOfReplies, ...props }) => {
     [children]
   );
 
+  const contextLink = useMemo(
```
tiagoalvesdulce: This should be a Component. It can be defined in this file because it's only used here.

src/containers/Comments/Comment/CommentWrapper.jsx
```diff
@@ -115,14 +117,19 @@ const CommentWrapper = ({ comment, children, numOfReplies, ...props }) => {
...
+    () => isFlatMode && parentid > 0 &&  <Link className={styles.contextLink} to={`/${recordType}s/${recordToken}/comments/${parentid}`}>see in context</Link>,
+    [isFlatMode, parentid, recordToken, recordType]
+  );
+
   const commentContent = useMemo(
```
tiagoalvesdulce: This could be a component as well. Also, try to create meaningful boolean variables.

tiagoalvesdulce:
Looking pretty good. But some sorting options are some unresponsive. Can you take a look? Here it works just fine after a refresh but until a refresh is made the sort is messed up. See:

top

Here, "comentario c" should be on top of "comentario d".

amassarwi:
@tiagoalvesdulce Nice catch, I could reproduce it - the issue was that the sort/rank of comments wasn't updated on comment like event.
I've dispatched a sort action on like events, which should handle the case properly.

tiagoalvesdulce:
IMO the "re-sorting" on comment like isn't looking good. It's weird to vote on a comment and see it jumping to another place. Can you remove the dispatch in the `handleCommentLike` handler and keep it in the `handleCommentsModeToggle` handler?

amassarwi:
@tiagoalvesdulce agree, it's bad UX, changing it.

Issue #1764, opened 2020-02-11
Improve useAPIAction hook to memoize arguments


victorgcramos:
Every time we want to use the `useAPIAction` passing arguments to it, we must memoize the request's parameter values, like:

https://github.com/decred/politeiagui/blob/7f95223734fa8172955dab23eb57796d56e2a2fb/src/hooks/api/useExchangeRate.js#L27-L32

and

https://github.com/decred/politeiagui/blob/7f95223734fa8172955dab23eb57796d56e2a2fb/src/containers/Invoice/Detail/hooks.js#L14-L19

...

The idea is to avoid this behaviour, and memoize those values inside the `useAPIAction` hook.

victorgcramos:
Thanks to @tiagoalvesdulce's observation on #1751.

fernandoabolafio:
Unfortunately, there isn't an optimal way of doing this because we will always be specifying an object as a parameter that will always fail in the shallow comparison. The only way it would work is could work is if the parameter is a non-object variable (e.g string, number, boolean). That said, I think what we can do to avoid a mistake is to discard the use of `useAPIAction` and in every component/hook we want an action we will call the dispatch and create the callback (if needed):

```jsx
const dispatch = useDispatch();

// either do
dispatch(someAction());

// or
const callback = useCallback(() => dispatch(someAction(params)), [params])

```

In terms of verbosity I think both approaches are equal but the one I presented will prevent mistakes of not memorizing the parameters thus leading to unwanted re-renders.

tiagoalvesdulce:
Your approach is OK. My concern about the current approach is that it's not obvious for newcomers that they have to memoize the `useAPIAction()` params before passing them to the hook in order to prevent infinite loops. We can either use your approach or create our custom memoization utility with deep comparison and make it clear to not use it with deeply nested objects.

fernandoabolafio:
@tiagoalvesdulce my fear is about the 'make clear' part of it. I would rather use an approach that prevents errors by itself. If we explicitly use the dispatch to fire actions we mitigate those risks. It's also weird that we need to enforce the memoization  of function arguments. 
Can we enforce what has to be the function argument for the function (like typescript) using something else like flow or just plain js? I am just thinking that would be fine to use `useAPIAction` as long as the parameters are not objects.

fernandoabolafio:
I was taking a look at this discussion: https://github.com/facebook/react/issues/14476 and Dan Abramov actually suggests something similar to what we're doing:

[...]

PR #1782, opened 2020-02-17
Proposal: refactor useAPIAction hook

tiagoalvesdulce:
This PR is my proposal to improve the useAPIAction hook as discussed in https://github.com/decred/politeiagui/issues/1764.

* Remove semantic usage of useMemo
* Uses useRef to create a cache
* Endorse hook users to pass an array of primitives to the args argument of useAPIAction

Closes https://github.com/decred/politeiagui/issues/1764

### DIFF
--- a/src/hooks/utils/useAPIAction.js
+++ b/src/hooks/utils/useAPIAction.js
@@ -1,8 +1,8 @@
-import { useEffect, useState } from "react";
+import { useRef, useEffect, useState } from "react";
 
 const DEFAULT_ARGS = [];
 
-function useAPIAction(action, args = DEFAULT_ARGS, enabled = true) {
+function useApplyAction(action, args = DEFAULT_ARGS, enabled = true) {
   const [loading, setLoading] = useState(false);
   const [error, setError] = useState(false);
   const [fired, setFired] = useState(false);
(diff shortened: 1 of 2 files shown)


src/hooks/utils/useAPIAction.js
```diff
@@ -25,4 +25,31 @@ function useAPIAction(action, args = DEFAULT_ARGS, enabled = true) {
...
+ * useAPIAction hook memoizes the args array and applies the action if enabled. args should be an array of primitives.
+ * @param {function} action
+ * @param {array} args
+ * @param {boolean} enabled
+ */
+function useAPIAction(action, args = DEFAULT_ARGS, enabled = true) {
+  const actionArgs = useRef(args);
+  const areArgsMatched =
+    args && actionArgs.current && areArgsEqual(args, actionArgs.current);
+  const cached = areArgsMatched ? actionArgs.current : args;
```
amassarwi:
```suggestion
  const newArgs = areArgsMatched ? actionArgs.current : args;
```

tiagoalvesdulce: hmm, I don't quite agree with `newArgs` because I want it to be clear that it references the same object if the args are matched and `new` is a tricky word for that. I can go with `cachedArgs` tho.

tiagoalvesdulce:
By results do you mean the return of the action?

tiagoalvesdulce:
Discussed with @fernandoabolafio on matrix and we decided to not return the result of the action to keep redux as the single source of truth.

PR #1788, opened 2020-02-21
[cms-v2][dcc-v2] DCC support/oppose action

### LINKED ISSUE #1689: [cms-v2] DCC Support/oppose
There should be just 2 buttons  for **support** and for **oppose**. The buttons behaviour is the same as v1.

V1 Reference issue: #1540

victorgcramos:
This diff allows users to support/oppose active dccs.

Closes #1689


amassarwi:
Also, let's align the actions styling with invoice actions - align to right and secondary action with no border, see below:
image

Issue #1733, opened 2020-01-29
Improve paywall error message


lukebp:
<img width="1327" alt="Screen Shot 2020-01-28 at 9 10 04 PM" src="https://user-images.githubusercontent.com/9039877/73327335-c3cc1d80-4212-11ea-86a6-dbf33438b3b6.png">

The paywall error message is lacking.  It doesn't tell you how to pay the paywall or provide a link to the user's credits page.

It should be more like the missing identity error message.

PR #1797, opened 2020-02-22
Improve paywall error message

amassarwi:
depends on: https://github.com/decred/pi-ui/pull/225
closes: #1733


amassarwi:
decred/pi-ui#225 isn't really needed - using `P` component as `Message` child does the job without css, updating the PR!

Issue #1202, opened 2019-05-14
[cms] Submit Invoice UI suggestions


xaur:
* name/location/... field labels are too small, make them same size and font like other text
* remove italics from field descriptions (they are already bold)
* rename "Add/Remove Row" to "Add/Remove Item" for terminology consistency with "line item"
* make "Add Row" same green color as "Attach a file" button (green is often used to signal "add")
* add instructions how to obtain proposal token
* explain which characters are allowed and not allwed in fields
* make Location field optional
* show uneditable column that computes USD value for line items
* show Total USD value
* show "4 - April" or "(4) April" instead of just "4" for month
* Description cell is too small for "thorough description", which implies multiline text
* add note that line items can be copied and pasted from LibreOffice, given enough rows in the invoice
* vertically align cell text to Top

amassarwi:
@xaur I believe all were addressed in the redesign work, found tiny one that we still could add:

> rename "Add/Remove Row" to "Add/Remove Item" for terminology consistency with "line item"

PR #1803, opened 2020-02-29
[cms-v2] Invoice small improvements

amassarwi:
The cms redesign work already addressed most of #1202, this diff adds the following missing two bits: 
- Make location field optional
-  Small wording change

Waiting on https://github.com/decred/politeia/pull/1142

### DIFF
--- a/src/componentsv2/InvoiceForm/validation.js
+++ b/src/componentsv2/InvoiceForm/validation.js
@@ -24,7 +24,6 @@ export const invoiceValidationSchema = ({
       .max(maxnamelength)
       .matches(...yupFieldMatcher("Name", cmsnamelocationsupportedchars)),
     location: Yup.string()
-      .required("required")
       .min(minlocationlength)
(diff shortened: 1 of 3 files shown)


src/componentsv2/InvoiceForm/validation.js
```diff
@@ -24,7 +24,6 @@ export const invoiceValidationSchema = ({
       .max(maxnamelength)
       .matches(...yupFieldMatcher("Name", cmsnamelocationsupportedchars)),
     location: Yup.string()
-      .required("required")
```
fernandoabolafio: IIRC the 'location' field is required in the API so we should not make it optional in the frontend unless we change the API. Please, verify that before moving this in.

amassarwi: @fernandoabolafio 
I've changed the api and linked the BE PR in the description of this PR.

PR #1804, opened 2020-02-29
[cms-v2] Sync master

fernandoabolafio:


fernandoabolafio:
@amassarwi thanks for catching that. I will fix it.
@victorgcramos we should not keep it. I will re-do the rebase b/c of amass comment and you can take another look.

victorgcramos:
LGTM

PR #1805, opened 2020-03-02
[cms-v2] Add go back link for invoice/dcc details page

### LINKED ISSUE #1802: [cms-v2] Add `back` link to invoices & DCCs details pages
Currently the only way to navigate back to list page in both DCC/invoice details views is by clicking the browser back button, we could add small `back` link like the one we have on proposal's details page.

thi4go:
closes #1802

### DIFF
--- a/src/componentsv2/Router/GoBackLink.jsx
+++ b/src/componentsv2/Router/GoBackLink.jsx
@@ -8,9 +8,6 @@ const GoBackLink = () => {
   const isDarkTheme = themeName === "dark";
   const { pastLocations, history } = useRouter();
   const previousLocation = pastLocations[1];
-  const returnToPreviousLocation = useCallback(() => history.goBack(), [
-    history
-  ]);
 
   const goBackLinkFromPreviousLocation = useMemo(() => {
     if (!previousLocation) return null;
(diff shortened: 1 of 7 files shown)


src/componentsv2/Router/GoBackLink.jsx
```diff
@@ -0,0 +1,35 @@
...
+import { Link, classNames, useTheme } from "pi-ui";
+import { useRouter } from "src/componentsv2/Router";
+import styles from "./GoBackLink.module.css";
+
+const GoBackLink = () => {
+  const { themeName } = useTheme();
+  const isDarkTheme = themeName === "dark";
+  const { pastLocations, history } = useRouter();
+  const previousLocation = pastLocations[1];
+  const returnToPreviousLocation = useCallback(() => history.goBack(), [
```
tiagoalvesdulce: This `useCallback` hook is unnecessary 'cause history is always changing and we are not passing `returnToPreviousLocation` down to the component tree so the memoization is doing nothing.

thi4go: Makes sense mate, thanks for the feedback. Corrected this.

PR #1806, opened 2020-03-03
[cms-v2][dcc-v2] DCC approve/reject actions

### LINKED ISSUE #1690: [cms-v2] DCC admin Approve/Reject
> Administrators should be shown separate buttons in the listings and dcc details page for Approve/Reject.

V1 issue reference: #1541

victorgcramos:
This commit adds the feature to approve or reject DCCs.

DCCs can be approved or rejected by administrators with a given reason.

Closes #1690

### DIFF
--- a/src/containers/DCC/Actions/Actions.jsx
+++ b/src/containers/DCC/Actions/Actions.jsx
@@ -1,0 +1,62 @@
+import React, { useCallback } from "react";
+import {
+  Button,
+  classNames,
+  useMediaQuery,
+  Dropdown,
+  DropdownItem
+} from "pi-ui";
+import { useAdminDccActions } from "./hooks";
+import AdminContent from "src/componentsv2/AdminContent";
+import { isDccActive } from "../helpers";
+
+const DccActions = ({ dcc, extended }) => {
+  if (!useAdminDccActions()) {
+    throw Error(
+      "Admin dccs actions requires an 'AdminActionsProvider' on a higher level of the component tree. "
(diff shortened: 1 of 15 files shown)


src/containers/DCC/Actions/Actions.jsx
```diff
@@ -0,0 +1,62 @@
...
+      onClick={withDcc(onReject)}
+      className={classNames("margin-right-s")}
+      noBorder
+      kind="secondary"
+    >
+      Reject
+    </Button>
+  );
+
+  return isDccActive(dcc) && (
```
tiagoalvesdulce: If the DCC is not active, you should return null

src/containers/DCC/Actions/AdminActionsProvider.jsx
```diff
@@ -0,0 +1,67 @@
...
+  ] = useBooleanState(false);
+  const [
+    showApproveModal,
+    openApproveModal,
+    closeApproveModal
+  ] = useBooleanState(false);
+
+  // set the dcc target before executing the function
+  const withDccTarget = (fn) => (dcc) => {
+    setTargetDcc(dcc);
```
tiagoalvesdulce: Well, `setTargetDcc` is an async method so you can't guarantee that the state will be updated with this code. Take a look at [this blogpost](https://sung.codes/blog/2018/12/07/setting-react-hooks-states-in-a-sync-like-manner/) for a better explanation.

PR #1815, opened 2020-03-07
Use pi-ui's select

amassarwi:
We've migrated `componentsv2/Select` to `pi-ui` code base, see: https://github.com/decred/pi-ui/pull/227
This diff delete the select component and uses the one provided by `pi-ui`


victorgcramos:
Hey! Looking good! I think the select component on `/user/<id>?tab=proposals` is missing the dropdown arrow... See:
Screen Shot 2020-03-13 at 13 13 52

PR #1820, opened 2020-03-15
[rfp] New proposal flow changes

### LINKED ISSUE #1798: [rfp] Adjust new proposal flow
According to the [specs](https://xd.adobe.com/view/44a3298e-a814-42b2-643e-3aa39c8dec34-3292/) 3 types of new proposals should be available:

- [x] Regular proposal
- [x] RFP
- [x] RFP submission 

The changes should also include:
- [x] Adjust edit proposal flow
- [x] Ensure drafts include new fields

amassarwi:
[specs](https://xd.adobe.com/view/44a3298e-a814-42b2-643e-3aa39c8dec34-3292/screen/73b5dadd-d93d-4f8f-b7b9-74c4ae8dd567/Create-RFP-proposal-/)

~~*State:* 
According to provided specs the RFP proposal need a datepicker, @fernandoabolafio has added lately a datepicker which supports month/year format dates with the cms redesign work - on `cmsv2` branch.
I'll be working on expanding it to support days as well, after that we could rebase this branch hopefully against master to sync all `cmsv2` work and continue from there.~~

Open tasks:
- [x] make rfp deadline work with no initial value - https://github.com/decred/pi-ui/issues/242
- [x] adjust edit proposal flow
- [x] add *conditioned* `required` validation for new fileds
- [x] ensure new fields are saved in drafts
- [x] align input fields border colors
- [x] use timestampes provided in policy as min, max for RFPs deadline ~~(waiting on: https://github.com/decred/pi-ui/issues/248)~~

closes #1798
~~depends on : decred/pi-ui#247 & decred/pi-ui#249~~
depends on https://github.com/decred/politeia/pull/1054


amassarwi:
@tiagoalvesdulce @fernandoabolafio 
After short research I've found that we are using https://github.com/nickeljew/react-month-picker on `cmsv2`.
It looks like it doesn't support days and it specific for picking a month.
I guess we will need to find other alternatives, found this: https://react-day-picker.js.org/api/DayPickerInput

What do you think ?

src/actions/api.js
```diff
@@ -353,7 +353,9 @@ export const onFetchProposalsBatch = (tokens, fetchVoteStatus = true) =>
       }
       const response = await Promise.all(promises);
       const proposals = response.find((res) => res && res.proposals).proposals;
+      const summaries = response.find((res) => res && res.summaries).summaries;
       dispatch(act.RECEIVE_PROPOSALS_BATCH({ proposals }));
+      return [proposals, summaries];
```
amassarwi, line 358: @tiagoalvesdulce 
I added a return statement in order to await the `onFetchProposalsBatch` function call and use the fetched proposal batch & vote summaries in proposal form submit and then validate if the provided token is valid (see `ProposalForm` component).
I see we don't do that in other actions so maybe I shouldn't do that ?

`useProposalsBatch` hook is returning the proposal list but only in the next render and in my case and need to resolve them to decide whether the submitted forms is valid or not.

What do you think would be the proper way to approach it?

tiagoalvesdulce: The way you are doing is fine because it only uses the data at the moment it's returned. You're not breaking the single source of truth by passing it down to other components, etc. Good job!

amassarwi:
@lukebp 
 1. Agreed, should it be an icon next to the textbox ?
 2. Right, it's currently depends on decred/pi-ui#247 & decred/pi-ui#249, one they're in calendar shouldn't be broken.

src/components/ProposalForm/ProposalForm.jsx
```diff
@@ -168,10 +251,29 @@ const ProposalFormWrapper = ({
...
+        const { type, rfpLink, ...others } = values;
+        if (type === PROPOSAL_TYPE_RFP_SUBMISSION) {
+          const [[proposal], summaries] = (await onFetchProposalsBatch([
+            rfpLink
+          ])) || [[], null];
+          const voteSummary = summaries && summaries[rfpLink];
+          if (
+            !proposal ||
+            !voteSummary ||
+            !isActiveApprovedRFP(proposal, voteSummary)
```
tiagoalvesdulce, line 267: Can you create a meaningful variable for this logic expression?

src/actions/tests/api.test.js
```diff
@@ -458,9 +464,6 @@ describe("test api actions (actions/api.js)", () => {
   test("on fetch liked comments action", async () => {
     const path = `path:/api/v1/user/proposals/${FAKE_PROPOSAL_TOKEN}/commentslikes`;
     const params = [FAKE_PROPOSAL_TOKEN, FAKE_USER.id];
-    // fetchMock.get(path, {}, { query: {
-    //   userid: FAKE_USER.id
-    // } })
```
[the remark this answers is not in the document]
amassarwi: what do you mean ? just deleted some commented out code

MariaPleshkova:
@amassarwi @lukebp yeah, I think we can just add info icon right next to deadline field.
Like this:
image
Added it to the specs.

amassarwi:
@lukebp @MariaPleshkova 
image

MariaPleshkova:
@amassarwi looks good to me

PR #1823, opened 2020-03-17
Add loading indicator for like comment action

### LINKED ISSUE #1792: Don't render cast comment votes until server response is received
A few server lags when casting comment likes prompted me to open the dev tools and notice that comment likes are rendered immediately and speculatively before receiving the confirmation from the server. Then if the request times out or errors, the changed comment score is reverted and the pressed like/dislike buttons are returned to unpressed state. Unexpectedly, this state reversion is also followed by a few redrawings of the comment areas.

This speculative behavior is undesirable because it fools the user to believe his comment votes were cast sooner than it really happens on the server. If I cast a few votes, see that they are immediately applied and close the tab while there is a connectivity or server issue, the votes will not be applied and I will never know about it.

Another undesirable effect is that this behavior hides the fact that requests are processed slowly.

thi4go:
This PR adds a visual indicator that the like comment request is still in progress, so that the user is sure his vote was casted properly.

closes #1792

### DIFF
--- a/src/componentsv2/Likes/Likes.jsx
+++ b/src/componentsv2/Likes/Likes.jsx
@@ -1,4 +1,4 @@
-import React, { useState, useCallback } from "react";
+import React, { useCallback } from "react";
 import PropTypes from "prop-types";
 import {
   Icon,
@@ -6,6 +6,7 @@ import {
   Text,
   getThemeProperty,
   useHover,
+  Spinner,
   classNames
 } from "pi-ui";
 import styles from "./Likes.module.css";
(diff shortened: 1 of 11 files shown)


thi4go:
Nice suggestion @tiagoalvesdulce, will fix this mate 👍

src/componentsv2/Likes/Likes.jsx
```diff
@@ -26,20 +27,30 @@ const Likes = ({ upLikes, downLikes, onLike, onDislike, option, disabled, loadin
   const dislikeColor =
     disliked || isDislikeHovered ? activeColor : defaultColor;
 
+  useEffect(() => {
+    if (apiLoading) {
+      setTimeout(() => setLoading(true), 1000);
```
tiagoalvesdulce: Can we use [lodash's delay](https://lodash.com/docs/4.17.15#delay) method? It's available on lodash fp and it's more declarative than a setTimeout.

thi4go: Done mate @tiagoalvesdulce

tiagoalvesdulce:
Doing a more detailed review/test I noticed the spinner shows even after the feedback is shown. See (to test use the Slow 3g network mode on chrome):

gifbugloading

The flow we want is:

1. The user clicks the vote button.
2. Does it return in less than ~ 1 sec? If yes, do nothing. If no, show the spinner until it gets the response from the API.

thi4go:
@tiagoalvesdulce nice catch mate. Our comment reducer was adding the count before we received confirmation that the request went through correctly.

Issue #1178, opened 2019-05-09
Fetch token inventory silently


fernandoabolafio:
We need to display the loading indicator while fetching the token inventory just in the first time. The subsequent fetchings need to be done silently. That way, we don't lose the list state (especially scroll position) when switching to other routes. 

**Edit:** We can even display a discrete indicator which does not disrupt the list view.

PR #1824, opened 2020-03-17
Fetch token inventory silently after the first time

tiagoalvesdulce:
Depends on https://github.com/decred/politeia/pull/1149.

Closes https://github.com/decred/politeiagui/issues/1178


fernandoabolafio:
@tiagoalvesdulce I am just curious why you needed that change in the politeia repo to the silent fetching?

tiagoalvesdulce:
Just to ease things out. I had to compare the tokenInventory response with the allByStatus state property. The backend was returning null for unreviewed and censored records when they were empty. Returning empty arrays instead made everything simpler.

Issue #1153, opened 2019-05-02
[cms] Preload invoice information


alexlyp:
From @matheusd on matrix:

> Going through the menu link "Submit New Invoice" should ideally pre-load the data for billing the last month. It would be helpful to also be able to pre-specify the name, location, etc

We could probably fetch the most recent invoice and then populate the corresponding fields.  Not sure if we would need an extra endpoint to accomplish this, maybe we could sort the already returned user invoices.

thi4go:
I'd like to work on this

alexlyp:
I'd hold off on this in lieu of the upcoming changes/additions of user information piwww.  We should then just be able to request the user information directly to be put in there.

xaur:
As a more general form of this feature, consider "Copy from" feature that allows to initialize the invoice form from an arbitrary past invoice.

PR #1828, opened 2020-03-20
[cmsv2] CMS redesign work

amassarwi:
This diff adds all CMS redesign work.
I've done the following in order to sync the branches:
 1. Merged master into cmsv2 to resolve all conflicts
 2. Updates react-script to fix build issue
 3. I've deleted `react-select` dep - it's now part of `pi-ui`
 4. Finally deleted  renamed `componentsv2` to `components` & removed old folder content

**Resolved - see update in comment below**
~~For some reason build is failing currently with the following error:~~
```
Failed to compile.

./node_modules/react-dev-utils/webpackHotDevClient.js
Error: [BABEL] /home/amass/src/politeiagui/node_modules/react-dev-utils/webpackHotDevClient.js: Could not find plugin "proposal-numeric-separator". Ensure there is an entry in ./available-plugins.js for it. (While processing: "/home/amass/src/politeiagui/node_modules/babel-preset-react-app/dependencies.js$0")
    at Array.map (<anonymous>)
    at Generator.next (<anonymous>)
    at Generator.next (<anonymous>)
```

~~**Update:** It seems like the build error is related to: https://github.com/facebook/create-react-app/issues/8680 - I've tried the suggested workaround here https://github.com/facebook/create-react-app/issues/8680#issuecomment-601851525 but it didn't resolve the issue.~~

PR #1825, opened 2020-03-18
Use pi-ui's datepicker

amassarwi:
I've added a datepicker component to `pi-ui`'s code base - see: https://github.com/decred/pi-ui/pull/232

> The new datepicker component supports either days(e.g rfp  deadline) or months(e.g cms invoice forms) mode

waiting on: https://github.com/decred/pi-ui/pull/232
part of: #1820
closes #1826

PR #1829, opened 2020-03-22
Preload user info on invoice submit page

tiagoalvesdulce:
I have some questions regarding this PR:

1. Shouldn't we set the contractor rate?
2. We have `githubname` and `matrixname` on the API but there are no fields to set them. Should I add them?
3. Aren't `matrixname` and `contractorcontact` going to be almost always the same?

@alexlyp 

Closes https://github.com/decred/politeiagui/issues/1153


tiagoalvesdulce:
I didn't _need_ to upgrade them. It's just a simple upgrade and I feel that another PR just for doing that would be unnecessary.

amassarwi:
@tiagoalvesdulce I see, my only concern here is conflicts nothing more, I already did deps change on #1828 and I though we may want to avoid the merge conflicts, it shouldn't be *real* issue I guess.

tiagoalvesdulce:
We can drop the yarn.lock file and do a fresh yarn to generate a new one if the conflicts are an issue.

amassarwi:
@tiagoalvesdulce true, just did it with merging #1825 into #1828

PR #1841, opened 2020-03-28
Legacy code cleanup

amassarwi:


amassarwi:
I've found out that the endless loading in account's proposals list was still happening for new users with no proposals, this commit here https://github.com/decred/politeiagui/pull/1828/commits/ea28c461167b1412d707ae24b7ef4e29fff263ea solved the issue only for users with 1 proposal at least.

I've investigated the issue further and found that our condition to determine whether the fetch is done isn't good enough. I changed it to use the redux `user num of proposals`  (on the way fixed it's selector) - see https://github.com/decred/politeiagui/pull/1841/commits/d366011a1025ecb17615fa4873099b6dd33c6328.

src/containers/Proposal/User/Submitted.jsx
```diff
@@ -45,7 +45,7 @@ const Proposals = (props) => {
   }
 
   const numOfProsalsLoaded = proposals.length;
-  const initialFetchDone = !!proposals.length;
+  const initialFetchDone = numOfProsalsLoaded === numOfUserProposals;
```
tiagoalvesdulce: Same typo

src/containers/Proposal/User/Submitted.jsx
```diff
@@ -45,7 +45,7 @@ const Proposals = (props) => {
   }
 
   const numOfProsalsLoaded = proposals.length;
```
tiagoalvesdulce: There's a typo here. Should be `numOfProposalsLoaded`

tiagoalvesdulce:
For some reason, the order of my comments above got messed up. Everything else looks good. Good job!

amassarwi:
@tiagoalvesdulce thanks mate, updated!

PR #1844, opened 2020-03-30
Cache normalized state for paywall data

### LINKED ISSUE #1490: Redux state normalization

thi4go:
This is the last PR of the effort to normalize our state data in its own branch, separated by models. It normalizes and caches paywall data by users.

Besides normalizing the state, this PR fixes a infinite loop that crashed the app when trying to open the credit payment modal for the second time, after a first buy was confirmed. It also adds minor improvements to the flow of the payment process.

closes #1490

### DIFF
--- a/src/containers/User/Detail/Credits/hooks.js
+++ b/src/containers/User/Detail/Credits/hooks.js
@@ -136,7 +136,7 @@ export function useCredits(userID) {
   ]);
 
   useEffect(() => {
-    if (proposalPaywallPaymentTxid === undefined) {
+    if (!proposalPaywallPaymentTxid) {
       onFetchProposalPaywallPayment();
     }
   }, [proposalPaywallPaymentTxid, onFetchProposalPaywallPayment]);
--- a/src/containers/User/Detail/Credits/hooks.js
+++ b/src/containers/User/Detail/Credits/hooks.js
@@ -88,7 +88,7 @@ export function useCredits(userID) {
   const clearProposalPaymentPollingPointer = useAction(
     act.clearProposalPaymentPollingPointer
   );
-  const { isPaid } = usePaywall();
+  const { isPaid } = usePaywall(userID);
   const proposalCredits = proposalCreditsUnspent.length;
   const proposalCreditsFetched = proposalCredits !== null;
   const isUserPageOwner = currentUserID === userID;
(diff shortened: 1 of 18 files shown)


thi4go:
Nice catch mate @victorgcramos, missed that one.

src/containers/User/Detail/Credits/hooks.js
```diff
@@ -94,17 +122,24 @@ export function useCredits(userID) {
...
   }, [
     pollingCreditsPayment,
+    proposalPaywallPaymentTxid,
     toggleCreditsPaymentPolling,
-    proposalPaywallPaymentTxid
+    onPollProposalPaywallPayment
   ]);
 
+  useEffect(() => {
+    if (proposalPaywallPaymentTxid === undefined) {
```
amassarwi, line 135:
```suggestion
    if (proposalPaywallPaymentTxid proposalPaywallPaymentTxid === "undefined") {
```

tiagoalvesdulce: why not just `if(!proposalPaywallPaymentTxid)` ?

amassarwi: true, i thought we are checking the undefined case here explicitly

thi4go:
Fixed mates @tiagoalvesdulce @amassarwi thanks for the review :+1:

PR #1848, opened 2020-04-03
Adjust icon component usage to use `width` prop instead of `size`

amassarwi:


amassarwi:
decided to keep the size number value to prevent specifying width, height values everywhere!

PR #1850, opened 2020-04-06
[cms] Replace BoxTextInput by TextArea on DCC Statement

victorgcramos:
This issue depends on https://github.com/decred/politeia/pull/1164.

It should improve the DCC usability, since it is a critical feature. Now the DCC sponsor can format her/his using the same Markdown formatting on proposals.

### DIFF
--- a/src/components/DCC/DCC.jsx
+++ b/src/components/DCC/DCC.jsx
@@ -19,7 +19,6 @@ import {
   isDccApproved
 } from "src/containers/DCC/helpers";
 import { SupportOppose, DccActions } from "src/containers/DCC/Actions";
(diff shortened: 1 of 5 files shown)


alexlyp:
Works great!  Only nit I have is that we should also have a markdown/large text display for the static dcc details page as well.

victorgcramos:
The markdown rendering was not possible due to some char limitations such as `*`, which was limiting the user experience regarding the Markdown usability. I'm now switching the Markdown view  to the TextArea implemented on https://github.com/decred/pi-ui/pull/246. This PR is closed and I'll open another one.

PR #1851, opened 2020-04-07
[cms] Replace BoxTextInput by TextArea on DCC Statement

victorgcramos:
- TextArea to DCC Form
- Multi-line view to DCC Details

This fixes the markdown experience issue reported on https://github.com/decred/politeiagui/pull/1850#issuecomment-610152199

Depends on https://github.com/decred/pi-ui/pull/246 and https://github.com/decred/politeia/pull/1165

### DIFF
--- a/src/components/DCC/DCC.jsx
+++ b/src/components/DCC/DCC.jsx
@@ -113,7 +113,7 @@ const Dcc = ({ dcc, extended }) => {
                 <Row justify="space-between" className={styles.topDetails}>
                   <div className={styles.field}>
                     <Text size="small">Statement</Text>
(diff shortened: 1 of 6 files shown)


tiagoalvesdulce:
Weren't we going with the MarkdownEditor for DCC Statements? Or is this a temporary fix?

victorgcramos:
@tiagoalvesdulce the markdown experience would be limited since not all markdown-key characters, such as `*` used to make **bold texts**, for example, are allowed by our policy. I do think we should improve the policy to accept that, but for now, I think that this should satisfy our needs.

PR #1857, opened 2020-04-14
[pi/cms] Refactor on legacy redux state tree

thi4go:
This is the final PR towards the refactor proposed on https://github.com/decred/politeiagui/issues/1490 to our state tree. This diff:

- keeps the `api` state branch with the sole purpose of returning `isApiRequesting` and `error`, while the data is kept under our `models/` tree, respecting the new single source of truth. 

- deletes a lot of unused legacy code throughout the redux boilerplate (selectors, reducers, actions...), resulting in more comprehensive files, which contains code that is actually being used throughout the app.

- more manageable smaller modules for the state tree.

It's important to note that this PR doesn't add any new features, but touches a lot of the pre-existing ones. Mainly updating hooks to fetch data from the correct single source of truth location.

This PR is built on top of #1844. When it gets merged, I'll rebase this PR against master to facilitate the review.


src/selectors/models/paywall.js
```diff
@@ -0,0 +1,51 @@
...
+  }
+  return paywall.status || PAYWALL_STATUS_WAITING;
+};
+
+const createPaywallSelector = (userID, key) =>
+  createSelector(makeGetPaywallByUserID(userID), (paywall) =>
+    paywall ? paywall[key] : null
+  );
+
+export const makeGetUserIsPaid = (userID) =>
```
amassarwi, line 32: :rocket:

thi4go:
@tiagoalvesdulce the identity issue is happenning on master as well; it's due to testnet dcrdata being down. I'll fix the rescan credits message and commit it.

Thanks for the review mate.

src/containers/User/Detail/Credits/components/RescanSection.jsx
```diff
@@ -1,9 +1,10 @@
 import React from "react";
 import { Message } from "pi-ui";
 
-export default ({ amountOfCreditsAddedOnRescan, errorRescan }) =>
+export default ({ amountOfCreditsAddedOnRescan, errorRescan }) => 
   <>
-    {amountOfCreditsAddedOnRescan !== null && (
+    {amountOfCreditsAddedOnRescan !== null &&
+     amountOfCreditsAddedOnRescan !== undefined && (
```
tiagoalvesdulce, line 7: Why does this check solve the issue? In the credits reducer, the initialState of `newcredits` is `null` and the fallback for `makeGetCreditsAddedOnRescan` is also `null`. Is there any situation where it can be undefined? If there is one, we should solve that instead and fallback to something that makes sense.

thi4go: Last commit fixes this mate, fallback to `newcredits` now will always be `null` :+1: thanks for pointing this out

PR #1867, opened 2020-04-16
[cms] Fix invoice lineitems selector

### LINKED ISSUE #1866: [cms] Lineitem's dropdown fields are broken
Currently when clicking on of the dopdown items the value doesn't change!

victorgcramos:
Fixes #1866

### DIFF
--- a/src/components/InvoiceDatasheet/components/SelectEditor.jsx
+++ b/src/components/InvoiceDatasheet/components/SelectEditor.jsx
@@ -14,7 +14,7 @@ const customStyles = {
 };
 
 const SelectEditor = ({ value, options, onChange, onCommit }) => {
-  const [newValue, setNewValue] = useAsyncState({ value });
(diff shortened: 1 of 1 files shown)


src/components/InvoiceDatasheet/components/SelectEditor.jsx
```diff
@@ -12,15 +13,23 @@ const customStyles = {
   })
 };
 
-const SelectEditor = ({ value, options, onCommit }) => {
-  const handleChange = ({ value }) => {
+const SelectEditor = ({ value, options, onChange, onCommit }) => {
+  const [newValue, setNewValue] = useAsyncState({ value });
```
amassarwi, line 17: Maybe we could use `getValueObj` to get initial value object ?
```suggestion
  const [newValue, setNewValue] = useAsyncState(getValueObj(value));
```

PR #1873, opened 2020-04-21
[cms] Improve Invoice Filters

### LINKED ISSUE #1862: [cms] Improve invoice list
Right now it is only possible to show invoices monthly or all of them, but tipically, a contractor send a monthly invoice, ending with a "list" of one invoice/month 

image

It would be good having an anual option to show them as well

victorgcramos:
This diff adds `all` options on invoice months filter, so it displays all invoices from the given year.

Screen Shot 2020-04-20 at 22 19 51

Closes #1862

### DIFF
--- a/src/containers/Invoice/helpers.js
+++ b/src/containers/Invoice/helpers.js
@@ -103,31 +103,36 @@ export const getInvoiceTotalAmount = (invoice) => {
   return rate * totalHours + totalExpenses;
 };
 
-export const sortDateRange = (initialRangeValue, finalRangeValue) => {
-  if (isEqual(initialRangeValue, finalRangeValue)) return initialRangeValue;
+export const sortDateRange = (firstSelectedDate, secondSelectedDate) => {
+  if (isEqual(firstSelectedDate, secondSelectedDate)) return firstSelectedDate;
 
-  const { month: initialMonth, year: initialYear } = initialRangeValue;
-  const { month: finalMonth, year: finalYear } = finalRangeValue;
-  const initialDate = new Date(initialYear, initialMonth).getTime();
-  const finalDate = new Date(finalYear, finalMonth).getTime();
+  const { month: firstMonth, year: firstYear } = firstSelectedDate;
+  const { month: secondMonth, year: secondYear } = secondSelectedDate;
 
-  if (initialDate > finalDate) {
+  const firstSelectedTime = new Date(firstYear, firstMonth).getTime();
+  const secondSelectedTime = new Date(secondYear, secondMonth).getTime();
+
+  if (firstSelectedTime > secondSelectedTime) {
     return {
-      start: finalRangeValue,
-      end: initialRangeValue
+      start: secondSelectedDate,
+      end: firstSelectedDate
     };
   }
 
   return {
-    start: initialRangeValue,
-    end: finalRangeValue
+    start: firstSelectedDate,
+    end: secondSelectedDate
   };
 };
 
 export const getPreviousMonthsRange = (range) => {
   const end = getCurrentDefaultMonthAndYear();
+
+  // Tweak from Date API. If month is negative, it returns the last month of the
+  //   previous year. See: https://www.w3schools.com/jsref/jsref_setmonth.asp
   const previous = new Date();
   previous.setMonth(previous.getMonth() - range + 1);
+
   return {
     start: getYearAndMonthFromDate(previous),
     end
--- a/src/containers/Invoice/helpers.js
+++ b/src/containers/Invoice/helpers.js
@@ -1,8 +1,11 @@
 import {
   getCurrentMonth,
   getCurrentYear,
-  getCurrentDateValue
+  getCurrentDateValue,
+  getYearAndMonthFromDate,
+  getCurrentDefaultMonthAndYear
 } from "src/helpers";
+import isEqual from "lodash/isEqual";
 import { INVOICE_STATUS_NEW, INVOICE_STATUS_UPDATED } from "./constants";
 
 /**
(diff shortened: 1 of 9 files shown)


amassarwi:
@victorgcramos thanks for the PR mate.
I suggest we go differenet approach here, I think we should provide the contractors a way to pick a range, it's a much more flexible option to go with IMO.
Actually the datepicker already supports `range` prop, it's not used anywhere and not mentioned in the docz as well.
Not sure if BE route support providing a range, but for the long run it's much more solid approach,  what do you think ? maybe @tiagoalvesdulce can share his thoughts about that as well.

victorgcramos:
I like the range approach. I'll add this too. My idea here is to provide a 1 click action that shows all invoices for each year..

tiagoalvesdulce:
The [user invoices endpoint](https://github.com/decred/politeia/blob/master/politeiawww/api/cms/v1/api.md#user-invoices) doesn't accept any parameters so we can do invoice filtering as we need in the FE. I like the range option mainly because you can overflow years e.g from Sep 2019 to Feb 2020. We should create some defaults as well. I suggest three: Past 3 months, past 6 months and whole year (already implemented by @victorgcramos in this PR).

amassarwi:
@tiagoalvesdulce @victorgcramos Like to combination approach 🚀

victorgcramos:
Nice suggestions, guys! Good job 🦾🤙🏼

thi4go:
@tiagoalvesdulce @victorgcramos @amassarwi just as a side-note, the routes for invoice token inventory and invoice batch are almost ready for the backend. This means that we will be able to lazy load them similarly to how we do with proposals. Not sure if this changes the approach needed, but would be good to take it in consideration :+1:

victorgcramos:
I think both approaches are good. The `tokeninventory` approach actually limits a little bit of the filtering experience, since it does not provide all invoices/proposal information.

I would suggest that, after https://github.com/decred/politeia/issues/1036, we provide a better filtering, for proposals, invoices and dccs. I feel that the current proposal filtering is limiting the ux, specially when you try to find an approved proposal a long time ago.

IMO, we should keep this FE filters and when @thi4go 's pr gets in, I jump to that filtering improvements.

src/helpers.js
```diff
@@ -425,11 +425,16 @@ export const getCurrentDateValue = () => {
...
 
-export const getPreviousMonthAndYear = (currentMonth, currentYear) => {
-  const month = currentMonth === 1 ? 12 : currentMonth - 1;
-  const year = currentMonth === 1 ? currentYear - 1 : currentYear;
-  return { month, year };
-};
+export const getDateFromYearAndMonth = ({ year, month }) =>
+  new Date(year, month);
+
+export const getYearAndMonthFromDate = (date) =>
```
tiagoalvesdulce, line 431: Here you are validating the `date` parameter and not returning anything if it comes in falsy. Ideally, it would return some message or throw an error. It would be nice to validate if the `date` parameter is in fact a `Date` object too.

src/containers/Invoice/helpers.js
```diff
@@ -99,3 +102,34 @@ export const getInvoiceTotalAmount = (invoice) => {
...
+  return {
+    start: initialRangeValue,
+    end: finalRangeValue
+  };
+};
+
+export const getPreviousMonthsRange = (range) => {
+  const end = getCurrentDefaultMonthAndYear();
+  const previous = new Date();
+  previous.setMonth(previous.getMonth() - range + 1);
```
tiagoalvesdulce, line 130: Well done with passing negative values here for the setMonth(). It would be nice if you could drop a comment saying why you could do it and a link to the docs. It can be confusing for someone unfamiliar with the `Date API` to see negative numbers here.

src/containers/Invoice/helpers.js
```diff
@@ -99,3 +102,34 @@ export const getInvoiceTotalAmount = (invoice) => {
   const rate = invoice.input.contractorrate / 100;
   return rate * totalHours + totalExpenses;
 };
+
+export const sortDateRange = (initialRangeValue, finalRangeValue) => {
```
tiagoalvesdulce, line 106: I would like to suggest a renaming of these parameters to `firstSelectedDate` and `secondSelectedDate`. `Initial` and `final` confuses the reader during the function. Also, a comment explaining the behavior would be nice here.

src/helpers.js
```diff
@@ -428,8 +428,18 @@ export const getCurrentDateValue = () => {
 export const getDateFromYearAndMonth = ({ year, month }) =>
   new Date(year, month);
 
-export const getYearAndMonthFromDate = (date) =>
-  date && { year: date.getFullYear(), month: date.getMonth() };
+export const getYearAndMonthFromDate = (date) => {
+  try {
+    const d = new Date(date);
+    const isValidDate =
+      date && !!date.getTime && !!d.getTime && d.getTime() === date.getTime();
```
tiagoalvesdulce, line 435: I would go with something simpler like:

```
function isValidDate(d) {
  return d instanceof Date && !isNaN(d);
}
```

src/helpers.js
```diff
@@ -428,8 +428,18 @@ export const getCurrentDateValue = () => {
...
+
+    if (!isValidDate) throw new Error(INVALID_DATE);
+    return { year: date.getFullYear(), month: date.getMonth() };
+  } catch (e) {
+    throw new Error(INVALID_DATE);
```
tiagoalvesdulce, line 440: Here if you want to do a `try...catch` block you should handle the error somehow on `catch` or return it to be handled ate the component level. If you want to trust in the ErrorBoundary and want this to be the fallback, only throwing the error should be enough.

src/helpers.js
victorgcramos: nice suggestion to use the `instanceof` operator! Thanks!

PR #1878, opened 2020-04-26
[pi/cms] Images modal navigation

### LINKED ISSUE #1865: [cms/piwww] Image modals feature enhancement
from @behindtext "i like the modals for the attached images. it would be nice if you could show the additional images by hitting left and right arrow keys"

fernandoabolafio:
This PR introduces a new hook called `useImagesGallery`  which is used to trigger the visualization of multiple images by navigating using the arrow keys or clicking directly in the navigation buttons.

**Preview:**

Screenshot 2020-04-26 at 18 50 42
Screenshot 2020-04-26 at 18 49 06

closes #1865


src/components/Files/ThumbnailGrid.jsx
```diff
@@ -17,13 +16,12 @@ export const ThumbnailGrid = ({
   const files = value.filter(
     (f) => f.name !== "index.md" && f.name !== "data.json"
   );
-  const [handleOpenModal, handleCloseModal] = useModalContext();
-  const openFullImageModal = (file) => {
-    handleOpenModal(ModalFullImage, {
-      image: file,
-      onClose: handleCloseModal
-    });
+  // const [handleOpenModal, handleCloseModal] = useModalContext();
```
tiagoalvesdulce, line 19: Comment

src/components/Files/ThumbnailGrid.jsx
```diff
@@ -37,7 +35,7 @@ export const ThumbnailGrid = ({
               key={`img-${key}`}
               file={f}
               viewOnly={viewOnly}
-              onClick={openFullImageModal}
+              onClick={() => openFullImageModal(key)}
```
tiagoalvesdulce, line 38: This inline function can be changed to a simple function using currying: `openFullImageModal = (idx) => () => ...` and then `onClick={openFullImageModal(key}`.

fernandoabolafio: Sure, but what is the advantage of doing that instead of the inline function in this case?

tiagoalvesdulce:
Testing more carefully I noticed a flickering when there's only one image. Can we prevent that? Ideally, we would have animations between the transition of images but that can be done in another PR.

test

fernandoabolafio:
Yeah, I did notice the flickering. I couldn't fix it because of the way we handle modals now. We no longer have a way of keeping a modal open and only change its props. So I had to close & open again every time the image changed. Do you have any suggestions on how to handle that?
One thing that may be worth trying is to remove modal animations in this given circumstance and see it works.

fernandoabolafio:
Good idea! I will do that :)

fernandoabolafio:
@tiagoalvesdulce I am merging this since it is already a solid improvement of what we have. I will enhance the flickering thing in a subsequent PR.

PR #1886, opened 2020-04-30
Proposal metadata

amassarwi:

This diff adds metadata prop to the posted payload for both `new proposal` & `edit proposal` routes as specified here: https://github.com/decred/politeia/pull/1175

### Solution description

adjusted `makeproposal` function to add metadata prop besides the files property and deleted `data.json` files which was added lately to support creating RFPs

#### Use Cases (if needed)

 1. Create new proposal 
 2. Edit proposal

Depends on: https://github.com/decred/politeia/pull/1175

### DIFF
--- a/src/lib/api.js
+++ b/src/lib/api.js
@@ -52,7 +52,6 @@ export const makeProposal = (
   rfpLink,
   attachments = []
 ) => ({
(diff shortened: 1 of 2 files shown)


lukebp:
I'm going to go ahead an merge 1175 since this is ready.

PR #1901, opened 2020-05-04
[cms] Fix invoice attachments

### LINKED ISSUE #1900: [cms] Attachments while attempting to edit not working
When attempting to attach while editing an existing invoice, it doesn't work.   You are able to select the file, but then you are brought back to the "Select File" modal and nothing appears on the invoice form.  

Even worse I had attached an image to the invoice previously, and then tried to attach another while editing.  After submiting the edit, all attachments were gone.

amassarwi:
### Bug & solution description
This diff fixes few issues regarding invoices attachments:
 - Map correct BE invoice record files property to initial values object
 - Define files custom files validation and call it on files change
 - Fire `global` error message when files aren't valid, and disable submit button.

#### Use Cases
1. Create invoice
2. Edit invoice

Fixes #1900

### DIFF
--- a/src/components/InvoiceForm/InvoiceForm.jsx
+++ b/src/components/InvoiceForm/InvoiceForm.jsx
@@ -1,4 +1,4 @@
-import React, { useMemo, useState, useCallback, useEffect } from "react";
+import React, { useMemo, useState, useCallback } from "react";
 import PropTypes from "prop-types";
 import { BoxTextInput, Button, Message, Spinner } from "pi-ui";
 import { Formik } from "formik";
@@ -23,6 +23,7 @@ import DraftSaver from "./DraftSaver";
 import ThumbnailGrid from "src/components/Files";
 import ExchangeRateField from "./ExchangeRateField";
 import useSessionStorage from "src/hooks/utils/useSessionStorage";
+import useScrollFormOnError from "src/hooks/utils/useScrollFormOnError";
 import { useAction } from "src/redux";
 import { onEditUser } from "src/actions";
 import styles from "./InvoiceForm.module.css";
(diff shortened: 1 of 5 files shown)


src/components/InvoiceForm/InvoiceForm.jsx
```diff
@@ -30,19 +34,24 @@ const InvoiceForm = React.memo(function InvoiceForm({
...
-  }, [errors]);
+  const files = values.files;
+  useEffect(
+    function scrollToTopOnGlobalError() {
+      if (errors.global) {
+        window.scrollTo(0, 0);
+      }
+    },
+    [errors]
+  );
```
tiagoalvesdulce, line 54: I think this could be moved to a custom hook. What do you think?

src/components/InvoiceForm/InvoiceForm.jsx
```diff
@@ -79,10 +92,12 @@ const InvoiceForm = React.memo(function InvoiceForm({
 
   const handleFileRemoval = useCallback(
     (v) => {
-      const fs = values.files.filter((f) => f.payload !== v.payload);
+      const fs = files.filter(
+        (f) => f.payload !== v.payload && f.name !== v.name
```
tiagoalvesdulce, line 96: hmm, why do we need to check the name and the payload here? Asking because only the payload should be enough...

src/components/InvoiceForm/InvoiceForm.jsx
amassarwi: Yep, good idea

src/components/InvoiceForm/InvoiceForm.jsx
amassarwi: You are right, I was uploading same image with different names and it was deleting both, I will revert it, should be stricter condition to filter out duplicates.

Issue #1853, opened 2020-04-07
Wrong html title before loading


thi4go:
Testnet is running latest master branch, and before the app loads, the browser tab shows `%REACT_APP_WEBSITE_NAME` as the title. It shows the variable before it loads any value. Imo the nicest option here would be to not show anything until the corresponding page title loads.

PR #1902, opened 2020-05-05
Fix HTML page title, add .env files for each app, remove redux-logger and cross-env

tiagoalvesdulce:

This PR:
1. Closes https://github.com/decred/politeiagui/issues/1853
2. Removes `redux-logger` dep. If you were still using it you can use [redux-devtools](https://github.com/reduxjs/redux-devtools) instead.
3. Removes `cross-env` dep.

### Solution description

I added 2 config files (`config/ploteia.env` and `config/cms.env`) to control app-specific environment variables. Before, we were setting them in the `package.json` script using the `cross-env` package.


victorgcramos:
LGTM!

PR #1903, opened 2020-05-06
Change storage key name from email to uuid

### LINKED ISSUE #1842: Logout permanently
We need an logout option that deletes all the local browser cache and data. Mainly this should delete the "keyvaluepairs. Ofc this should come with a big warning and a prompt that the user should backup the keys.

### LINKED ISSUE #1843: Remove email from the local keyname
Currently the local DB key for storing the key pairs contains the email of the user as reference. Ideally this should be a random value. 

If that is not possible then it should be replaced with the UUID.

image

thi4go:

The storage key name was disclosing a personal user information, the email. This diff changes the key name to be his public uuid instead of email. This is valid both for the identity key and for local storage draft records. 

This diff also adds a logout modal with the option to permanently logout from the app. This will delete his identity key and his draft entries.

Depends on https://github.com/decred/politeia/pull/1186
Closes #1842 and #1843

### Solution description

Straightforward change on the storage key name we use to set/get a user key. A workaround had to be done for a freshly registered user. Regarding this:

**Problem**: we need the user key to make the `/register (cms)` and `/user/new (pi)` calls on the backend. Also for verifying his email. At this stage, the frontend does not have the user's uuid to set the appropriate storage key name.

**Solution**: use the username as storage key name temporarily, until the user makes his first login on the app. When he does, we check if there's any local key saved under his username, and if so, we change it to be his uuid. This requires a minor change in the backend (PR https://github.com/decred/politeia/pull/1186) so that a user's username is included in the verify email URL. This way we can retrieve his signing key before making the request.


degeri:
this does not break password reset ?

thi4go:
It does not break password reset, mate @degeri. We discussed this and figured uuid would be enough, since it is already a public data. We also need a way to associate the random string to the user, so that he can retrieve his key from the storage properly. This means that even if it were a random string, it would need to be associated with some user data on the storage, which would defeat the initial purpose of it. Bringing this problem up to the backend is not ideal imo, since we would'nt gain much from it being a random string.

degeri:
just reiterating matrix chat here. We have decided to replace email with username to store the keyvalue during the initial setup (before first login) . After that it will be replaced with UUID.

src/lib/pki.js
```diff
@@ -22,31 +22,33 @@ export const toUint8Array = (obj) =>
     ? obj
     : Uint8Array.from(Object.keys(obj).map((key) => obj[key]));
 
-export const loadKeys = (email, keys) =>
-  localforage.setItem(STORAGE_PREFIX + email, keys).then(() => keys);
+export const loadKeys = (uuid, keys) =>
+  localforage.setItem(STORAGE_PREFIX + uuid, keys).then(() => keys);
+export const removeKeys = (uuid) =>
+  localforage.removeItem(STORAGE_PREFIX + uuid);
```
victorgcramos, line 28: I don't know if this is the expected behaviour about this method from what I can see by the [docs](https://localforage.github.io/localForage/#data-api-removeitem), but this does not necessarily remove the keys from the local storage.

thi4go:
@victorgcramos good catch mate. This was actually involving the way we used to clear the `localStorage`, and not `localforage`. Anyhow, I pushed the fix for this :+1:

degeri:
This removes all the users keys  or only the logged in user ?

thi4go:
Only the logged in user @degeri. Thought it would be best, since deleting all user data would involve others identities and drafts that might not be backed up yet. What do you think?

victorgcramos:
@thi4go I agree with your thoughts, makes sense.

degeri:
Sounds fine I am good with it.

tiagoalvesdulce:
@thi4go can you rebase?

Issue #1800, opened 2020-02-28
[rfp] RFPs & submissions details page


amassarwi:
According to screens [6](https://xd.adobe.com/view/44a3298e-a814-42b2-643e-3aa39c8dec34-3292/screen/63bdd388-0241-4bb8-ba18-985433ca7f90/RFP-description-page) & [7](https://xd.adobe.com/view/44a3298e-a814-42b2-643e-3aa39c8dec34-3292/screen/e94fd1ab-256b-4ad7-9d71-eb5aef0b1933/Submission-description-page) in the provided specs:

- [x] RFP should list all submission & show vote info if submissions are under vote
- [x] Submission should be linked to RFP parent

PR #1870, opened 2020-04-18
[rfp] Author/Admin actions

amassarwi:
closes #1801 
closes #1877 
~~based on: #1869 - should be rebased when it's merged`~~

PR #1910, opened 2020-05-09
[rfp] Display submissions list with vote progress info in RFP proposal detials view

amassarwi:
This diff includes the last part of the RFP work, it implements the first requirement in #1800 - *RFP should list all submission & show vote info if submissions are under vote/ finished voting*.

[Specs](https://xd.adobe.com/view/44a3298e-a814-42b2-643e-3aa39c8dec34-3292/screen/63bdd388-0241-4bb8-ba18-985433ca7f90/RFP-description-page)

### Solution description
This diff uses `useProposalBatchWithoutRedux` custom hook which was added & called in #1870 to fetch RFP data to display a link to RFP in submissions body, now we call it to fetch the submissions data in case of RFP proposal being displayed.

#### Use Cases
 - RFP with no submissions
 - RFP with *waiting to vote* submissions
 - RFP with *active* submissions
 - RFP with *finished voting* submissions
 - RFP with abandoned submission

### Dependencies
Based on #1870 and should be rebased when it's merged.
Depends on: https://github.com/decred/pi-ui/pull/257
Closes: #1800 
Closes: #1912 

### UI Changes Screenshot
image

image

image

image

---


src/components/ProposalsList/ProposalItem.jsx
```diff
@@ -0,0 +1,130 @@
...
+              />
+            </div>
+          </Row>
+        )}
+      </Row>
+    )
+  );
+};
+
+// TODO: add props types
```
amassarwi, line 129: On it

src/components/ProposalsList/ProposalsList.jsx
```diff
@@ -0,0 +1,27 @@
...
+          key={index}
+          proposal={proposal}
+          voteSummary={voteSummaries[proposals[index].censorshiprecord.token]}
+        />
+      ))}
+    </div>
+  );
+};
+
+// TODO: propstype
```
amassarwi, line 26: On it

src/components/Proposal/Proposal.jsx
```diff
@@ -46,17 +47,20 @@ const ProposalWrapper = (props) => {
     voteEndTimestamp
   } = useProposalVote(getProposalToken(props.proposal));
   const [proposedFor, setProposedFor] = useState(null);
-  const { linkto } = props.proposal;
-  const [[rfpProposal]] = useProposalBatchWithoutRedux(
-    linkto ? [linkto] : null,
+  const { linkto, linkedfrom } = props.proposal;
+  // if linkto provided => this is a submission => fetch RFP to display link
```
tiagoalvesdulce, line 51: The comment here is nice, please keep it, but I think we can improve the readability of this code a bit more.

I would do:

1. Create a variable called `batchTokens` for the first parameter of `useProposalBatchWithoutRedux` and then do the logical expression.
2. Create a variabale called `fetchVoteSummary` and attribute `!!linkedfrom`.
3. When is the return of `useProposalBatchWithoutRedux` falsy so it defaults to [[]] ? Wouldn't be better to handle this inside the `useProposalBatchWithoutRedux` hook?

This is minor but you can also create a variable `isSubmission` for `linkto` maybe? Because down in the code it's hard to understand the meaning of `linkto` and why it's used in the logical expressions.

src/components/ProposalsList/ProposalItem.jsx
```diff
@@ -0,0 +1,130 @@
...
+  const proposalToken = censorshiprecord && censorshiprecord.token;
+  const { commentsURL, proposalURL } = useProposalURLs(
+    proposalToken,
+    null,
+    null,
+    null
+  );
+  const { voteEndTimestamp } = useProposalVoteTimeInfo(voteSummary);
+
+  const goToFullProposal = useCallback(() => {
```
tiagoalvesdulce, line 46: This function is also defined in the `Proposal.jsx` file. Do we need it in both of this files? If we do, shouldn't we add it to an utils file?

src/components/Proposal/Proposal.jsx
amassarwi: Reasonable, will do!

src/components/ProposalsList/ProposalItem.jsx
amassarwi: Right, I think it can go to one of the `helpers.js` files or something, good point :+1:

src/components/Proposal/helpers.js
```diff
@@ -46,7 +46,9 @@ export const getProposalStatusTagProps = (proposal, voteSummary) => {
       case PROPOSAL_VOTING_NOT_AUTHORIZED:
         return {
           type: "blackTime",
-          text: "Waiting for author to authorize voting"
+          text: proposal.linkto
```
tiagoalvesdulce, line 49: Use `isRFP` variable. The `linkto` property isn't descriptive, we should avoid using it.

src/containers/Proposal/helpers.js
```diff
@@ -227,3 +227,7 @@ export const getCommentsUrl = (proposalToken, javascriptEnabled) =>
 
 export const getAuthorUrl = (userid, javascriptEnabled) =>
   javascriptEnabled ? `/user/${userid}` : `${NOJS_ROUTE_PREFIX}/user/${userid}`;
+
+export const goToFullProposal = (history, proposalURL) => {
```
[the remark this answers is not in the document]
tiagoalvesdulce: partially applied function is the correct name, I guess hahaha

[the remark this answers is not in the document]
amassarwi: True that :+1:

tiagoalvesdulce:
About 3. I'm not seeing the multiple requests here. Does it happen with all users? I could see the duplicated request on #1844 but thought it was fixed after @victorgcramos's comment here: https://github.com/decred/politeiagui/pull/1844#pullrequestreview-395122639. Are they related?

lukebp:
Maybe it had something to do with using the dbutil to add proposal credits.  Try creating a user with these commands.

```
piwww newuser admin@example.com admin password  --verify
politeiawww_dbutil -cockroachdb -testnet -setadmin admin true
politeiawww_dbutil -cockroachdb -testnet -addcredits admin 50
```

tiagoalvesdulce:
Yeah, I was able to repro using the CLI. Couldn't repro doing the entire signup/login flow trough the UI. Probably state related, the implementation might be depending on some previous state to run correctly, which is a bug. LMK if you need help with it @amassarwi

src/components/MonthPickerField/MonthPickerField.jsx
```diff
@@ -130,7 +130,20 @@ const MonthPickerField = ({
                 years={years}
                 values={values[name]}
                 multiChoice={multiChoice}
-                lang={[1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12]}
+                lang={[
```
victorgcramos, line 133: This is minor, but I'd suggest you to create a constant with those values, like `MONTH_NAMES` and import them on both MonthPicker and DatePicker.

amassarwi:
@victorgcramos Thanks for the review mate, will address your note, regarding yarn.lock these changes are result of upgrading `pi-ui`, so they are intended.

amassarwi:
@lukebp 

1. Fixed
2. This generic error is shown for any error which isn't mapped to a custom error message in gui, in this case this error wasn't suppose to be thrown/handled therefore, making sure deadline is bigger than min should be enough here.
5. Fixed

3. I'll add length validation & allowed chars validation.
4. Adding different errors messages for the different cases.

amassarwi:
@lukebp 

3. 
image

4.
image

amassarwi:
I added this more detailed error message when proposal exists but status isn't valid:
image

amassarwi:
> "64 character long hex string" is not very user friendly error message. We can't expect our users to know what hex is or even what a string is. Can we improve this? Compare this too "Name must be at least 8 characters", which is a good error message. It tells the user exactly what is wrong and what they must do to fix it.

image

> The RFP token validation doesn't show an error when its longer than 64 limits.

image

amassarwi:
>Once a RFP submission has been made public, edit the proposal and try to change the linkto token. The user is not shown an error even though politeiawww returns a 400.

> Related to 4, once a proposal is public the linkto token is not able to be changed on the backend. We should enforce this in the gui too so they don't try.

I ended up disabling the type dropdown/select field for all public proposals beside disabling the linkto field for public RFP submission

amassarwi:
> Edit an unvetted RFP submission, change the token to a valid token but not the RFP token, press the edit button. The request never gets sent, but no error is displayed to the user and the token field is updated so it looks like the request was sent and the proposal was updated.

Fixed, good catch - redux re-rended the edit form which swallowed the error, I've decoupled the fetcing from redux state to prevent rerending on fetch error.

amassarwi:
> Once the runoff vote is started the proposal submissions show up on the RFP details page, but why do they not show up on the RFP details page before the vote? They should.

This was resolved with @lukebp's latest changes on BE side.
image

amassarwi:
@lukebp 

> When approving an RFP proposal as an admin, the proposal metadata fields like linkto are not visible. They need to be. There is going to be additional proposal metadata added in the future so we need to go ahead and make a way to display it for the admin.

Added `expires` \ `expired` event to proposal to display deadline:

image

Regarding `linkto`, both unvetted & public proposals show the nice `proposed for` link, to link submissions with their parent RFP, this is sufficient IMO, what do you think?

src/components/ModalStartVote/ModalStartVote.jsx
```diff
@@ -16,7 +16,7 @@ import usePolicy from "src/hooks/api/usePolicy";
 import { isRfpReadyToVote } from "src/containers/Proposal/helpers";
 import { VOTE_TYPE_STANDARD, VOTE_TYPE_RUNOFF } from "src/constants";
 
-const preDefinedDurations = [2016, 2880, 4032];
+const preDefinedDurations = [2, 2880, 4032];
```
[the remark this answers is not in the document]
tiagoalvesdulce: Made my comment in a previous commit, but the question persists.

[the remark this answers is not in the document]
amassarwi: Good idea, will add few testnet values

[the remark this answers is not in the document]
amassarwi: testnet:
image

mainnet:
image

amassarwi:
@tiagoalvesdulce updated mate, thanks

amassarwi:
@lukebp 
When I try to edit an unvetted or public RFPs (changed deadline date) I get the following error from BE:
image

Editing proposal's title or body works as expected tho, thoughts?

lukebp:
It's a known issue.  Include a change to the md file for now until I get a chance to fix it.

decred/politeia#1194

amassarwi:
@lukebp 
Fixed for good

PR #1930, opened 2020-05-15
Fix temp invite property mapping

### LINKED ISSUE #1928: [cms] Invite Temp Contractor not working
Looks like the ModalInviteContractor isn't properly sending the isTemp value down to the the api action (temp is undefined in actions/api/onInviteContractor).  So the request gets sent out to www as false.

amassarwi:
Fixes #1928 

### Solution description
Fixed form property mapping in action

### DIFF
--- a/PULL_REQUEST_TEMPLATE.md
+++ b/PULL_REQUEST_TEMPLATE.md
@@ -1,5 +1,3 @@
-### Bug (or issue) description
-
 > Quick explanation about the issue that the PR is aiming to solve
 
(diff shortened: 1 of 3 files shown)


PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,5 +1,3 @@
-### Bug (or issue) description
```
amassarwi, line 1: Not related to the bug.
It's a bit cleaner when the PR body starts with the description without the deleted header.

Issue #1856, opened 2020-04-14
Add config flag to switch paywall fetching off


amassarwi:
*Motivation:* When paywall data isn't available we would like to run the app without the paywall fetching.

PR #1931, opened 2020-05-15
Poll payment routes only if the paywall is enabled in the backend

tiagoalvesdulce:
This PR is using the `paywallenabled` flag from the BE to enable the paywall in the gui side. With that, https://github.com/decred/politeiagui/issues/1856 is no longer necessary. Closes https://github.com/decred/politeiagui/issues/1856

### Solution description

This diff adds a function to inject the flag sent by the BE in the `policy` route to the app config and uses this flag to enable/disable polling paywall routes, including proposal credits.

**Note:** this PR will not hide the Credits tab nor its contents in the Account page.

### Testing this PR

Comment keys `paywallxpub` and `paywallamount` in the politeiawww config file. You should be able to login, comment and create proposals without any restriction. Uncomment them and you will have to pay the paywall to comment and buy proposal credits to create proposals. This PR shouldn't affect the CMS app.

### DIFF
--- a/src/apps/cms/config.json
+++ b/src/apps/cms/config.json
@@ -6,6 +6,7 @@
   "aboutContent": "",
   "enableAdminInvite": true,
   "enableCommentVote": false,
+  "enablePaywall": false,
   "enableCredits": false,
   "onBoardContent": "onboard-cms",
   "onBoardLink": "https://docs.decred.org/contributing/overview/",
--- a/src/apps/cms/config.json
+++ b/src/apps/cms/config.json
@@ -7,7 +7,6 @@
   "enableAdminInvite": true,
   "enableCommentVote": false,
   "enableCredits": false,
-  "enablePaywall": false,
   "onBoardContent": "onboard-cms",
   "onBoardLink": "https://docs.decred.org/contributing/overview/",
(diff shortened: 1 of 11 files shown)


src/apps/cms/config.json
```diff
@@ -6,8 +6,8 @@
   "aboutContent": "",
   "enableAdminInvite": true,
   "enableCommentVote": false,
-  "enableCredits": false,
   "enablePaywall": false,
```
amassarwi, line 9: Can be deleted?

src/apps/politeia/config.json
```diff
@@ -8,7 +8,6 @@
   "enableAdminInvite": false,
   "enableCommentVote": true,
   "enableCredits": true,
```
amassarwi, line 10: Is this still needed if you inject both in `Config.js`

src/apps/cms/config.json
tiagoalvesdulce: I think it's nice to keep this for CMS as a fallback because the `paywallenabled` flag is only added to proposals at the moment.

src/apps/politeia/config.json
tiagoalvesdulce: removed!

PR #1935, opened 2020-05-19
[cms] Add new contractor types

### LINKED ISSUE #1929: [cms] Add TempRevoked and Temporary contractor types
Admins should be able to update users into temporary revoked and temporary contractor types using the manage user form.

amassarwi:
This diff closes #1929 - adds the following options `Nominee`, `Revoked`,`Temp. Contractor`, `Revoked Temp.` to the admin contractor edit form.

### Solution description
Added new value to defined list of possible values

### DIFF
--- a/src/containers/User/Detail/ManageContractor/helpers.js
+++ b/src/containers/User/Detail/ManageContractor/helpers.js
@@ -3,7 +3,6 @@ export const typeOptions = [
   "Direct",
   "Supervisor",
   "Sub Contractor",
-  "Nominee",
   "Revoked",
   "Temp. Contractor",
   "Revoked Temp."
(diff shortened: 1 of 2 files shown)


src/containers/User/Detail/ManageContractor/helpers.js
```diff
@@ -1,18 +1,11 @@
...
-
-export const typeViewOptions = [
   "No type defined",
   "Direct",
   "Supervisor",
   "Sub Contractor",
-  "Not a contractor",
-  "Dormant",
-  "Nominee"
+  "Revoked",
```
victorgcramos, line 6: what if contractortype = 4? From the API, the contractortype = 4 means a nominee user, but in this case, the user will be labeled "Revoked", right?

amassarwi: good point, i should skip `nominee` only in the select options mapper function but keep it here to not break the indexes which should match BE values.

amassarwi:
@victorgcramos updated mate

PR #1942, opened 2020-05-26
[cms] Fix Supervisor ID's list

### LINKED ISSUE #1937: [cms] Unable to set supervisor for subcontractor
Currently it shows a dropdown but shows "no options" and i'm unable to manually fill in anything there.

victorgcramos:
This diff closes #1937 issue, which reported a bug with the supervisor ids selector on Manage Contractor form. No options were being shown.

### Solution description

The list of supervisors options was empty because it was getting information from user's supervisor list, not the supervisors list available on the CMS platform. The solution is: create a `useSupervisors` hook to fetch all supervisors (since this can only be performed by admins) and populate the selector's options to display them.

#### Use Cases (if needed)

The `useSupervisors` hook can be used for fetching supervisors.

### UI Changes Screenshot

Before:
Captura de Tela 2020-05-26 às 00 21 02

After:
Captura de Tela 2020-05-26 às 00 20 26

### DIFF
--- a/src/containers/User/Detail/ManageContractor/ManageDccForm.jsx
+++ b/src/containers/User/Detail/ManageContractor/ManageDccForm.jsx
@@ -1,24 +1,29 @@
-import { Spinner, Button, Message, H2 } from "pi-ui";
+import { Spinner, Button, Message, H2, Select } from "pi-ui";
 import SelectField from "src/components/Select/SelectField";
-import React, { useCallback, useMemo, useState } from "react";
+import React, { useCallback, useState, useMemo } from "react";
 import InfoSection from "../InfoSection.jsx";
-import { selectTypeOptions, selectDomainOptions } from "./helpers";
-import UserSearchSelect from "src/containers/User/Search/SearchSelector";
+import {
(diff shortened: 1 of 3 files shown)


src/containers/User/Detail/ManageContractor/ManageDccForm.jsx
```diff
@@ -1,47 +1,54 @@
...
-        };
-      }),
-    [users]
+      supervisorsOptions.filter((option) =>
+        supervisoruserids.some((supervisorid) => supervisorid === option.value)
+      ),
+    [supervisorsOptions, supervisoruserids]
   );
 
+  const isLoading =
```
tiagoalvesdulce, line 49: Do we really have to check for undefined? what are the possible types of `domain` and `contractortype`? Maybe we can improve this logical expression

victorgcramos: hey! This checking is done because domains and contractortypes can both be `0`. This is why we check if they are `undefined`

PR #1944, opened 2020-05-26
Add Timeout for Comments Anchoring

### LINKED ISSUE #1941: Workaround for timeout during anchoring
decred/politeia#1112 is a well known issue that keeps causing users a lot of problems.  This will be fixed with the new tlog backend, but that work is still in progress.  As a temporary workaround, can we prevent submitting comments in the gui starting at minute 58 of every hour and lasting for 5 minutes?  We should show an error message to the user explaining the situation.  Something along the lines of:

"Commenting temporarily unavailable while an hourly censorship resistance routine is in progress. Sorry for the inconvenience. This will be fixed soon. Check back in 5 minutes."

victorgcramos:
This diff Closes #1941 issue, which demands the addition of a 5 minutes timeout on minute 58 of every hour.

### Solution description

The solution is basically to check for the current browser time, and if its `minutes + seconds + milliseconds` are greater than 58 minutes. If so, block comments and display the following warning message:
> Commenting temporarily unavailable while an hourly censorship resistance routine is in progress. Sorry for the inconvenience. This will be fixed soon. Check back in 5 minutes.

One thing that is added by this diff is the real time handler for this warning, since users can comment while navigating on the screen. It shows up on minute 58, and disappears 5 minutes later.

### UI Changes Screenshot
When comments are blocked due to comments anchoring:

- Light Mode: Captura de Tela 2020-05-26 às 20 16 01

- Dark Mode: Captura de Tela 2020-05-26 às 20 16 10

### DIFF
--- a/src/containers/Comments/Comments.jsx
+++ b/src/containers/Comments/Comments.jsx
@@ -235,9 +235,7 @@ const Comments = ({
       if (isAnchoring) {
         timeout = setTimeout(handleToggleAnchoring, ANCHORING_TIMEOUT_MS);
       }
-      return () => {
-        if (timeout) clearTimeout(timeout);
-      };
+      return () => timeout && clearTimeout(timeout);
     },
     [isAnchoring, handleToggleAnchoring]
   );
@@ -247,9 +245,7 @@ const Comments = ({
       if (!isAnchoring) {
         timeout = setTimeout(handleToggleAnchoring, getTimeLeftToAnchor());
       }
-      return () => {
-        if (timeout) clearTimeout(timeout);
-      };
+      return () => timeout && clearTimeout(timeout);
     },
     [isAnchoring, handleToggleAnchoring]
   );
--- a/src/containers/Comments/Comments.jsx
+++ b/src/containers/Comments/Comments.jsx
@@ -36,6 +36,10 @@ import useLocalStorage from "src/hooks/utils/useLocalStorage";
 
 const COMMENTS_LOGIN_MODAL_ID = "commentsLoginModal";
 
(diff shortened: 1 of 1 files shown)


src/containers/Comments/Comments.jsx
```diff
@@ -209,6 +213,46 @@ const Comments = ({
...
+    [isAnchoring, handleToggleAnchoring]
+  );
+  useEffect(
+    function commentsAllowedListener() {
+      let timeout = null;
+      if (!isAnchoring) {
+        timeout = setTimeout(handleToggleAnchoring, getTimeLeftToAnchor());
+      }
+      return () => {
+        if (timeout) clearTimeout(timeout);
```
amassarwi, line 251:
```suggestion
        return () => timeout && clearTimeout(timeout);
```

src/containers/Comments/Comments.jsx
```diff
@@ -209,6 +213,46 @@ const Comments = ({
...
+  }, [setIsAchoring, isAnchoring]);
+
+  useEffect(
+    function anchoringModeListener() {
+      let timeout = null;
+      if (isAnchoring) {
+        timeout = setTimeout(handleToggleAnchoring, ANCHORING_TIMEOUT_MS);
+      }
+      return () => {
+        if (timeout) clearTimeout(timeout);
```
amassarwi, line 239:
```suggestion
      return () => timeout && clearTimeout(timeout);
```

victorgcramos:
@tiagoalvesdulce I talked to @lukebp and we agreed that since this is temporary, we should not worry about it.

PR #1945, opened 2020-05-27
Update message after verify new identity

### LINKED ISSUE #1917: [CMS] Wrong message after clicking the link in the email to verify new identity
The success message right now is: "Ok, go to proposals". In CMS it should be: "OK, go to your invoices".

blaltarriba:
This diff closes #1917 - update the message after clicking the link in the email to verify new identity.

### Solution description

Update the message to `OK, go to your invoices`

### UI Changes Screenshot

Before diff:
Screenshot 2020-05-27 at 12 20 45

After diff:
Screenshot 2020-05-27 at 12 26 27

### DIFF
--- a/src/pages/User/VerifyKey/VerifyKey.jsx
+++ b/src/pages/User/VerifyKey/VerifyKey.jsx
@@ -50,6 +50,7 @@ const VerifyKey = ({ location, history }) => {
   const success = verifyUserKey && verifyUserKey.success;
   const error = verifyUserKeyError;
   const pushToHome = useCallback(() => history.push("/"), [history]);
+  const successButtonText = (isCMS ? 'Ok, go to invoices' : 'Ok, go to proposals')
 
   return (
     <SingleContentPage>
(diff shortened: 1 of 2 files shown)


src/pages/User/VerifyKey/VerifyKey.jsx
```diff
@@ -92,7 +92,7 @@ const VerifyKey = ({ location, history }) => {
               className={classNames("margin-top-l", styles.btn)}
               onClick={pushToHome}
               type="button">
-              Ok, go to proposals
+              Ok, go to invoices
```
amassarwi, line 95: This message should conditioned, depends on which the app is running, we have a redux state boolean `isCms` used cross the app, we need to use here as well and show two messages according to the boolean state.

blaltarriba: Ouch, I though this was just used in the CMS. Sure, I'll check that then. Thanks for your reply.

Issue #1906, opened 2020-05-07
Scroll to record's height on list view


victorgcramos:
I just noticed this while navigating through old approved/rejected proposals. When I select one old proposal from the list to see its details, and then return to the listing page, the proposals are loaded again. We already have them cached on our state, so I guess it can be done easily.

If user has no info cached, load information and don't scroll.

See below:
![navigation-scroll](https://user-images.githubusercontent.com/22639213/81332081-e45d0c80-9078-11ea-8d21-a48e35a39650.gif)

PR #1949, opened 2020-05-28
Fix unnecessary tokeninventory fetch

tiagoalvesdulce:
Closes https://github.com/decred/politeiagui/issues/1906

### Solution description

~~Really simple one-liner that fixes the tokenInventory selector but took me some time to come up with it :)~~ Turn out the solution was a bit more complex and I also improved the way we handle request actions. I'm now only requesting the tokenInventory when it's needed.

### DIFF
--- a/src/selectors/models/proposals.js
+++ b/src/selectors/models/proposals.js
@@ -15,7 +15,7 @@ export const allByStatus = createDeepEqualSelector(
 );
 export const tokenInventory = createDeepEqualSelector(
   get(["api"]),
-  ({ tokenInventory }) => tokenInventory.response
+  ({ tokenInventory }) => tokenInventory.payload
 );
 export const newProposalToken = get(["proposals", "newProposalToken"]);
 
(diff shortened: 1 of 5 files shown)


thi4go:
Hey mate, so I was wondering something. Since we finished with this redux refactor (https://github.com/decred/politeiagui/issues/1490), shouldn't the selector be getting the token inventory response from the normalized models reducers? Maybe `models/proposals`? I think that by adding the response payload on `receive`, which will add data to the `api` state branch, we would be back to square one and hurt our single source of truth principle, that after the refactor, is supposed to be on `models` branch.

tiagoalvesdulce:
Not sure if I get what you're saying but we want to check the raw tokenInventory from the API with the `allByStatus` (normalized state). The token inventory payload shouldn't be used anywhere else.

src/selectors/models/proposals.js
```diff
@@ -15,7 +15,7 @@ export const allByStatus = createDeepEqualSelector(
 );
 export const tokenInventory = createDeepEqualSelector(
   get(["api"]),
```
thi4go, line 17: I think that tokenInventory should fetch the data from `models/proposals`, and no the `api` state branch. Is there any specific reason it is fetching this from the `api`?

tiagoalvesdulce: Yes, there is. I want to check if both are in sync, if they aren't, I want to do a new fetch.

PR #1950, opened 2020-05-28
Reduce testnet validation values

tiagoalvesdulce:
### Solution description

Reduce FE validation of quorum % and pass % to 1 when on testnet.

### DIFF
--- a/src/components/ModalStartVote/validation.js
+++ b/src/components/ModalStartVote/validation.js
@@ -1,30 +1,39 @@
 import * as Yup from "yup";
 import { minAmountMessage, maxAmountMessage } from "src/utils/validation";
 
-const MIN_QUORUM_PERCENTAGE = 20;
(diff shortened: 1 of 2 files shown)


src/components/ModalStartVote/validation.js
```diff
@@ -1,30 +1,39 @@
...
+      )
+      .max(
+        MAX_QUORUM_PERCENTAGE,
+        maxAmountMessage("quorum percentage", MAX_QUORUM_PERCENTAGE)
+      ),
+    passPercentage: Yup.number()
+      .required("required")
+      .min(
+        getMinPassPercentage(isTestnet),
+        minAmountMessage("pass percentage", getMinPassPercentage(isTestnet))
```
amassarwi, line 33: Minor: we can add a `const minPassPercentage = getMinPassPercentage(isTestnet)` instead of calling the function twice, same holds for `getMinQuorumPercentage`.

PR #1954, opened 2020-05-28
Improve up/down vote behavior

tiagoalvesdulce:
This PR aims to improve the behavior of comments up/down votes.
 
### Solution description

I decided to go with a simple solution because it's good enough for me.

We have a small delay (1sec) to show the action payload in the UI and this delay works as a debounce and prevents multiple requests. I did some research and IMO the approach on this PR is better then most websites. On reddit, for example, you can fire hundreds of requests clicking the up/down arrows and eventually the backend will block you and the UI will become non-responsive.

Does anyone has a suggestion of a better approach/implementation for the up/down votes?

### DIFF
--- a/src/components/Likes/Likes.jsx
+++ b/src/components/Likes/Likes.jsx
@@ -81,7 +81,6 @@ const Likes = ({
         <>
           <div className={styles.leftLikeBox}>
             <button
-              disabled={disabled}
               ref={likeRef}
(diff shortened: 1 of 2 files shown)


lukebp:
I agree. This sounds like a good solution.

victorgcramos:
Makes sense! LGTM!

PR #1955, opened 2020-05-28
Fix Reset Password flow

### LINKED ISSUE #1958: Reset Password and resend verification email redirecting to the wrong page
It's the same issue in both cases. When clicking on the `Reset password` or `Resend email` button you get redirected to the home.

victorgcramos:
This PR refers to an issue reported by @tiagoalvesdulce on [Politeia's Channel on Matrix](https://matrix.to/#/!VFRvyndKpzcLrVslQD:decred.org/$1590700619764ppHaA:decred.org?via=decred.org&via=matrix.org&via=zettaport.com) which described a bug regarding the **resetting password** flow. We all agreed in the discussion that this piece of code should be refactored in order to follow the expected information flow, which is:

Login form > Click on [reset password]() > fill the form > redirect to a custom page telling user to check his/her email > user goes to email and clicks on the link sent by Politeia's servers > password reset form

Closes #1958 

### Solution description

This bug was caused by a null key access on `RECEIVE_RESET_PASSWORD` reducer. Since we only get the `verificationtoken` on dev environments, there's no need for using a reducer to store this information.

Also, the *email message* page was the same as the *request reset password* page, which was displayed after the `/password/reset` request gets completed. This diff adds a custom page to handle this.

### UI Changes Screenshot

Before:
reset-password-pre

After:
reset-password-post


src/pages/User/ResetPasswordMessage.jsx
```diff
@@ -0,0 +1,38 @@
...
+    [history, username, verificationtoken]
+  );
+
+  return (
+    <SingleContentPage>
+      <EmailSentMessage
+        email={"your email"}
+        title={"Please check your mailbox to reset your password"}
+      />
+      {verificationtoken && (
```
tiagoalvesdulce, line 25: I don't think we need this conditional rendering here if we are already passing `verificationtoken` to the show prop.

src/pages/User/ResetPasswordMessage.jsx
```diff
@@ -0,0 +1,38 @@
...
+import { Button } from "pi-ui";
+import { withRouter } from "react-router-dom";
+import qs from "query-string";
+import EmailSentMessage from "src/components/EmailSentMessage";
+import SingleContentPage from "src/components/layout/SingleContentPage";
+import DevelopmentOnlyContent from "src/components/DevelopmentOnlyContent";
+
+const PageRequestPasswordMessage = ({ location, history }) => {
+  const { verificationtoken, username } = qs.parse(location.search);
+  const pushToResetPage = useCallback(
```
tiagoalvesdulce, line 11: no need to use `useCallback` here since we are not passing this function down the component tree.

src/pages/User/ResetPasswordMessage.jsx
victorgcramos: we need this to avoid the div padding

amassarwi:
@tiagoalvesdulce Sure thing, will have a look tomorrow morning.

Issue #1951, opened 2020-05-28
RFP Coordination Checklist


victorgcramos:
This issue is a guide for RFP Testing. Just an idea of what should we pay attention to:

Reference: 
*multi: Add RFP proposal. https://github.com/decred/politeia/pull/1054*

User:

- [ ] Create an RFP **Proposal**
- [ ] Edit an RFP **Proposal**
- [ ] Authorize RFP Proposal voting
- [ ] Wait for RFP Proposal to reach the min quorum
- [ ] Create multiple RFP **Submissions**
- [ ] Edit an RFP **Submission**
- [ ] Wait for RFP Proposal Deadline gets ended
- [ ] Comments on an RFP Proposal
  - [ ] New comment
  - [ ] Upvote and Downvote comments
  - [ ] Reply comments

Admin:
- [ ] Create an RFP Proposal
- [ ] Make it public & start voting on RFP
- [ ] Approve it by casting enough votes
- [ ] When RFP Proposal is approved, add RFP Submissions
   - [ ] Create multiple public submissions
- [ ] Vote on multiple RFP Submissions
- [ ] Censorship:
  - [ ] Censor RFP Comments
  - [ ] Mark RFP Proposal as Spam
  - [ ] Mark RFP Submission as Spam

Things to keep in mind:
  - Once the approved RFP submission deadline expires, an admin can start the voting process on the RFP submissions at any time. This is a new type of vote called a runoff vote. Whe an admin starts the runoff vote, the voting period on all public, non-abandoned RFP submissions will
begin simultaneously
  - Unlike a normal proposal vote, no vote authorization is required from the submission author
  - The winning proposal is the proposal that meets the quorum % requirement, meets the pass % requirement, and that has the most net yes votes. *Note: this means that it will be possible for a proposal to meet the quorum and pass requirements but still be rejected if it does
not have the most net yes votes.*
  - A link is not official until the RFP submission is made public, leaving this decision in the hands of the admins.

@tiagoalvesdulce @amassarwi @thi4go @lukebp feel free to add suggestions for this issue and checklist! Thanks!

amassarwi:
Proposal form need some love on mobile:
![image](https://user-images.githubusercontent.com/10324528/83189306-546b2980-a131-11ea-88bd-1523c07a243a.png)

PS: on it

victorgcramos:
Got this error when selecting the first available date on RFP Proposal deadline:
[...]

PR #1956, opened 2020-05-29
[rfp] Improve proposal form for mobile

amassarwi:
This diff improves the new proposal form for mobile screens - see before/after screenshot below.
Mentioned here: https://github.com/decred/politeiagui/issues/1951#issuecomment-635585549

### Solution description

Improved *flex* styling for mobile screen which included extending the `Row` component to accept `wrap` prop to make each item span to full row on mobile/when `wrap` boolean is true.

### UI Changes Screenshot

After:
Screenshot 2020-05-29 at 02 14 06
Screenshot 2020-05-29 at 02 19 29

Before:
image

### DIFF
--- a/src/components/ProposalForm/ProposalForm.jsx
+++ b/src/components/ProposalForm/ProposalForm.jsx
@@ -120,6 +120,7 @@ const ProposalForm = React.memo(function ProposalForm({
       )}
       <Row
         noMargin
+        wrap={smallTablet}
         className={classNames(
           styles.typeRow,
           isRfpSubmission && styles.typeRowNoMargin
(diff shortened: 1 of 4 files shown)


amassarwi:
@tiagoalvesdulce Fixed mate, for some weird reason *composes* was applied also for normal screens even tho it's wrapped inside the media query!

tiagoalvesdulce:
Screen Shot 2020-05-29 at 11 01 58

The icon goes down to its own row on screens smaller than 410px

amassarwi:
@tiagoalvesdulce Icon *flexed* as well

PR #1962, opened 2020-05-29
[wip] Allow short token URLs

### LINKED ISSUE #1939: Allow short token URLs
decred/politeia#1136 adds the ability to use shortened tokens, the first 7 characters of the token, to retrieve the proposal details.  This should also work for all routes that require the token as a route param.

Now that we can lookup proposals using the shortened token, we want to be able to reference proposals using shortened URLs.

https://proposals.decred.org/proposals/0230918

amassarwi:
This diff allows users to use short proposal's token in URLs - first 7 chars of original 64 chars hex.

### Solution description
- Adjusted `useProposal` hook to fetch voteSummary only when full token is fetched and stored in redux store.
- Updated `makeGetProposalByToken` selector to try find the stored proposal by using the 7 chars prefix instead of full length token.

### Dependencies
Closes #1939

### DIFF
--- a/src/containers/Proposal/Detail/hooks.js
+++ b/src/containers/Proposal/Detail/hooks.js
@@ -55,9 +55,21 @@ export function useProposal(token, threadParentID) {
         return;
       }
       onFetchProposal(token);
-      onFetchProposalsVoteSummary([token]);
     },
     [proposal, token, onFetchProposal, onFetchProposalsVoteSummary]
+  );
+  useEffect(
+    // fetch vote summary only when proposal details loaded => full token is avaliable in redux store
+    // thus can be used to fetch vote summary
+    function fetchVoteSummary() {
+      if (proposalFromState) {
+        const {
+          censorshiprecord: { token }
+        } = proposalFromState;
+        onFetchProposalsVoteSummary([token]);
+      }
+    },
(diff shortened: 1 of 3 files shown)


amassarwi:
@tiagoalvesdulce Intended mate, edit route need the full token so I kept it as is in that case - when edit is done I do use the short token when redirecting the user to proposal's details page

src/components/Proposal/helpers.js
```diff
@@ -27,19 +26,6 @@ export const getStatusBarData = (voteSummary) => {
     .sort((a) => (a.label === "yes" ? -1 : 1));
 };
 
-export const getProposalUrl = (token, javascriptEnabled) =>
```
victorgcramos, line 30: Is there any specific reason for removing this piece of code?

amassarwi: hehe sure, it's just a cleanup at some point we created `useProposalURLs` hook and it uses the same functions which are defined in https://github.com/decred/politeiagui/blob/master/src/containers/Proposal/helpers.js#L225-L236

amassarwi: image

victorgcramos: oh good! Thanks! haha

PR #1964, opened 2020-05-31
Add copy to clipboard to proposal token

### LINKED ISSUE #1936: Add copy to clipboard button to the token in the proposal details page
In order to do that we have to remove the current pencil icon that is next to the token and add our default copy to clipboard button to the right of it.

blaltarriba:
This diff closes #1936 - Add copy to clipboard button to the token in the proposal details page.

### Solution description

This diff add the possibility to copy to clipboard the proposal token in the proposal details page. 
To remove the pencil icon, I stopped to use `RecordToken`. I didn't change `RecordToken` as is used in other places.

### UI Changes Screenshot

Before diff:
Screenshot 2020-06-01 at 00 12 35

After diff:
Screenshot 2020-06-01 at 00 12 04

### DIFF
--- a/src/components/Proposal/Proposal.jsx
+++ b/src/components/Proposal/Proposal.jsx
@@ -173,8 +173,7 @@ const Proposal = React.memo(function Proposal({
           Header,
           Subtitle,
           Edit,
-          Status,
-          RecordToken
+          Status
         }) => (
           <>
             <Header
@@ -266,7 +265,15 @@ const Proposal = React.memo(function Proposal({
             />
             {extended && (
               <Row topMarginSize="s">
-                <RecordToken token={proposalToken} />
+                <Text id={`proposal-token-${proposalToken}`} truncate>
+                  {proposalToken}
+                </Text>
+                <CopyLink
+                  className={classNames(
+                    isPublicAccessible && styles.copyLink && "margin-left-s"
+                  )}
+                  url={proposalToken}
+                />
               </Row>
             )}
             {hasvoteSummary && (


tiagoalvesdulce:
Thanks for the PR @blaltarriba. I think that would be nice to use our [copyable component](https://decred.github.io/pi-ui/components/copyable) for this. What do you think?

blaltarriba:
Make sense if there is already a component including the copy-to-clipboard icon but checking `Copyable` component I see the icon(a blue notebook) is different than the icon used at the bottom of proposal details page(a white chain with grey background). I don't like the idea to have different icons to make the same action at the same page so I would keep as is. What do you think @tiagoalvesdulce ?

tiagoalvesdulce:
Hmm, I get your point but our icon in the right corner of the proposal description is to copy the link of the page. Tbh I don't know what is best. Let's get @MariaPleshkova input on that.

So, Maria, we want to add the "copy to clipboard" functionality to the proposal token in the proposal details page because it can be useful now that we have RFPs. @blaltarriba implemented the solution with the link copy icon but I think we should go with the `copy to clipboard` component we have on pi-ui. What do you prefer? Feel free to suggest a different solution too. :)

src/components/Proposal/Proposal.jsx
```diff
@@ -266,7 +265,15 @@ const Proposal = React.memo(function Proposal({
             />
             {extended && (
               <Row topMarginSize="s">
-                <RecordToken token={proposalToken} />
```
victorgcramos, line 269: I think using this component, and adding the copy button to it, might be a better option in terms of readability.

If you want to avoid invoices and dccs to have this copy icon, I think you should consider using a prop for this component to trigger the "copy" mode.

blaltarriba: I'm mainly backend developer but I started with frontend because is easy to get context about how Politeia works.

Any feedback to improve my frontend coding is appreciated. I'll check your suggestion. Thanks @victorgcramos

MariaPleshkova:
@tiagoalvesdulce @blaltarriba well, as for me this blue notebook icon means "copy something you see". Like here, for example:
image
But in our case we can copy the whole link, not only displayed numbers. So I think we can use the gray chain icon here.
I also think maybe we can remove it from the bottom then? What do you think?

blaltarriba:
Thank you for your reply @MariaPleshkova 

As far as I understand this issue means to add an icon to copy-to-clipboard just the token. That means both icons copy-to-clipboard different things, one only the token and the other one the whole link. Make sense to keep both or just keep one to copy the whole link? I have a lack of context here so I'm not sure what is the best approach.

In case to keep one icon, I would keep the new one and remove the icon from the bottom.

What do you think @tiagoalvesdulce @MariaPleshkova ?

src/components/Proposal/Proposal.jsx
victorgcramos: Awesome! Anytime, mate! Thank you!

victorgcramos:
IMO, I still don't think that changing the bottom icon is necessary. One is for links and other is for copying what is being displayed on the field.

tiagoalvesdulce:
@MariaPleshkova the icon discussed in this PR added by @blaltarriba would allow the user to copy only the proposal token, not the full link.

amassarwi:
I also think we should use `copy to clipboard` component here, don't see any issue with having the link icon and copy to clipboard token icon/component.

MariaPleshkova:
Ok. There are 2 ways then (as @blaltarriba already mentioned):
1. keep 2 different icons as they mean different actions
2. keep only "chain" icon to copy the whole link

What I want to know is if copying only the proposal token instead of the whole link is an action that users may need? If yes, then let's use blue note icon for it and keep the bottom one as it is.

blaltarriba:
Seeing some of you agree with  it and the issue means that, I'm going to use `CopyableText` component and keep the bottom icon as is. (Option 1 of last @MariaPleshkova comment)

The `CopyableText` component will copy "what you see" and the grey chain icon in the bottom copy the whole link, so there will be different icons for different actions.

tiagoalvesdulce:
Hey @blaltarriba, looks like the `CopyableText` truncate prop is not working as expected. I think its implementation is wrapper dependant (this means it varies depending on the parent positioning, which is a bug). The PR looks great but we will have to hold on it for now. If you want to investigate that and create a detailed issue on `pi-ui` it will be very useful and a good way to get more involved with CSS stuff. Let me know if you want to work on that.

tiagoalvesdulce:
@blaltarriba I fixed this issue on https://github.com/decred/pi-ui/pull/275. Can you rebase and run `yarn upgrade pi-ui`?

PR #1966, opened 2020-06-01
Prevent comments and proposals submission when anchoring

tiagoalvesdulce:
This PR adds the `isAnchoring` function and prevents proposals and comments submission during a given time (default to 06:58 UTC and 10 minutes).

### Solution description

My solution is a simple diff between the current time and the target time and check if its greater than 0 and less than the set duration time.

Depends on https://github.com/decred/politeia/pull/1205

### DIFF
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -473,5 +473,3 @@ export const isAnchoring = (
   );
   return timeDiffMinutes >= 0 && timeDiffMinutes < anchoringDuration;
 };
-
-isAnchoring();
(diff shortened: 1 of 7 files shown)


src/helpers.js
```diff
@@ -447,3 +447,31 @@ export const fromUSDCentsToUSDUnits = (cents) =>
...
+  targetDate.setUTCHours(startHour);
+  targetDate.setUTCMinutes(startMinute);
+  const timeDiffMinutes = getTimeDiffInMinutes(
+    new Date().getTime(),
+    targetDate.getTime()
+  );
+  return timeDiffMinutes >= 0 && timeDiffMinutes < anchoringDuration;
+};
+
+isAnchoring();
```
victorgcramos, line 477: Why are you calling this here? Any specific reason? If so, could you please leave a comment here referencing the method you are using? Thanks a lot!

tiagoalvesdulce: It was for testing purposes and I forgot to remove it. Thanks for noticing :)

PR #2018, opened 2020-06-17
[cms] Add formatting helper for cents to USD

alexlyp:
This PR makes sure that all values displayed on the payouts page and exported csv are consistently shown as XXX.XX USD and XXX.XX DCR.

### Solution Description

Add a formatting helper to utils `formatCentsToUSD` that takes cents as the argument and returns a string with XXX.XX USD.


xaur:
> XXX.XX DCR

Truncating DCR to 2 decimal places doesn't sound right, unless I'm missing something.

PR #2033, opened 2020-06-25
[cms] Add `domain` & `contractor type` filters to user search page

### LINKED ISSUE #1959: [cms] Add CMSUsers request for contractor type and domain
This could probably be another page in the form of "Search for users" but instead of username/email, it would allow listing of users by domain/contractor type.  This is already provided by the CMSUsers request.

amassarwi:
This diff adds two options to CMS admin's search user page filters:
 - `Domain`
 - `Contractor Type`

### Solution description

Extends existing user search page, now in CMS mode the page will have different filtering options, instead of `username` & `email` we have in proposals mode, admin can search CMS users by `contractor type` & `domain`.

This diff uses the isCMS redux flag to determine which filters to display to user, how to map returned result from BE to table and to decide which endpoint to call(`/users` or `/cmsusers`.

### UI Changes Screenshot

image

Closes #1959

### DIFF
--- a/src/containers/User/Search/Search.module.css
+++ b/src/containers/User/Search/Search.module.css
@@ -12,6 +12,16 @@
   margin-top: 0;
 }
 
+.select {
+  height: 4.6rem;
+  width: 24rem;
+  margin: 0 2rem 2rem 0;
+}
+
+.select [class*="customSelect__control"] {
+  height: 97%;
+}
+
 .searchButton {
   max-height: 4.2rem;
 }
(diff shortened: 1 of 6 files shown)


src/containers/User/Search/Search.module.css
```diff
@@ -12,6 +12,16 @@
...
 }
 
+.select {
+  height: 4.6rem;
+  width: 24rem;
+  margin: 0 2rem 2rem 0;
+}
+
+.select [class*="customSelect__control"] {
+  height: 97%;
```
tiagoalvesdulce, line 22: Why 97%?

amassarwi: Good point, replaced with `calc()`.

PR #2044, opened 2020-07-06
fix: Filter abandoned proposals out when starting runoff vote

amassarwi:
As abandoned proposals shouldn't participate in the runoff
vote this diff filters abandoned proposals tokens out when
firing start runoff vote action.

### DIFF
--- a/src/containers/Proposal/Actions/PublicActionsProvider.jsx
+++ b/src/containers/Proposal/Actions/PublicActionsProvider.jsx
@@ -105,15 +105,16 @@ const PublicActionsProvider = ({ children }) => {
       const [submissions] = await onFetchProposalsBatchWithoutState(
         proposal.linkedfrom
       );
-      // Filter abandoned submmsions out.
-      const publicSubmissions = submissions.filter(
-        (prop) => !isAbandonedProposal(prop)
-      );
-      const submissionVotes = publicSubmissions.map(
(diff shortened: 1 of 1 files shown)


src/containers/Proposal/Actions/PublicActionsProvider.jsx
```diff
@@ -104,7 +105,11 @@ const PublicActionsProvider = ({ children }) => {
       const [submissions] = await onFetchProposalsBatchWithoutState(
         proposal.linkedfrom
       );
-      const submissionVotes = submissions.map(
+      // Filter abandoned submmsions out.
+      const publicSubmissions = submissions.filter(
+        (prop) => !isAbandonedProposal(prop)
+      );
+      const submissionVotes = publicSubmissions.map(
```
tiagoalvesdulce, line 112: There's no need to go through the array 2 times. We can use a `flatMap` to accomplish the same thing. See [Array.prototype.flatMap](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap#For_adding_and_removing_items_during_a_map).

amassarwi: Nice one, will do

xaur:
Does the backend reject abandoned tokens too when starting the runoff vote?

PR #2041, opened 2020-07-02
[cms] feat: switch proposal token to proposal name on invoice lineitem

victorgcramos:
This diff Closes #2016 and changes the lineitem selector for proposal token to proposal name.

### Solution description

Creates a `useApprovedProposals` hook to manage approved proposals when needed. Now, invoices send a `batch proposals` request to fetch proposals according to the lineitem's token.

On `InvoiceForm`, when selecting the proposal token lineitem, we do a `tokeninventory` request to fetch all proposals tokens, and then `batch` the approved ones to get the corresponding proposal.

#### Use Cases

Now, `useApprovedProposals` hook can be used when fetching approved proposals is needed.

### Dependencies

Depends on https://github.com/decred/politeia/pull/1242

### UI Changes Screenshot

before:
Screen Shot 2020-07-02 at 7 36 19 PM

after:
Screen Shot 2020-07-02 at 7 33 51 PM

PR #2046, opened 2020-07-07
[CMS] feat: add proposalsowned and proposalbilling

tiagoalvesdulce:
This PR adds the `Proposals Owned` tab if the user owns proposals and proposal billing details.

### Solution description

Added a new tab and `proposals/billing` endpoint.

### UI Changes Screenshot

Screen Shot 2020-07-08 at 15 08 11

Screen Shot 2020-07-07 at 18 00 53

### Dependencies

Depends on decred/politeia#1242 and https://github.com/decred/politeiagui/pull/2041

### Question

~What's the best option to get the proposal title? I'm currently showing the proposal token returned by `proposalsowned`. Should we hit `batchProposals` on the FE or should the BE return an object of the format: `{title, token}` for `proposalsowned` instead of just the `token` string?~


alexlyp:
Needs a rebase, then should be good to go

src/containers/User/Detail/ProposalsOwned/hooks.js
```diff
@@ -0,0 +1,23 @@
+import * as act from "src/actions";
+import * as sel from "src/selectors";
+import { useSelector, useAction } from "src/redux";
+import { useMemo } from "react";
+
+export function useProposalsOwnedBilling(token) {
+  const billingInfoSelector = useMemo(
```
thi4go, line 7: Do we need a `useMemo` here? Doesn't the `createSelector` from `reselect` already memoizes the value seleciton?

tiagoalvesdulce: here the useMemo is preventing unnecessary function calls if the token is the same. Not needed but not bad. The `useSelector` memoizes the value if passed a selector, we are passing a function that returns a selector.

tiagoalvesdulce:
Thanks for the review! I agree with the go back, will add it. The caching can be done in another PR.

src/containers/Invoice/Detail/Detail.jsx
```diff
@@ -1,24 +1,56 @@
...
   const { invoice, loading } = useInvoice(invoiceToken);
-  const tokens = useMemo(() => getProposalsTokensFromInvoice(invoice), [
-    invoice
-  ]);
+  const tokens = useMemo(
+    () =>
+      invoice &&
+      invoice.input &&
+      invoice.input.lineitems &&
+      flow(
```
amassarwi, line 27: 👍

PR #2054, opened 2020-07-11
[CMS] Default admin invoice filters to unreviewed

### LINKED ISSUE #2012: [cms] Admin Invoices page should default to "Unreviewed"
It currently defaults to "All" but admin review would be much better if it were to always show just unreviewed on page load.

thi4go:
> Quick explanation about the issue that the PR is aiming to solve

This diff defaults the admin invoice filters to unreviewed, to facilitate the review process.

Closes #2012 

### Solution description

Indicate to `InvoiceFilterForm` if the renderer is the admin page or not. Also adding a local storage key which sets the preference for the admin invoice filter. Before, both the admin invoices page and user invoices page used the same local storage entry for specifying the filter preferences.

### DIFF
--- a/src/components/InvoiceFilters/InvoiceFilterForm.jsx
+++ b/src/components/InvoiceFilters/InvoiceFilterForm.jsx
@@ -121,7 +121,7 @@ const OnChangeFiltersModifier = ({ formikProps, isAdminPage = false }) => {
 
   useEffect(() => {
     if (all) {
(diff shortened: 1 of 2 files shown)


alexlyp:
Hrm something doesn't seem to be working correctly with this PR.  Unreviewed is selected, and I can click some of the other types, but ALL doesn't not allowed to be clicked at all.

thi4go:
Hey @alexlyp was passing a wrong parameter for the filter modifier. Fixed it with the last commit :+1:

PR #2055, opened 2020-07-13
[CMS] Set owned proposals on manage cms user

### LINKED ISSUE #2020: [cms] Allow administor to set a user as proposal owner
In the manage user tab, an administrator should be able to edit the proposal owned field.  That field should work just like the supervisor id field except with approved proposal titles populating that dropdown instead.

thi4go:
This diff adds the option to set owned proposals on the manage cms user tab.

Depends on #2041
Closes #2020

### Solution description

Add proposals owned field to the manage cms form. Modifies calls to manage cms user endpoint to set the corresponding field correctly.

### UI Changes Screenshot

ownedprops


src/containers/User/Detail/ManageContractor/ManageDccForm.jsx
```diff
@@ -46,14 +52,39 @@ const ManageDccForm = ({ onUpdate, user }) => {
...
+        : [],
+    [proposals]
+  );
+  const initialOwnedProposals = useMemo(
+    () =>
+      proposalsOptions.filter((option) =>
+        proposalsowned.some((token) => token === option.value)
+      ),
+    [proposalsOptions, proposalsowned]
+  );
```
tiagoalvesdulce, line 77: These are examples of bad usage of `useMemo`, you are including values that always change reference in your dep list (array types, keep in mind that `[] !== []`). Please take some time to read [this great article about the using of useMemo and useCallback](https://kentcdodds.com/blog/usememo-and-usecallback). TBH I think a `for ... of` here would improve readability a lot, you can even put it in a function to call inside this component. Something like:

```
const getInitialAndOptionsArrays = (proposals, proposalsowned) => {
  const proposalsOptions = [];
  const initialOwnedProposals = [];
  for (let {
    name,
    censorshiprecord: { token },
  } of proposals) {
    const newVal = {
      label: name,
      value: token,
    };
    if (proposalsowned.includes(token)) { // includes from lodash
      initialOwnedProposals.push(newVal);
    } else {
      proposalsOptions.push(newVal);
    }
  }
  return {
    initialOwnedProposals,
    proposalsOptions
  }
}
```

src/containers/User/Detail/ManageContractor/helpers.js
```diff
@@ -1,3 +1,5 @@
+import { get } from "lodash";
```
tiagoalvesdulce, line 1: Favor default imports when using lodash to reduce bundle size.

src/containers/User/Detail/ManageContractor/ManageDccForm.jsx
thi4go: Got it mate, very nice article about the useMemo and useCallback hooks. So in this situation, since we are dealing with non-primitive values (array), the options and initial values would always get rerendered, even with the useMemo hook and the same exact dep list, since `[] !== []`. This is a very interesting subject; having these optimization hooks available can easily be blown out of proportion with unnecessary optimizations, making the code more verbose, with nothing to be really gained, and in some cases even adding more runtime cost to react. Appreciate the linked article mate.

Will soon push a fix for this. think i'll also do the above with the supervisors memoizations, since they are also in a bad usage format.

thi4go: Updated @tiagoalvesdulce . Went with the above approach for both owned proposals and supervisors. Both had the memoization misusage.

PR #2061, opened 2020-07-14
fix: properly include exchangerate to the invoice draft data

### LINKED ISSUE #2056: [cms] Exchange rate error when submitting a draft
CMS is giving a exchange rate error when we try to submit draft invoices. Same exact invoice gets submitted propertly when not in draft mode, and gives an error when submitting on draft. see below

Screenshot from 2020-07-13 12-52-35

tiagoalvesdulce:
Fixes #2056.

### Solution description

The exchangerate key was left out by mistake in the original implementation. To fix it I improved the code to be more generalist. All keys added to the Formik form will be stored in the localStorage when a draft is saved.


tiagoalvesdulce:
image

PR #2079, opened 2020-07-25
feat: Add `quiesce` mode

### LINKED ISSUE #2078: Support `quiesce` mode
`Quiesce` mode is being added to the BE, see https://github.com/decred/politeia/pull/1253, to give the admins a way to switch the BE to readonly mode
that they could backup politeia safely.

BE will be emitting a websokcet event each time an admin sets new `quiesce` mode value, we would like to add a redux middleware as described in [this article](https://dev.to/aduranil/how-to-use-websockets-with-redux-a-step-by-step-guide-to-writing-understanding-connecting-socket-middleware-to-your-project-km3) to be able to receive the WS notification from BE.

Credits to @tiagoalvesdulce for the approach and article.

amassarwi:
This commit adds `quiesce` mode, which will be toggled from BE via websocket event. 
If quiesced politeia will run in readonly mode and would block all writing commands.

### Solution description

This commit adds redux middleware to support websocket as described in [this article](https://dev.to/aduranil/how-to-use-websockets-with-redux-a-step-by-step-guide-to-writing-understanding-connecting-socket-middleware-to-your-project-km3).
It subscribes to `quiesce` websokcet event and toggles `quiesce` mode according to the value in the websocket event payload. 

**Note:**
 - Use `127.0.0.1` in your browser instead of `localhost` as the BE rejected the WS connection from localhost
 - As we use `wss://` we need to confirm the WS route in browser, so hit `https://127.0.0.1/v1/aws` and confirm the connection, only then the browser will establish the WS connection successfully.
 
### Dependencies

Depends on https://github.com/decred/politeia/pull/1253
Closes #2078 

### UI Changes Screenshot

image

image

image


src/actions/api.js
```diff
@@ -54,6 +54,41 @@ export const onRequestMe = () => (dispatch) => {
     });
 };
 
+export const onWsConnect = (host) => (dispatch) =>
```
tiagoalvesdulce, line 57: here I think we can use only one function: `onWs` and pass an action as a param: `connect, connecting, etc.` then map Ws actions with the param string: `wsActions = {"connect": act.WS_CONNECT, ...}` and dispatch like this: `dispatch(wsActions[action]({host))`

src/helpers.js
```diff
@@ -152,6 +152,7 @@ export const getHumanReadableError = (errorCode, errorContext = []) => {
     74: `Invalid proposal linkby. ${errorContext[0]}`,
     75: `Invalid runoff vote. ${errorContext[0]}`,
     76: `Wrong proposal type. ${errorContext[0]}`,
+    77: "Politeia is currently operating in readonly mode, please try again later",
```
tiagoalvesdulce, line 155: 👍

src/redux/socketMiddleware.js
```diff
@@ -0,0 +1,83 @@
+import * as act from "src/actions";
+
+const socketMiddleware = () => {
```
tiagoalvesdulce, line 3: well done!

src/redux/socketMiddleware.js
```diff
@@ -0,0 +1,83 @@
...
+          store.dispatch(act.toggleQuiesceMode(payload.quiesce));
+          break;
+        default:
+      }
+      commandRecieved = "";
+    }
+  };
+
+  // the middleware part of this function
+  return (store) => (next) => (action) => {
```
tiagoalvesdulce, line 47: would be nice if you could name this func for stack trace.

amassarwi: good point mate, will do!

src/redux/socketMiddleware.js
amassarwi: 🥇

src/actions/api.js
amassarwi: I Like that 🌮

amassarwi:
@tiagoalvesdulce thanks for the review mate, will address once the BE PR is in.

victorgcramos:
Code looks really good! Congrats!

amassarwi:
irrelevant

PR #2051, opened 2020-07-09
[cms] Add domain invoices page for all users

amassarwi:
This diff updates the invoices list `admin/invoices` to `/invoices` as it now 
available to normal cms contractors, the invoices page content should change 
according to logged in user & should display admins only content (e.g invite 
contractor, payouts, etc.)

Contractors shouldn't see sensitive info like hour rate & dcr paid of each other. 

### Solution description
Tweaked existing route/page, and added some user type specific logic

### Dependencies
Need  https://github.com/decred/politeia/pull/1222
Closes #2014 

### Screenshot
image

PR #2093, opened 2020-08-03
fix: change admin invoices route

victorgcramos:
since https://github.com/decred/politeia/pull/1222 got in, we no longer have the `admin/invoices` route. This commit fixes it.


amassarwi:
Hey mate thanks for this PR, I already made this change in #2051 which need testing and maybe some last touches

victorgcramos:
@amassarwi thanks for the feedback! I think this could get in first to avoid cms issues.

PR #2112, opened 2020-08-17
fix: improve boolean logic to verify proposal type

### LINKED ISSUE #2111: [RFP] Missing RFP label on newly created RFP proposal
missingrfpfb

tiagoalvesdulce:
Closes #2111 

### UI Changes Screenshot

**Before:** see issue.
**After:**

fixrfp

### DIFF
--- a/src/components/Proposal/Proposal.jsx
+++ b/src/components/Proposal/Proposal.jsx
@@ -13,6 +13,7 @@ import ModalSearchVotes from "../ModalSearchVotes";
 import RecordWrapper from "../RecordWrapper";
 import IconButton from "src/components/IconButton";
 import { getProposalStatusTagProps, getStatusBarData } from "./helpers";
+import { PROPOSAL_TYPE_RFP, PROPOSAL_TYPE_RFP_SUBMISSION } from "src/constants";
 import {
   getMarkdownContent,
   getVotesReceived,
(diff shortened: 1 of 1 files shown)


src/components/Proposal/Proposal.jsx
```diff
@@ -203,7 +205,7 @@ const Proposal = React.memo(function Proposal({
               subtitle={
                 <Subtitle>
                   <Author username={username} url={authorURL} />
-                  {isRfp && (
+                  {isRfp && linkby && (
```
amassarwi, line 208: Why is `linkby` needed again here ? we have already in `isRfp` definition ? 🤔

tiagoalvesdulce: We need to verify if `linkby` is truthy in order to pass it to `<Event />`. Here:

https://github.com/decred/politeiagui/blob/e95dac9a11c8dc7c05bf71c33fccd9b15b7d0176/src/components/Proposal/Proposal.jsx#L209

We can have a proposal without a `linkby` in the case described by the issue (newly created proposal). As soon as we fetch this proposal info, the linkby will be populated.

tiagoalvesdulce:
@victorgcramos answered as a reply to @amassarwi's comment.

src/components/Proposal/Proposal.jsx
tiagoalvesdulce: Regarding the title, I opened the PR from the CLI and didn't notice it was so damn long lol. Sorry about that.

victorgcramos:
Fair enough.

src/components/Proposal/Proposal.jsx
amassarwi: Makes sense, thanks for clarifying.

PR #2121, opened 2020-08-25
fix: improve proposal credits workflow

### LINKED ISSUE #2108: Proposal credit workflow bugs
- The modal "Waiting for payment" tag does not update when the tx is registered.  You can see the tx show up in the payments list in the background, but the modal still displays the "Waiting for payment" tag.

Screenshot from 2020-08-14 10-27-31

- If you refresh the modal once the tx is registered, the tag will say "Waiting for confirmations" and will still be the same yellow color as the "Waiting for payment" tag.  I found this to be confusing from a UX standpoint for two reason. Its the same color as the previous "waiting for payment" tag so its hard to notice the change and it's different from the tag being shown in the payments list.  Seems like the modal tag and the list tags should be consistent.

Screenshot from 2020-08-14 07-55-07

- Purchase more than one proposal credit (ex. 4). The modal will say "Send 0.4 dcr to...".  Close the modal and open it back it.  The modal now says "Send 0.1 dcr to..." when it should still say "Send 0.4 dcr to...".

- If you click the "Purchase more" button while the tx is pending then it opens correctly.  Once the tx has been confirmed, clicking the "Purchase more" button doesn't do anything.  You need to refresh the page to get the modal to open properly again.

victorgcramos:
This diff closes #2108 issue that reported a weird behavior on proposal
credits workflow.

### Solution description

The addition of `useProposalCreditsPaymentInfo` hook to improve the data
flow on proposal credits payment status. Previously, props were static
and were passed only when modal was opened. To solve this, the hook
returns the status info and the polling credits info.

### DIFF
--- a/src/actions/external_api.js
+++ b/src/actions/external_api.js
@@ -6,10 +6,10 @@ export const payWithFaucet = (address, amount, userid) => (dispatch) => {
   return external_api
     .payWithFaucet(address, amount)
     .then((json) => {
-      if (json.Error) {
-        return dispatch(
-          act.RECEIVE_PAYWALL_PAYMENT_WITH_FAUCET(null, new Error(json.Error))
-        );
+      if (json.error) {
+        const error = new Error(json.error);
+        dispatch(act.RECEIVE_PAYWALL_PAYMENT_WITH_FAUCET(null, error));
+        throw error;
       }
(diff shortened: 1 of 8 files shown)


amassarwi (approved):
tACK 👏

PR #2127, opened 2020-09-04
[totp] feat: set and verify tools

### LINKED ISSUE #2120: totp: Add Set/Verify TOTP in account area
This will be available in both cms and proposals.

This is the basic functionality for the Set and Verify TOTP:
1) A user requests to SetTOTP (can be a button) on a new accoun tab.
2) www returns a secret and an image (QR code png in bytes).
3) Display the QR code and add a Verification Code field below.
4) User enters a valid code from their TOTP client application and clicks verify.
5) Success message is shown if valid and they are all set for TOTP.

victorgcramos:
This PR Closes #2120 

This diff adds the SET and VERIFY TOTP (**T**ime-based **O**ne-**T**ime **P**assword) features. Now users will be able to set the Two-Factor Authentication (2FA) on their accounts , which will be required to perform the login action once it's enabled.

### How to set the 2FA?

This tutorial was made using the [FreeOTP](https://github.com/freeotp/freeotp-android) app, an open source two-factor authentication application for systems utilizing one-time password protocols.

1. 👤  On your **Account** page, go to the `Two-Factor Authentication` tab.
  Before setting the 2FA on your account, please **remember to save the backup key**. This will be required in case you lost your authenticator. 
    > The backup key should be treated the same way you treat a password and should not be stored digitally.

2. 🤳 **Scan** the QR Code with your authentication app. In this case, we'll be using FreeOTP.
    
3. 📲    **Verify** the 6-digits code generated by the authenticator.

#### Example:
totp-set-cms-test
> The 2FA can also be enabled on CMS

😄 🔐  Done! You have enabled the 2FA on your account 

### How to disable the 2FA

After setting the TOTP, you can disable it on Account > Two-Factor Authentication.

Just fill in the blanks with the 6-digits code generated by your authenticator

#### Example:
totp-reset-cms-test-2

### How to recover the authentication code.

Once you have the backup key saved, you can recover it by generating a new QR Code from it. In this case, I'll be using the [FreeOTP QR Code Generator](https://freeotp.github.io/qrcode.html):

1. Change the algorithm to SHA1 instead of SHA256
2. Set it to **Timeout**
3. On **Account** field, type your username
4. On **Secret (Base 32)** insert the saved backup key

    You should see something like this:
    totp-free-otp-example-2

5. Scan the generated QR Code. Now you have recovered it and you can use it to login to your account.


alexlyp (approved):
tACK works well and looks pretty good.

victorgcramos:
Now, since TOTP is required on login for enabled users, I made a couple of changes on the instructions, added some confirmation modals, and improved the TOTP setup flow.

Here's a quick preview:
Screen Shot 2020-12-07 at 6 10 49 PM

Screen Shot 2020-12-07 at 6 18 30 PM

Screen Shot 2020-12-07 at 6 19 09 PM
Screen Shot 2020-12-07 at 6 23 06 PM
Screen Shot 2020-12-07 at 6 19 41 PM

Screen Shot 2020-12-07 at 6 20 26 PM
Screen Shot 2020-12-07 at 6 20 34 PM

alexlyp:
works great on login!

only suggestion i have is to have a definite cursor or highlight on mousing over an already populated digit.  

Currently once a digit is populated the blue line doesn't change to it's hard to know if you have that digit selected.

Maybe change it to a slightly darker blue tone when already populated and hovered?

src/containers/User/Totp/Set.jsx
```diff
@@ -0,0 +1,52 @@
...
+  !isAlreadySet ? (
+    <div className="margin-bottom-m">
+      <H2>Set Two-Factor Authentication</H2>
+      <div className={styles.setContainer}>
+        <div className={styles.qrcode}>
+          {totp.image && (
+            <img src={`data:image/png;base64, ${totp.image}`} alt={totp.key} />
+          )}
+        </div>
+        {totp && (
```
tiagoalvesdulce, line 15: this is boolean expression is a bit unclear since because if `SetTotp` receives `undefined`, it will default to `{}` and be `true` which is not very clear for the reader. You should be more explicit here saying: `totp !== null` or removing the default `{}` and moving the `totp &&` expression up.

src/containers/User/Totp/Totp.jsx
```diff
@@ -0,0 +1,84 @@
...
+      cb();
+    });
+
+  const handleVerifyTotp = (code, cb) => () =>
+    onVerifyTotp(code).then(() => {
+      setIsTotpSet(true);
+      cb();
+    });
+
+  const handleOpenTOTPModal = (isReset = true) => (code, cb) => {
```
tiagoalvesdulce, line 30: This param could be renamed to `shouldShowResetInstructions` or `isTotpSet`.

src/containers/User/Totp/Verify.jsx
```diff
@@ -0,0 +1,79 @@
...
+            {buttonLabel}
+          </Button>
+        )}
+      </div>
+    </div>
+  );
+};
+
+VerifyTotp.propTypes = {
+  onVerify: PropTypes.func,
```
tiagoalvesdulce, line 61: I think this component doesn't make sense unless the onVerify function is required

src/components/DigitsInput/DigitsInput.jsx
```diff
@@ -0,0 +1,90 @@
...
+        })}
+      </div>
+    </>
+  );
+};
+
+DigitsInput.propTypes = {
+  length: PropTypes.number,
+  onChange: PropTypes.func,
+  onFill: PropTypes.func
```
tiagoalvesdulce, line 81: `onFill` is never used and `className` and `code` are used but not defined in the propTypes.

tiagoalvesdulce (commented):
Good job, left a few comments.

src/containers/User/hooks/useTotp.js
```diff
@@ -0,0 +1,61 @@
...
+import { useAction, useSelector } from "src/redux";
+import useFetchMachine from "src/hooks/utils/useFetchMachine";
+import isEqual from "lodash/isEqual";
+
+const ERROR_CODE_INVALID_PARAMS = 77;
+
+const isInvalidParamsError = (error) =>
+  error && error.errorCode === ERROR_CODE_INVALID_PARAMS;
+
+export default function useTotp() {
```
thi4go, line 12: sweet hook

src/components/DigitsInput/DigitsInput.jsx
```diff
@@ -0,0 +1,90 @@
+import React, { useState, useRef, useEffect } from "react";
+import PropTypes from "prop-types";
+import { NumberInput, classNames } from "pi-ui";
+import styles from "./DigitsInput.module.css";
+
+const getCodeArray = (code = "", length) => {
```
thi4go, line 6: would be great to have a comment that explains what this function accomplishes for the component as a whole

thi4go (commented):
Hey mate functionality working great, nice job. Got just few comment.

As of now, I have to fail a login first to enable totp code insertion, and then do a successful login. Not sure the best way to go here, but I thought of two things; we can store this info on the user's browser (not a fan of this), or we could add a checkbox on the login form that shows/hides the totp input part.

src/containers/User/Totp/Set.jsx
victorgcramos: I guess we don't need this `totp` checker anymore. The parent component already forces the loading state when totp is undefined. I think removing it is okay.

src/containers/User/Totp/Verify.jsx
victorgcramos: I agree!

src/containers/User/hooks/useTotp.js
victorgcramos: thanks bro!

tiagoalvesdulce (changes requested):
Code is looking pretty good.

Some points:

1. These should be submitted when hitting enter.

Screen Shot 2021-01-20 at 12 45 29

Screen Shot 2021-01-20 at 12 46 44

2. The switch should be focused after hitting tab on password (`tabindex`)

Screen Shot 2021-01-20 at 12 48 16

3. This loading indicator should be centered

2facauth

4. Question: Can you edit the recommended logo in the TOTP authenticator app?

image

victorgcramos:
@tiagoalvesdulce updated the code to include the requested changes. About item 4, I don't think we can change that. From what I could see, this is up to the TOTP App, not the code itself.

After taking a look at https://github.com/pquerna/otp (the one we use on backend), I couldn't figure a way out to handle this default image.

tiagoalvesdulce (approved):
Looking great!

PR #2147, opened 2020-09-30
fix: force markdown's wide 'code' tag to breakline

### LINKED ISSUE #2136: [firefox] width longer than 1385px breaks the design
It looks like that:

image

### LINKED ISSUE #2138: Downvote button not showing
The downvote button gets cut off on a 14'' screen.  I'm using Firefox.

Screenshot from 2020-09-23 14-30-52

amassarwi:
This diff fixes the display issues reported on #2136 & #2138, usage of wide 'code' inside
the markdown renderer was causing the issue.

### Solution description

Force 'code' tag with wide text to breakline in markdown.

### Dependencies

Fixes #2136 
Fixes #2138 

### UI Changes Screenshot

Before:
image

After:
image

### DIFF
--- a/src/components/Markdown/styles.css
+++ b/src/components/Markdown/styles.css
@@ -446,7 +446,7 @@
   font-size: 100%;
   margin: 0;
   padding: 0;
-  white-space: pre;
+  white-space: pre-wrap;
   word-break: normal;
 }
 
@@ -465,7 +465,6 @@
   border-radius: 3px;
   font-size: 85%;
   line-height: 1.45;
-  overflow: scroll;
   padding: 16px;
   max-width: 100%;
 }
@@ -481,8 +480,6 @@
   display: inline;
   line-height: inherit;
   margin: 0;
-  max-width: auto;
-  overflow: visible;
   padding: 0;
   word-wrap: normal;
 }


victorgcramos (commented):
Hey @amassarwi thanks for the PR. Very clean (as always) 👍. 

Though, I think the Code tag should not be breaking lines by default, since it's a raw text format, and in most of the cases, the screen width would cause some weird code breaking like image below. I couldn't think of a way to deal with it, so I'd suggest to add some way to not break those lines when I don't want to.

Screen Shot 2020-10-11 at 11 31 21 AM


Like I said, this is just my opinion. I'd like to hear @tiagoalvesdulce's and @thi4go's thoughts on this one.

tiagoalvesdulce (approved):

xaur:
I agree with @victorgcramos, code blocks should _not_ be wrapped and the After [screenshot](https://user-images.githubusercontent.com/10324528/94680537-8df3b180-032a-11eb-8f0e-7338ebe79267.png) looks wrong. Horizontal scrollbar is expected behavior.

The real problem with that [proposal](https://proposals.decred.org/proposals/f0a00d5) was not that code blocks are too wide, but that code blocks were used in the first place. It should have been quote blocks (`> quote...`), not code blocks.

PR #2142, opened 2020-09-26
[tlog] Adjust app according to the new pi API specs

amassarwi:
This diff adjusts the API routes and their payloads according to the new API 
implemented in https://github.com/decred/politeia/pull/1180

### Main changes:
- Adjust API according to new specs.
- Add separate routes for vetted & unvetted proposal (for details & edit views) as the state is needed to communicate with BE.
- Make it possible to censor public not voting proposals
- Add comments to unvetted propsoals
- Rename UnvettedList to AdminList as it now includes `Censored Vetted` proposal in addition to old `Censored Unvetted` & `Unvetted` proposals. 

*Note:* Currently user's proposals page isn't functioning, as the BE work for it isn't ready yet.

### Dependencies
Needs https://github.com/decred/politeia/pull/1180

PR #2153, opened 2020-10-11
diff: improve diff loading state and minor fixes

victorgcramos:
Diff processing is computationally expensive, and this was causing some UX issues.

The first noticed issue was that Diffs for **version 1** was not being displayed, and the second was that we needed to add some visual feedback for the user to show him/her that the page wasn't actually broken.

### Solution description
This PR adds a `DiffLoader` component to indicate the loading state of some proposal or invoice diff, and fixes the version 1 bug.

### UI Changes Screenshot

Screen Shot 2020-10-11 at 11 06 32 AM

### DIFF
--- a/src/components/ModalDiff/ModalDiff.module.css
+++ b/src/components/ModalDiff/ModalDiff.module.css
@@ -27,8 +27,15 @@
   overflow: scroll;
 }
 
+.loaderWrapper {
+  width: 920px;
+}
+
 @media screen and (max-width: 560px) {
   .diffTabContent {
     max-height: calc(70vh);
+  }
+  .loaderWrapper {
+    width: 100%;
   }
 }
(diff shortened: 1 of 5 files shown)


tiagoalvesdulce (approved):

amassarwi:
I would suggest we keep this for after tlog, this component is going to be heavily refactored in #2142 as we now use the proposal batch to fetch proposals as we deprecated the proposal details route.

xaur:
Is this gradient background animated?

victorgcramos:
@xaur sure!

xaur:
@victorgcramos thanks. The reason I ask is that these gray CSS-animated "loading" backgrounds tend to be CPU hogs. They spin up laptop fans even if javascript is disabled and you have to add CSS filter rules to prevent that abuse.

Now that I am typing this I realize that Pi GUI currently needs js enabled anyway. So the content will eventually load, replace the animation and stop spinning the CPU (opposed to spinning it forever because js is disabled and the content will never load). Making it not as bad as I originally thought. But still I would think twice about using these background animations to minimize resource use.

Issue #2143, opened 2020-09-27
Inline markdown images


lukebp:
This issue is meant to serve as a discussion on the topic of adding the ability to reference and render proposal image attachments inline in the proposal markdown file.

Is this feasable?
How would it work?
What work would be required?

tiagoalvesdulce:
Hey, that's a good idea.

**Is this feasible?**
Yes, it is.

**How would it work?**
We have to allow users to upload images through the markdown editor and render them with the markdown renderer in the right placement.

**What work would be required?**
React-mde (our editor) currently allows image upload (https://github.com/andrerpena/react-mde/issues/189) and react-markdown (our renderer) allows the rendering of images (https://github.com/rexxars/react-markdown#node-types). We have to make sure we use the versions that support images and adapt our code. More importantly, we have to guarantee that we are applying the XSS filters to the markdown rendered images.

victorgcramos:
I agree with @tiagoalvesdulce.

lukebp:
Sounds like a plan.  Let's get this going.

Issue #681, opened 2018-10-11
Markdown Issues (external link warning bypass, image loading in markdown, force logout)


degeri:
So I spent the last few days reviewing markdown specifically this https://daringfireball.net/projects/markdown/syntax

There are some edge cases where our filtering is not working cause we failed to fully account for all the different types of markdown code's.

Both seem to be an issue with labeling in markdown links and images.

See examples of "labeling" under. 
https://daringfireball.net/projects/markdown/syntax#link
https://daringfireball.net/projects/markdown/syntax#img

**Point A.**

https://test-proposals.decred.org/proposals/eb24bd6170abcc2eaee9bb1da90afbeb695f3c9b2bc7e68b0ca68f528fb70736 (See comment Click me no warning.)

_POC 1:_

If we make a link with:

[Test][link]
[link]: http://google.com/

or 

[Google][]
[Google]: http://google.com/

the external link warning is no longer working and taking the user directly to the site http://google.com/

**Point B.**

https://test-proposals.decred.org/proposals/eb24bd6170abcc2eaee9bb1da90afbeb695f3c9b2bc7e68b0ca68f528fb70736 (See comment This bad boy can fit so many images.)

We are able to embed image using 

_POC 2:_

![Something][test]
[test]: https://test-proposals.decred.org/static/media/spritesheet.23cd19a2.png

or

![lol][]
[lol]: https://test-proposals.decred.org/static/media/spritesheet.23cd19a2.png

Since we are having CSP we are having some protection. Only "self" (samedomain) and "data:"  (get converted to x-data: by xss filter) images are allowed. But we should still fix this in the code level. 

Without CSP the following would also work:

![Something][test]
[test]: https://upload.wikimedia.org/wikipedia/commons/e/e9/Imgur_logo.svg

![image](https://user-images.githubusercontent.com/41839293/46814446-59316d00-cd96-11e8-9a3e-a45cddeda30a.png)

I spent some more time trying to chain some vulnerabilities to make image loading more critical issue and came up with this.

There are some insecure pages like "https://test-proposals.decred.org/api/v1/logout"  (Will open separate issue for this)

Lets say a user embeds the below POC as an image. Even though it is not a valid image the browser sends a request with valid user cookies and the user will be logged out.

_POC 3:_
![You are logged out!!][id]
[id]: https://test-proposals.decred.org/api/v1/logout

Note:
[...]

PR #2158, opened 2020-10-13
Inline markdown images

tiagoalvesdulce:
Closes https://github.com/decred/politeiagui/issues/2143

### Solution description

This diff allows the user to reference images in the markdown. In order to do that I had to bump `react-mde` package to `10.3.0-alpha1` and I created the new `customSaveImageCommand`.

#### Use Cases

You should be able to copy and paste and to drag and drop images in the markdown editor. Also, there's a new `image` icon that can be used to include a new image.

Screen_Shot_2020-10-13_at_16_38_47

### UI Changes Screenshot

localhost_3000_ (1)

**TODO**

- [x] Do not show the images referenced in the markdown in the image gallery
- [ ] ~Fix proposal diff~
- [x] Update image attach modal copy to remove .txt
- [x] Don't allow users to attach .txt

The way we do proposals DIff right now is really difficult to render images in place with a good performance.


victorgcramos (commented):
Hey @tiagoalvesdulce thanks for the PR. Code is looking really good as always!

One thing I noticed is that MIME types validation should be included in order to avoid image submission errors. Besides that, It's all good for this first version.

degeri:
we had alot of security issues due to rendering images thats why we chose to not allow it during launch. Allowing someone to let the user load content is a risk.

eg:

https://github.com/decred/politeiagui/issues/681

tiagoalvesdulce:
I fixed the proposal editing route, improved the error message for images with mime types not allowed (like `svg`) and prevent usage of images in the comments. I'll be completing the itens of the todo list in the description. Let me know if I need to add something else.

degeri:
images in comments is still allowed its just not showing the option. We need to completely disable images from being rendered in comments. 

image

tiagoalvesdulce:
Sorry, I forgot to change from image renderer to link renderer on comments. Should be good now. Only need to polish the Diff and the PR will be ready.

degeri:
Checked. Looks good.

victorgcramos (approved):
LGTM!

src/components/Markdown/helpers.js
```diff
@@ -80,30 +62,51 @@ const LinkRenderer = ({ url, children }) => {
   );
 };
 
-const imageHandler = ({ src, alt }) => {
-  return <LinkRenderer url={src}>{alt}</LinkRenderer>;
+// Use external link renderer when images are not allowed
+const imageHandler = (renderImages) => ({ src, alt }) => {
+  return renderImages ? (
```
amassarwi, line 67: No need for the return here

src/components/MarkdownEditor/commands.js
```diff
@@ -0,0 +1,61 @@
...
+  image: imageSVG
+};
+
+/**
+ * Returns the React node to be displayed for a given command name.
+ * @param {string} commandName
+ * @returns {Object} React node
+ */
+export const getCommandIcon = (commandName) => {
+  return <img alt={commandName} src={commandsIcons[commandName]} />;
```
amassarwi, line 60: Ditch return

src/components/Proposal/Proposal.jsx
```diff
@@ -144,6 +172,11 @@ const Proposal = React.memo(function Proposal({
   const { themeName } = useTheme();
   const isDarkTheme = themeName === DEFAULT_DARK_THEME_NAME;
 
+  const { text, markdownFiles } = replaceImgDigestWithPayload(
```
amassarwi, line 175: Could we use useMemo here ?

src/components/ProposalForm/ProposalForm.jsx
```diff
@@ -286,9 +360,14 @@ const ProposalFormWrapper = ({
             );
           }
         }
+        const { description, files } = replaceBlobsByDigestsAndGetFiles(
```
amassarwi, line 363: Usememo ?

amassarwi (approved):
Nice work, added few minor suggestions 👍

src/components/ProposalForm/ProposalForm.jsx
tiagoalvesdulce: This is inside an useCallback so I can't use useMemo

amassarwi: opps

PR #2170, opened 2020-10-21
fix: proposals lazy list fetching fixes

### LINKED ISSUE #2169: Bug: Proposals list chunks fetching is broken
While working on tlog I experienced a bug in all proposal lists, currently having 5 proposals _is discussion_ but seeing only the first page/chunk which is 4 proposals.

This bug originating in #2084.

victorgcramos:
This diff fixes #2169 , which describes a bug on lazy list for proposals. The proposals were not being fetched after the first initial batch due to a race condition caused by the way we fetch remaining proposals.

### DIFF
--- a/src/components/RecordsView/RecordsView.jsx
+++ b/src/components/RecordsView/RecordsView.jsx
@@ -1,13 +1,6 @@
-import React, {
-  useState,
-  useEffect,
-  useReducer,
-  useMemo,
-  useCallback
-} from "react";
+import React, { useState, useMemo, useEffect, useCallback } from "react";
 import { Tabs, Tab, useMediaQuery } from "pi-ui";
 import difference from "lodash/difference";
-import union from "lodash/union";
 import LazyList from "src/components/LazyList";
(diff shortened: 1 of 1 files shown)


src/components/RecordsView/RecordsView.jsx
```diff
@@ -63,63 +46,45 @@ const RecordsView = ({
...
   const [index, onSetIndex] = useQueryStringWithIndexValue("tab", 0, tabLabels);
   const tabOption = tabLabels[index];
-
-  const filteredTokens = recordTokensByTab[tabOption] || [];
-
   const isMobileScreen = useMediaQuery("(max-width:560px)");
 
-  const filteredRecords =
-    (records && getRecordsByTabOption(records, filteredTokens)) || [];
+  const [filteredRecords, filteredTokens] = getFilteredRecordsAndToken(
```
amassarwi, line 56: useMemo maybe ?

amassarwi (approved):
tACK - nice work @victorgcramos, ended up with cleaner & readable logic.

Left one suggestion.

PR #2201, opened 2020-11-11
Inline markdown images on drafts

tiagoalvesdulce:
This PR makes possible to save drafts with markdown embedded images


amassarwi (approved):
ACK

PR #2215, opened 2020-11-24
[cms] fix: make user selector searchable on admin invoices

### LINKED ISSUE #2210: cms: Admin Invoices user dropdown
User reported that the User dropdown is not populating so it's not able to select any users to view their invoices.

victorgcramos:
This diff closes #2210, an issue that reported an unexpected behavior on User Search on admin invoices page where users were not being loaded.

### Solution description

The only thing was to pass the `isSearchable` prop to the `SearchSelector` component, and do some minor tweaks in order to make the selector good to go.

### UI Changes
Screen Shot 2020-11-24 at 2 31 56 PM

### DIFF
--- a/src/components/InvoiceFilters/InvoiceFilterForm.jsx
+++ b/src/components/InvoiceFilters/InvoiceFilterForm.jsx
@@ -1,5 +1,5 @@
 import React, { useEffect } from "react";
-import { Formik } from "formik";
+import { Formik, Form } from "formik";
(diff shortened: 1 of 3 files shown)


tiagoalvesdulce (approved):
Code looks good!

In the first click I think we should show all user options and filter while the admin types:
Screen Shot 2020-11-24 at 15 17 09

tiagoalvesdulce (approved):
LGTM

PR #2220, opened 2020-11-25
[cms/header] style: disable 'All Invoices' option for non-contractors

victorgcramos:
Noticed that users without a DCC were able to navigate to `/invoices/all` using the header dropdown. But when the user navigates to it, it shows the following message:
> Error: You are not authorized to perform this action.

This PR adds a DCC checker for NavHeader in order to prevent that behavior.

### UI Changes
Screen Shot 2020-11-25 at 4 34 49 PM

### DIFF
--- a/src/actions/api.js
+++ b/src/actions/api.js
@@ -40,12 +40,15 @@ export const requestApiInfo = (fetchUser = true) => (dispatch) => {
     });
 };
 
-export const onRequestMe = () => (dispatch) => {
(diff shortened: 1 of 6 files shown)


amassarwi:
Thanks for the PR @victorgcramos.
I would suggest we don't show it at all if it's not complex

victorgcramos:
sounds good!

thi4go (approved):
tACK !

amassarwi (approved):

PR #2223, opened 2020-11-26
build: upgrade various deps

dajohi:


tiagoalvesdulce (approved):
LGTM but has some conflicts

PR #2226, opened 2020-11-30
Move logout permanently to account tab

### LINKED ISSUE #2225: Logout: remove modal and add another way to clear browser data
We are showing a Modal when the user clicks on logout in the dropdown:
Screen Shot 2020-11-30 at 15 04 44

This is a bit annoying and the `Logout permanently ` option is rarely used. We can move the option to clear user data and the text explaining it to the `Account` tab.

tiagoalvesdulce:
Closes #2225 

### UI Changes Screenshot

Screen Shot 2020-11-30 at 15 12 29


thi4go (changes requested):
Hey mate, thanks for this PR, this is indeed much better UX for clearing user data.

I suggest we add a confirmation modal or smth similar for the `Clear data` button, so that we can avoid headaches with missclicks. Currently it wipes the data with the first click, which could lead to unintended behaviour for users that are not careful with this.

tiagoalvesdulce:
Good idea, will do

thi4go (approved):
Sweet. LGTM!

amassarwi (approved):

PR #2231, opened 2020-12-02
fix: remove old version of react-mde

### LINKED ISSUE #1993: upgrade react-mde commands
Noticed this after upgrading our `react-mde` package to the latest version. `commands` module is no longer exported, so we need to handle these changes on `MarkdownEditor` https://github.com/decred/politeiagui/blob/450e34fb5b8f60d8c0c83949332dadd44f2006f7/src/components/MarkdownEditor/commands.jsx#L2

tiagoalvesdulce:
Closes #1993


amassarwi (approved):
LGTM

PR #2232, opened 2020-12-02
[cms/invoice] fix: invoice datasheet paste tweaks

victorgcramos:
Some users reported that the pasting experience is not working as expected for the **proposal** lineitem. When some user typed an invalid proposal token (by pasting external datasheet items), a loading icon appears and is displayed until the user selects a valid proposal token, or paste a valid one.

Also, if the user pastes an invalid **type** lineitem, the page breaks.

### Solution description

The solution was simple: just display a message telling if the fields are invalid.

### UI Changes Screenshot
Before:
**Type lineitem**
Screen Shot 2020-12-02 at 5 29 50 PM

**Proposal lineitem**
Screen Shot 2020-12-02 at 5 31 39 PM

After:
Screen Shot 2020-12-02 at 5 25 19 PM

### DIFF
--- a/src/components/InvoiceDatasheet/helpers.js
+++ b/src/components/InvoiceDatasheet/helpers.js
@@ -120,7 +120,7 @@ export const convertLineItemsToGrid = (
           value: newLine.type,
           error: rowErrors && rowErrors.type,
           valueViewer: () =>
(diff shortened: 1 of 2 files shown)


tiagoalvesdulce (approved):
LGTM

Issue #2234, opened 2020-12-05
[cms] Put name first in Invoice tab title


xaur:
#2208 has improved tab title by adding name and month but it missed to place name at the beginning of the string. Try opening a lot of tabs and they all will show `Invoice from`.

I suggest to format tab title as `{name} {month} invoice`.

victorgcramos:
as I mentioned on https://github.com/decred/politeiagui/pull/2238#pullrequestreview-549207907, I think we can remove this "invoice" suffix from the title on invoice list, and keep only on details.

PR #2238, opened 2020-12-09
[cms] fix: reword invoice title

tiagoalvesdulce:
Closes https://github.com/decred/politeiagui/issues/2234


victorgcramos (approved):
LGTM! Thanks for the PR! I think we can remove this "invoice" and leave just the `user - date` format on listing. IMO, we should just use the "invoice" word on details. Just to avoid the redundancy on invoice listing. But it's my opinion.

xaur:
We could live without "invoice" too :)

A combination of username and date makes it obvious that it is an invoice. It may change when we get another entity with similar short text representation, but then we could just add "invoice" back.

victorgcramos (approved):
tACK!

amassarwi (approved):
lgtm

PR #2240, opened 2020-12-10
[cms] fix: endless loading on error in invoice page

### LINKED ISSUE #2204: [cms] Invoice details & edit pages endless loading
When having an invoice with wrong proposal token which causes a 400 error when displaying invoice details, the gui keeps loading forever!

amassarwi:
This diff fixes endless loading indicator when there's an error fetching
a proposal details with a wrong token in invoice page.

Partly fixes #2204 - will keep the issue open as we will need to prevent wrong tokens from being submitted by adding 
a validation for the proposal token column in invoice lineitems.

### DIFF
--- a/src/containers/Invoice/Detail/hooks.js
+++ b/src/containers/Invoice/Detail/hooks.js
@@ -61,7 +61,6 @@ export function useInvoice(invoiceToken) {
           onFetchProposalsBatch(unfetchedProposalsTokens, false)
             .then(() => send(VERIFY))
             .catch((e) => send(REJECT, e));
-          return send(FETCH);
         }
         return send(RESOLVE, { invoice, proposals: values(proposalsByToken) });
       },


victorgcramos:
LGTM! Working as expected. 1 Line diff fixed the issue 🚀

victorgcramos:
But I'd suggest to keep the fetching command to the machine. Since we have a request, we should send it to the loading state, and then display the correct issue. I noticed that the error state is returned by the `proposalsError` attribute, and the issue is that we don't handle it correctly.

victorgcramos (commented):
But I'd suggest to keep the fetching command to the machine. Since we have a request, we should send it to the loading state, and then display the correct issue. I noticed that the error state is returned by the proposalsError attribute, and the issue is that we don't handle it correctly.

alexlyp:
So this appears to allow invoice details to load for the invalid proposal token invoices.   The errors displayed need to be fixed as well.

But most importantly this PR seems to cause working invoices not load their proposal tokens.  I just get an endless loading indicator.

amassarwi:
@alexlyp I'm currently running it and invoices with proper tokens are being loaded without any issues.
I have 2 lineitems and it seems to work as expected.

@victorgcramos I have played with it a bit and I have to say I don't agree here, we should show the invoice details while fetching
the proposals names, as we already showing loading indicator inside the proposal token cell  in each lineitem while looks faster
and smoother than hiding the entire page with the invoice loading indicator.

victorgcramos:
nice! I agree with you @amassarwi! Good point!

alexlyp (approved):

PR #2251, opened 2020-12-14
fix: force DateTooltip's timeAgo to always be on the past

### LINKED ISSUE #2246: Comments from the future
A small thing, but I noticed that my comments will sometimes appear as they're scheduled to appear in future, with "in less than a minute" time description.
image

victorgcramos:
Closes #2246 by forcing timestamps to always be on the past on DateTooltip. This can be caused by the git issue tlog is attempting to solve in the upcoming release.


tiagoalvesdulce (commented):
Looks good but I feel it won't be necessary after tlog, right?

victorgcramos:
Yes, this is temp!

victorgcramos:
Agree. This won't be necessary!

PR #2263, opened 2020-12-17
[cms] fix: endless loading inside proposal token empty cell

amassarwi:
This diff improves the `isLoading` check inside lineitems' proposal token cell, which
used to show loading indicator if no proposals available but that's not true because
we have the edge case where invoice has no proposal tokens which was causing the cell 
to show the loading endlessly.

This diff ensures that the loading indicator displayed only inside cells which have a proposal token to fetch.


tiagoalvesdulce (approved):
I tested and it fixes the issue.

PR #2274, opened 2020-12-28
fix: use cms_logout action for all cms reducers

tiagoalvesdulce:
Set CMS related reducers to default state on logout


amassarwi (approved):
lgtm

PR #2284, opened 2021-01-07
deps: use latest table pagination from pi-ui

### LINKED ISSUE #2219: [cms/user-search] Improve page counting for long list results
This is minor, but I noticed that if we have a very long user list as a result of searching users, the list counting will overflow the screen.

Screen Shot 2020-11-25 at 4 47 58 PM

victorgcramos:
This Fixes #2219 by updating the Table component on pi-ui.

### Dependencies

https://github.com/decred/pi-ui/issues/311

### UI Changes Screenshot

Before:
before

After:
Screen Shot 2021-01-07 at 11 02 32 AM

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -9778,7 +9778,7 @@ performance-now@^2.1.0:
 
 "pi-ui@https://github.com/decred/pi-ui":
   version "1.0.0"
-  resolved "https://github.com/decred/pi-ui#8f5124149a6820726905956e894cd94e82719987"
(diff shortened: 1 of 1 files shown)


amassarwi (commented):
Looks like there are some unintended `yarn.lock` changes.

Have you done `yarn upgrade pi-ui` ? we should always use `yarn add pi-ui` which
adds the latest version of pi-ui without upgrading other deps, not like `yarn upgrade` which updates other deps as well.

tiagoalvesdulce:
This is not needed anymore because of https://github.com/decred/politeiagui/pull/2287

Issue #2265, opened 2020-12-19
[cms] Always reset invoice filters


xaur:
A few times I found myself not noticing new invoices because the filters were enabled.

For example, during a review session I can enable By Date filter on All Invoices page, finish my session and forget about it. Then I open All Invoices a few days later, don't pay attention to enabled filters and miss new invoices. They even persist across logins.

I suggest to reset all filters to their least restrictive state when the user is leaving the page that has filters (All Invoices, My invoices). This applies to all filters (by date, by status, by user).

It is better to be annoyed a bit and having to enable filters again, than to miss stuff.

PR #2285, opened 2021-01-07
feat: do not save filters on localStorage

tiagoalvesdulce:
Some users are missing info because they forget their filters settings so I think we should not persist them between sessions.

Closes https://github.com/decred/politeiagui/issues/2265


victorgcramos:
tACK!

victorgcramos (approved):
tACK

amassarwi (approved):
ACK

PR #2298, opened 2021-01-26
feat: change filters to show as much data as possible by default

### LINKED ISSUE #2296: [cms] Default domain invoice filters to show maximum data
#2265 was partially fixed by #2285 and now domain invoice filters are not stored across page visits.

However, this part was not addressed:

> I suggest to reset all filters to their **least restrictive state** when the user is leaving the page that has filters (All Invoices, My invoices). This applies to all filters (by date, by status, by user).

The main idea here is to help auditors not miss any recent data, by showing as much as possible by default.

Currently when I open Domain Invoices, the filters are set to show only one month of invoices and most importantly, only those in Unreviewed state.

I suggest to change defaults to show 6 months of invoices in any state (the `All` box).

tiagoalvesdulce:
Closes #2296


victorgcramos (approved):
LGTM!

amassarwi (approved):
LGTM

PR #2301, opened 2021-02-09
Revert proposal token input back to text

### LINKED ISSUE #2023: [mobile] votes left field empty after quorum
Before quorum:
image

After quorum:
image

We either need to remove the text or change it to something like "quorum reached" etc.

tiagoalvesdulce:
To allow short tokens we need to adapt the `/proposals/batch` endpoint to accept them.

Closes #2023
Closes #2030


amassarwi (approved):
LGTM.

> To allow short tokens we need to adapt the /proposals/batch endpoint to accept them.

This is done in the tlog PR afair.

PR #2305, opened 2021-02-25
[cms/dcc]: avoid infinite loading state on dcc details

victorgcramos:
This diff fixes the infinite loading issue on DCC details. This happens when you access the dcc details page without fetching the dcc list.

### Solution description
Basically, I used a new `useFetchOnce` hook, which replaces the `useAPIAction` previously used. Now we can control the loading and error states just by using the fetch state machine.

#### Use Cases
Whenever we want to fetch an action only once, we can now use the `useFetchOnce` hook. Now we don't need to memoize the args for fetching, because the state machine forces an action to happen only once, at the initial state.

Example:
```javascript
const [loading, error] = useFetchOnce(action, args, isEnabled);
```

The following diagram represents the useFetchOnce's fetch machine state transitions:

state-machine-fetch-once (1)

### DIFF
--- a/src/containers/DCC/Detail/hooks.js
+++ b/src/containers/DCC/Detail/hooks.js
@@ -2,7 +2,7 @@ import { useMemo } from "react";
 import * as sel from "src/selectors";
 import * as act from "src/actions";
 import { useSelector, useAction } from "src/redux";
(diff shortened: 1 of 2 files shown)


tiagoalvesdulce (approved):
ACK

PR #2313, opened 2021-03-31
fix: bundle/timestamp structure download fix

thi4go:
This diff fixes the comment bundle/timestamp structure, and the votes timestamp structure, available for download from the GUI. It also solves a type error that might occur when votesummary details is not yet loaded.

edit: This PR now also normalizes `proposalVotes` state. We were saving both full token state and short token state, which would lead to inconsistencies on our single source of truth.

### DIFF
--- a/src/containers/Proposal/Download/hooks.js
+++ b/src/containers/Proposal/Download/hooks.js
@@ -26,7 +26,7 @@ export function useDownloadVoteTimestamps(token, votesCount) {
     actions: {
       initial: () => {
         const ts = loadVotesTimestamps(token);
-        const hasProofs = ts?.votes?.reduce((acc, v) => 
+        const hasProofs = ts?.votes?.reduce((acc, v) =>
           acc && v.proofs.length > 0, true
         );
         if (ts?.votes?.length === votesCount && hasProofs) {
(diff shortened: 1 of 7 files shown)


src/containers/Proposal/Download/hooks.js
```diff
@@ -24,46 +26,55 @@ export function useDownloadVoteTimestamps(token, votesCount) {
...
-                setTimestamps([...timestamps, ...resp.votes]);
-              } else {
-                return send(REJECT, "fetching outbound vote pages");
-              }
-              setProgress(((timestamps.length * 100) / votesCount).toFixed(2));
+              setVotes([
+                ...votes,
+                ...resp.votes
+              ]);
+              setProgress(((votes?.length * 100) / votesCount).toFixed(2));
```
tiagoalvesdulce, line 77: here if the `votes` state is not loaded yet you will end with something like: `undefined / integer`, which is NaN. NaN should be avoided. I would say we do something like: `votes = votes ? votes.length : 0`

tiagoalvesdulce (changes requested):

src/containers/Proposal/Download/hooks.js
thi4go: nice catch mate thx. fixed

thi4go: improved the whole flow of setting the download progress. wasn't taking into account the async nature of useState so now using useEffect it is in fact consistent on the updates.

tiagoalvesdulce (approved):

PR #2332, opened 2021-04-08
fix: remove initial conditional fetch

tiagoalvesdulce:
This PR fixes an issue we were having that a newly created proposal was blocking the inventory fetch and the admin proposals list wasn't being populated properly, only showing the new proposal. A page refresh fixed the issue.

**To test this:**  create a new proposal and go to the admin tab without a refresh. You should be able to see all unvetted proposals.


thi4go (approved):
Tested, this diff fixes the issue. Good catch 👍

amassarwi (approved):

PR #2364, opened 2021-05-01
fix: `yarn test`.

amassarwi:


amassarwi:
@tiagoalvesdulce What about commented out file: https://github.com/decred/politeiagui/blob/master/src/lib/tests/api.test.js ?

I guess we need a separate issue for that one? 🤔

tiagoalvesdulce (approved):

tiagoalvesdulce:
Yeah, we can open an issue for that. LGTM

PR #2384, opened 2021-05-08
Fix comments bugs

### LINKED ISSUE #2374: Comment sorting bug
I think the "Top" comment sorting might be backwards.  The comments with the most votes are displayed on the bottom while the comments with no votes are displayed on the top.

https://proposals.decred.org/record/91cfcc8?scrollToComments=true&sort=Top

### LINKED ISSUE #2375: Own comment votes displayed incorrectly
Scenario 1: Open proposal page and upvote a comment that I did not upvote before.

Vote buttons change from

to

i.e. the count is incremented correctly but the thumbs up icon does not change to solid black.

If I click again the count returns back to 2 as expected.

Scenario 2: Open proposal page and upvote a comment that I already upvoted.

Upon opening, the counter is at `3` as expected (it includes my upvote). But if I click it it goes to `4`

If I refresh the page it changes back to `2`, removing my upvote.

tiagoalvesdulce:
Closes #2374 
Closes #2375

This was a tough one to debug...

### Solution description

**Comment sorting (#2374)**
Comment sorting was wrong because `resultvotes` wasn't being updated properly in the comments reducer. I moved it from there to the `Container` level because we don't need to store it at all since we can easily calculate it using up and downvotes.

**Wrong vote color (Scenario 1 on #2375)**
The `useComments` hook was passing `actionData.data` as the vote option to the Likes component but the right property is `vote`. Changing it to `actionData.vote` fixed the issue.

**Wrong vote count (Scenario 2 on #2375)**
The `RECEIVE_LIKED_COMMENTS` action on comments reducer was using a wrong property name to store commentsLikes.

**Unnecessary component updates**
While working on this issue I noticed some unnecessary updates when hovering the vote button. See:
multipleupdatecomment

This was happening because we were using a javascript approach with the [useHover](https://github.com/decred/pi-ui/blob/master/src/hooks/useHover.js) hook from `pi-ui`. In this specific case I could come up with a CSS-only solution to deal with hover and prevent unnecessary component updates.

hovercommentsimproved

**Wrong behavior on error (tested max number of vote changes)**
I would like a review from @victorgcramos on this one because he was the last one to touch this piece of code. I don't get why we need `RESET_SYNC_LIKE_COMMENT` and why we are dispatching it on error. This was leading to some weird errors when hitting the max number of votes error.

**Improve code comments**
The function that was calculating `upvotes` and `downvotes` based on the user previous vote had one edge case wrong and was very hard to understand/debug. I refactored it and add some comments.

**Comment vote UX (https://github.com/decred/politeiagui/issues/2375#issuecomment-833826426)**
Removed the not-allowed symbol during api loading and added a 500ms debounce in the voting functions so the user can't spam the button.

### DIFF
--- a/src/actions/api.js
+++ b/src/actions/api.js
@@ -976,7 +976,8 @@ export const onCommentVote = (currentUserID, token, commentid, vote, state) =>
         dispatch(act.RECEIVE_SYNC_LIKE_COMMENT({ token, commentid, vote }));
       })
       .catch((error) => {
-        dispatch(act.RESET_SYNC_LIKE_COMMENT({ token }));
+        // xxx: why reset on error?
+        // dispatch(act.RESET_SYNC_LIKE_COMMENT({ token }));
         dispatch(act.RECEIVE_LIKE_COMMENT(null, error));
       });
   });
--- a/src/components/Likes/Likes.jsx
+++ b/src/components/Likes/Likes.jsx
@@ -6,7 +6,6 @@ import {
   useTheme,
   Text,
   getThemeProperty,
-  useHover,
   Spinner,
   classNames
 } from "pi-ui";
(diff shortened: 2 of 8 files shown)


victorgcramos:
SYNC_LIKE actions can be removed. I think they are pointless.

lukebp (approved):
ACK

Issue #2371, opened 2021-05-05
Initial Load Flickering


lukebp:
There's all kinds of weird flickering going on during the intial load.  See the attached video. Can we smooth this out?

https://user-images.githubusercontent.com/9039877/117089949-95de5e80-ad1c-11eb-95f0-208e7b6a9c04.mp4

xaur:
Also getting some flickering, slightly different.

amassarwi:
will investigate...

lukebp:
Just noticed this, but what is up with the super skinny `About Politeia`?  That looks like a bug.

tiagoalvesdulce:
Closed by #2395. I couldn't repro the skinny `About Politeia` section, please open a new issue if it's still happening with more details about screen size, browser, etc. See:


https://user-images.githubusercontent.com/13955303/118833197-a69fe000-b897-11eb-9171-b703b88574d6.mp4

PR #2395, opened 2021-05-12
fix: Initial Load Flickering.

amassarwi:
This improves the initial load flickering, before this the paywall message was displayed for some time and then
when the paywall info received from backend the message was gone.

This commit changes this behavior to hide the message by default and only after receiving from the backend that user
hasn't paid the message would show up.

Closes #2395.

### DIFF
--- a/src/actions/api.js
+++ b/src/actions/api.js
@@ -42,26 +42,24 @@ export const onSetHasTotp = (payload) => (dispatch) =>
 
 export const requestApiInfo =
   (fetchUser = true) =>
-  (dispatch) => {
+  async (dispatch) => {
     dispatch(act.REQUEST_INIT_SESSION());
-    return api
-      .apiInfo()
-      .then((response) => {
-        dispatch(act.RECEIVE_INIT_SESSION(response));
-        // if there is an active session, try to fetch the user information
-        // otherwise we make sure there are no user data saved into localstorage
-        if (!response.activeusersession) {
-          clearStateLocalStorage();
-        } else if (fetchUser) {
-          dispatch(onRequestMe());
-        }
+    try {
+      const response = await api.apiInfo();
+      dispatch(act.RECEIVE_INIT_SESSION(response));
+      // if there is an active session, try to fetch the user information
+      // otherwise we make sure there are no user data saved into localstorage
+      if (!response.activeusersession) {
+        clearStateLocalStorage();
+      } else if (fetchUser || response.activeusersession) {
+        await dispatch(onRequestMe());
+      }
 
-        return response;
-      })
-      .catch((error) => {
-        dispatch(act.RECEIVE_INIT_SESSION(null, error));
-        throw error;
-      });
+      return response;
+    } catch (error) {
+      dispatch(act.RECEIVE_INIT_SESSION(null, error));
+      throw error;
+    }
   };
 
 export const onRequestMe = () => (dispatch, getState) => {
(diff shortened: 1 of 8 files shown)


src/actions/api.js
```diff
@@ -95,14 +93,13 @@ export const onPollUserPayment = () => (dispatch, getState) => {
...
-      if (verified) {
-        dispatch(
-          act.UPDATE_USER_PAYWALL_STATUS({
-            status: PAYWALL_STATUS_PAID,
-            userid
-          })
-        );
-      } else {
+      dispatch(
+        act.UPDATE_USER_PAYWALL_STATUS({
```
amassarwi, line 97: we rely on this action to determine is the user paid or not

src/components/PaywallMessage.jsx
```diff
@@ -17,7 +17,8 @@ const PaywallMessage = ({ wrapper, ...props }) => {
       title: "Complete your registration",
       onClose: handleCloseModal
     });
-  const showMessage = !!currentUserEmail && !isPaid;
+  const showMessage = !!currentUserEmail && isPaid === false && isPaid !== null;
```
amassarwi, line 20: If `isPaid` is null we don't want to show the message as this means the isPaid info not available yet.

src/containers/Loader/Loader.js
```diff
@@ -29,16 +28,13 @@ const Loader = ({ children }) => {
       try {
         const apiInfo = await onRequestApiInfo(false);
         setApiInfo(apiInfo);
-        if (apiInfo.activeusersession) {
```
amassarwi, line 32: we have the same check already in `onRequestApiInfo`.

src/reducers/models/paywall.js
```diff
@@ -11,18 +11,11 @@ const paywall = (state = DEFAULT_STATE, action) =>
     ? state
     : (
         {
-          [act.RECEIVE_ME || act.RECEIVE_LOGIN]: () => {
-            const { userid, paywalladdress } = action.payload;
-
-            return update(["byUserID", userid], (paywall) => ({
-              ...paywall,
-              isPaid: paywalladdress === ""
```
amassarwi, line 19: now we update `isPaid` only when on `UPDATE_USER_PAYWALL_STATUS` action, as the backend returns explicit 
`haspaid` field.

amassarwi: this was an issue for paid users; `isPaid: paywalladdress === ""` => resulted in `isPaid = false` this why the paywall 
message was visible for few moments until `UPDATE_USER_PAYWALL_STATUS` was dispatched with status paid which switched `isPaid` to true and only then the paywall message disappeared.

tiagoalvesdulce (commented):
Please fix your PR description, it should close the issue https://github.com/decred/politeiagui/issues/2371.

I'm still seeing some ugly flickering in the UI contents. See:

https://user-images.githubusercontent.com/13955303/118498550-87b91680-b6fc-11eb-8d03-83f254a94812.mp4

I think the initial flicker reported by luke on https://github.com/decred/politeiagui/issues/2371 is fixed so you can choose to create new issues to deal with the issues I mentioned above.

amassarwi:
@tiagoalvesdulce Thanks for the screen record, I would like to fix that in this PR, the paywall shouldn't flicker like that, 
i didn't see that while testing tho, will try now using slower network using the dev tools.

tiagoalvesdulce:
OK, I'll create an issue.

tiagoalvesdulce (approved):

PR #2405, opened 2021-05-21
fix: delete search results on logout

### LINKED ISSUE #2404: Search results persisting after logout
The search results are not being removed from the redux store on logout. To repro do the following without refreshing the page:

Login > search for users > logout > login again > click on search for users in the menu > you should see the table with the old results.

tiagoalvesdulce:
Closes #2404


amassarwi (approved):
LGTM.

lukebp (approved):
ACK

PR #2407, opened 2021-05-24
Change link color to brighter blue when the website is in dark mode

### LINKED ISSUE #2300: External links in Dark Mode difficult to read
Blue on top of blue is difficult to read. I suggest going with a different color: light green or something that will stand out more. See below: 
image

This should just be a simple css change. May be a good first issue.

vibros68:
Resolve #2300

Light mode preview
proposal_link_light_mode

Dark mode preview 
proposal_link_dark_mode

There is another text color is user name. It musts be changed in another repository (pi-ui)


src/components/Markdown/styles.css
```diff
@@ -85,6 +85,10 @@
   text-decoration: none;
 }
 
+.dark .markdown-body a {
+  color: #02ffa3;
```
tiagoalvesdulce, line 89: We prefer to use CSS-variables for the colors, see: https://github.com/decred/pi-ui/blob/master/src/theme/darkTheme.js. Pi-ui css-variables are already imported into pgui so you can just use `color: var(--color-primary)`, for example.

tiagoalvesdulce (changes requested):
As @MariaPleshkova said [here](https://github.com/decred/politeiagui/issues/2300#issuecomment-779867754) I think we should stick to blue for the links. Any new thoughts, Maria?

src/components/Markdown/styles.css
vibros68: Hi @tiagoalvesdulce . I changed as your request. Please take a look

vibros68: There is an var. That is `--color-primary-dark` .But  I did not pick it because it is grey. It  is not good to see too. So I pick `--color-green`

MariaPleshkova:
> As @MariaPleshkova said [here](https://github.com/decred/politeiagui/issues/2300#issuecomment-779867754) I think we should stick to blue for the links. Any new thoughts, Maria?

Let's keep them blue `#2970FF` for now.

Link color is still `#0366d6` on proposals.decred.org :
image

`#2970FF`(one of our accent colors from the palette) is a bit brighter:
image

skynet0590:
> > As @MariaPleshkova said [here](https://github.com/decred/politeiagui/issues/2300#issuecomment-779867754) I think we should stick to blue for the links. Any new thoughts, Maria?
> 
> Let's keep them blue `#2970FF` for now.
> 
> Link color is still `#0366d6` on proposals.decred.org :
> image
> 
> `#2970FF`(one of our accent colors from the palette) is a bit brighter:
> image

Thank you

tiagoalvesdulce (approved):
LGTM. Thanks for your first contribution @vibros68 !

vibros68:
Thank you

tiagoalvesdulce:
@vibros68 can you rename this PR since we decided to go with a brighter blue instead of light green?

tiagoalvesdulce:
Hey @MariaPleshkova I did some research in accessibility/color contrast and google chrome dev tools has a pretty good tool to debug this.

The contrast between our links with `#2970FF` and the background is 2.8, which is considered low. Refer to [this article](https://web.dev/color-and-contrast-accessibility/?utm_source=devtools) for more details.

Screen Shot 2021-06-07 at 21 28 11

Ideally we would increase this contrast ratio. Can we do something about it?

MariaPleshkova:
@tiagoalvesdulce hmm, yeah, maybe it's better to change `#2970FF` then. But I think that using the `#2ED8A3` green is also not a good idea, because we use it for another kind of elements like a tab stroke, field color when entered data is valid, etc. So, I guess we have to add one more color to the palette. I played around with the tool and ended up with lighter blue `#56A4FF`. 
image
The contrast ratio while using this color is 4.5. AA is good, but AAA is not. I didn't find any appropriate color during the use of which AAA is ok. So, maybe one good index is enough?
image

tiagoalvesdulce:
`#56A4FF` Looks good. @vibros68 can you add this color to https://github.com/decred/pi-ui and make that the default color for all our links on dark theme?

vibros68:
I am sorry because have not followed the discussion I will do it now

vibros68:
@tiagoalvesdulce If I create a new variable for the new color. I will need to add some more css on politeiagui. Or if I just change the value of the old variable (color-primary) to `#56A4FF` I will don't need to change  any css on politeiagui but the color will effect other element like the buttons. You can see the picture
Screenshot from 2021-07-13 17-49-43

tiagoalvesdulce:
You can create a new variable on lightTheme (https://github.com/decred/pi-ui/blob/master/src/theme/lightTheme.js) called `link-color` and assign `color-primary` to it. Then, you go to the darkTheme (https://github.com/decred/pi-ui/blob/master/src/theme/darkTheme.js) create a variable with the same name (`link-color`) and override it with `#56A4FF`. Now you just have to go to the places we style links and change the color. Keep in mind that to work with `pi-ui` you have to know [yarn links](https://classic.yarnpkg.com/en/docs/cli/link/)

vibros68:
Thank you. yarn link is exactly what I need at the moment

vibros68:
The last commit will work after [pi-ui/pull/343](https://github.com/decred/pi-ui/pull/343) being merged

tiagoalvesdulce:
Now that https://github.com/decred/pi-ui/pull/343 is in, can you upgrade pi-ui? I like to run upgrade-interactive and upgrade only pi-ui. See: https://classic.yarnpkg.com/en/docs/cli/upgrade-interactive/

tiagoalvesdulce:
I fixed a small issue on https://github.com/decred/pi-ui/pull/351, can you upgrade pi-ui again?

vibros68:
Okay

vibros68:
@tiagoalvesdulce should this PR be merge?

tiagoalvesdulce (approved):
LGTM

vibros68:
Thank you

Issue #2415, opened 2021-05-31
View raw markdown


lukebp:
If somebody wants to view the raw markdown of a proposal there's currently no easy way to do this.  They must download the proposal bundle and then decode the base64 manually in order to view it.  It would be nice to have the option of viewing the raw markdown in the browser, similar to how github allows you view raw markdown.

Does the markdown rendering library that we use allow for this?
How difficult would this be to implement?

cc: @tiagoalvesdulce

tiagoalvesdulce:
It should be pretty straight forward. We don't need the markdown renderer to implement this because we can just print the raw string on the page. The only thing that will be a pain is to deal with images because we use base64 so the raw string is too big for a good UX. My suggestion is that we parse and print images but keep everything else raw. What do you think?

amassarwi:
We could also do something similar to github with images, show some kind of `<img ....` tag or something.

lukebp:
Not sure I follow.  Are you talking about inline images?  If so, wouldn't we just print the markdown that references the image?

If you're not talking about images then I don't think this issue would apply.  This would just be for `.md` files.

tiagoalvesdulce:
hmm, yeah, we can just print the markdown that references the image (image digest).

xaur:
I was missing this very feature recently. Access to raw Markdown allows to save it locally and diff proposal revisions with the tool of choice. Thanks for submitting and implementing.

PR #2416, opened 2021-06-02
feat: allow user to see raw markdown

tiagoalvesdulce:
Closes https://github.com/decred/politeiagui/issues/2415
Depends on https://github.com/decred/pi-ui/pull/339

https://user-images.githubusercontent.com/13955303/121043004-9e4a0f00-c78a-11eb-8925-43f1b5b49c24.mp4

### DIFF
--- a/src/components/Proposal/Proposal.jsx
+++ b/src/components/Proposal/Proposal.jsx
@@ -6,10 +6,9 @@ import {
   useMediaQuery,
   useTheme,
   Tooltip,
-  Link,
   DEFAULT_DARK_THEME_NAME
 } from "pi-ui";
-import React, { useMemo, useState } from "react";
+import React, { useMemo } from "react";
 import Markdown from "../Markdown";
 import ModalSearchVotes from "../ModalSearchVotes";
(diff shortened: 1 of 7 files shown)


lukebp:
Might it be better to create an icon for it and add it to this list?  Seems like the appropriate place for it and would be cleaner.

image

lukebp (approved):
ACK

When you merge this, can you include an explanation in the commit message for why we decided to render the markdown on a different page vs a toggle that renders it in place?  The thought process behind these design decisions is important to document in case we need to re-visit this feature in the future.  Might not be a bad idea to include this in the codebase too.

amassarwi (approved):
code LGTM

xaur:
Adding to the commit message (62879da), "opening raw Markdown in a new tab" translates to "raw Markdown has its own unique URL", which is a very desirable property for any _resource_ or piece of content. Now it is easy to script (`curl https://proposals.decred.org/record/2895755/raw`), which also means the content can be accessed without javascript (and even without browser!), and software like browsers or download managers will have a working link to resume/restart the download.

PR #2428, opened 2021-06-13
fix: token inventory pagination adjustments 

victorgcramos:
This Diff fixes some pagination issues regarding votes/records token inventory. After submitting > 20 proposals, the inventory was not being fetched for page 2, because the pagination only works when a status is given.

### Solution description

The solution was to use the correct status for each list when fetching the token inventory's next page. The first page will return all tokens from all statuses, but the following requests will fetch the inventory according to its status.

### UI Changes

No UI changes expected.


tiagoalvesdulce (approved):
tACK

PR #2431, opened 2021-06-15
Combine 'In Discussion' and 'Voting' tabs

### LINKED ISSUE #2394: PoC: Combine 'In Discussion' and 'Voting' tabs
**Problem**

We removed indicators on the home page tabs (`In Discussion`, `Voting`, `Approved`, etc) that displayed the proposal counts.  This was done for a good reason, but now there is unnecessary friction required when a user wants to see if there are any proposals being voted on.  The page defaults to the `In Discussion` tab.  The user must now click to the voting tab to see if there are any proposals being voted on.

**Solution**

Lets put together a proof of concept that combines the `In Discussion` tab and the `Voting` tab into a single tab.  The home page will default to this tab.  You can display the proposals in the same order that they appear in the `ticketvote` `InventoryReply`.  `Started` would appear first then `Authorized` then `Unauthorized`.

The name `Active` comes to mind for this tab, but I think we should reserve `Active` to mean the proposal is actively being billed against (see the proposal lifecycle section of the [pi development proposal](https://proposals.decred.org/record/91cfcc8)).  Any ideas for this tab name?

Full matrix discussion
```
davecgh
Is there a technical reason the counts were removed from the tabs, btw?  I miss
them!  It was easy to quickly see the status of proposals from just the tabs.
Now I have to click through to voting to see if it's started/still voting.

lukebp
the old politeia had a route that returned all tokens in the inventory
categorized vote status. The gui just keyed off of those to display the tab
count. The new politeia paginates that response. The gui can't key off of it
anymore.

a new route could be added that returns the proposal counts, but generally
speaking, I'm not a fan of the tab count UI. Its only helpful when there are a
small number of proposals in the system. Something like Approved (323) doesn't
really provide you with any valuable information. Tab counts don't scale well
as a UI feature.

I think the UI will likely evolve and move away from the tabs as the number of
proposals grow and as more proposal metadata is required, ex. adding the domain
to the proposal (dev, marketing, etc). This is more of a shower thought at this
point, but I think a design worth exploring would be for the home page to
display proposals based on recent activity (in discussion, voting, recently
ended) and then you would have domain based subreddits that you would navigate
to if you wanted to look back through the proposal history.

maybe combining the In Discussion and Voting tabs into a single tab would be a
good idea so that you don't have to click through to see all the proposals that
are relevant for the user at that point in time

jy-p
tag the entries by color to denote which ones are discussion versus voting

lukebp
we can explore combining In Discussion and Voting into just Active and have the
list items differentiate between In Discussion and Voting using either color or
the status tag that's currently in the top right corner
```

vibros68:
Closed #2394 

### Solution description

Change status config by tab from a number of status into an array of status

Change status argument from number into array

### UI Changes Screenshot

Screenshot from 2021-06-15 18-28-51


src/hooks/api/useProposalsBatch.js
```diff
@@ -97,6 +101,17 @@ export default function useProposalsBatch({
...
   const onFetchTokenInventory = useAction(act.onFetchTokenInventory);
   const hasRemainingTokens = !isEmpty(remainingTokens);
+  const getTokenByStatus = (tokensMap, isByRecordStatus, statuses) => {
+    let tokens = [];
+    statuses.forEach((status) => {
+      const proposalStatusLabel = getProposalStatusLabel(
+        status,
+        isByRecordStatus
+      );
+      return (tokens = [...tokens, ...tokensMap[proposalStatusLabel]]);
```
victorgcramos, line 111: This union here will break the token pagination size rule.

victorgcramos (changes requested):
Hey @vibros68, thanks for your contribution!

I think we shouldn't go for the multiple status array on the `onFetchTokenInventory` action. I think the best approach here is to use the fetch machine do control the list.

Also, I noticed that the pagination is not working as expected. After fetching the first inventory completely, the inventory is fetched again without changing its status, returning the same previously fetched tokens. My inventory for `unauthorized` has only 17 tokens, and it's being fetched twice.

Screen Shot 2021-06-15 at 10 22 41 AM

The fetch machine should only work for 1 status, or else it will break. Each status list have the same rules and the same behavior. Changing the status list to an array will mess with the page size... For `Under Review` it will be > 20, because you are merging 2 lists (see my inline reference below).

In case you don't understand about how the machine works, here's a quick explanation:
1. [START] Fetch the token inventory if not loaded before. Page 0, so it returns the first page for all statuses. Update the page according to the token batch size, then go to VERIFY state;
2. [VERIFY] Checks if there are tokens from the inventory that are not properly fetched. If so, fetch the first record batch page (5 records). After fetching it, it checks for any RFP link existence (which is also a record token). If RFP links, send to VERIFY again, so it recursively fetch the missing RFP tokens. Once all 5 records + RFPs are fetched, the machine goes to DONE.
3. [DONE] means that the cycle for fetching the regarded records batch is complete.

Outside the machine, on `RecordsView`, you can trigger the `onFetchMoreProposals` to fetch the next records batch, meaning it will only change the machine state to VERIFY, so we repeat the step 2 mentioned above ☝️. If all tokens from the inventory are fetched, instead of sending it to VERIFY, we send it to START, where it will load the next token inventory page.

IMO you should focus on the state transitions, and keep the machine working for only 1 status.

lukebp:
I'm not sure how you're odering the proposals, but something looks off.  All proposals with the vote status `started` should be displayed first, then `authorized`, then `unauthorized`.  The image you posted has them mixed together.

tiagoalvesdulce (changes requested):
Found a couple of issues:

1. I'm not seeing the right sorting. As Luke said:

> The Under Review tab will list the proposals with the vote status started, authorized, unauthorized. They will be listed in that order. started will be listed first, then authorized, then unauthorized.

What I'm actually seeing:

Screen Shot 2021-06-15 at 10 15 59

This is related with the batch not being done properly. See:

This is the return of the votes inventory (it is fetching some unnecessary vote statuses but I'll address this later):

Screen Shot 2021-06-15 at 10 25 42

Focus on the proposal returned as `started`. This proposal should be in the first records batch request but instead it's fetching the `authorized` and `unauthorized` proposals before. See:

Screen Shot 2021-06-15 at 10 24 26

You changed the `status` argument of `onFetchInventory` to an array of status. This has no point if in the first fetch we send `statuses = [0]` because the BE understands `0` as no status provided and in this case it will return a page of tokens for all statuses and the GUI will cache it, making all subsequent requests useless (for the first page) since you already got the chunk of data you need to work with.

We have to make deeper changes in order of this to work and I don't think that just sending an array to `onFetchInventory` will do. My suggestion is: keep the status as an integer. If you are in the `under review` tab, fetch vote inventory with status 3 (vote started). You'll get something like this:

```
{
  "vetted": {
    "started": [
      "8676ba0257545570"
    ]
  },
  "bestblock": 705653
}
```

Now you can start fetching with `onFetchProposalsBatch` (5 tokens per time).

If the amount of tokens returned from the `vote inventory` is 20, go ahed and fetch page 2 with the same status.

If the user keeps scrolling and you finish fetching the `started` status, move to `authorized` and do the same thing.

Fixing this will partially fix the sorting. With that implemented we should remove the FE sorting by status because we will already be fetching in the right order. Notice that the BE sorts the arrays of tokens returned in the votes inventory by timestamp of the last status change, from the most recent to the last recent. 

2. When I start a vote I'm seeing this:

Screen Shot 2021-06-15 at 10 14 55

The Voting tab no longer exists so this message should be rephrased. `The proposal vote has been successfully started!` is OK.

Left a few inline comments as well.

src/actions/api.js
```diff
@@ -535,52 +536,70 @@ export const onFetchProposalDetails = (token, version) => async (dispatch) => {
...
+                (votes && votes.vetted && votes.vetted.rejected) || [],
+              [INELIGIBLE]:
+                (votes && votes.vetted && votes.vetted.ineligible) || []
+            };
+            !isVoteStatus && dispatch(act.RECEIVE_RECORDS_INVENTORY(byRecords));
+            isVoteStatus && dispatch(act.RECEIVE_VOTES_INVENTORY(byVotes));
+            return { records: byRecords, votes: byVotes };
+          });
+        })
+      ).then((tokenInventories) => {
```
tiagoalvesdulce, line 597: I don't get the need to merge this that since they won't return tokens in the same context. See the `isVoteStatus ` boolean, you will always have `records` OR `votes` but never `records` AND `votes`.

src/containers/Proposal/Vetted/helpers.js
```diff
@@ -26,6 +25,8 @@ export const getProposalTokensByTabOption = (tabOption, proposalsTokens) => {
   if (!proposalsTokens) return [];
   const { pre, started, approved, rejected, ineligible } = proposalsTokens;
   switch (tabOption) {
+    case tabValues.UNDER_REVIEW:
```
tiagoalvesdulce, line 28: Be careful with sorting here.

src/hooks/api/useProposalsBatch.js
```diff
@@ -106,21 +121,19 @@ export default function useProposalsBatch({
         if (page && page === previousPage) return send(RESOLVE);
         onFetchTokenInventory(
           recordState,
-          page && currentStatus, // first page fetches all status
+          page ? currentStatus : [0], // first page fetches all status
```
tiagoalvesdulce, line 124: The thing I said about status being 0 in the review.

src/containers/Proposal/Vetted/Vetted.jsx
```diff
@@ -72,6 +71,7 @@ const VettedProposals = ({ TopBanner, PageDetails, Sidebar, Main }) => {
 
   const getEmptyMessage = useCallback((tab) => {
     const mapTabToMessage = {
+      [tabValues.UNDER_REVIEW]: "No proposals under review",
       [tabValues.IN_DISCUSSION]: "No proposals under discussion",
```
tiagoalvesdulce, line 75: This should leave.

tiagoalvesdulce (changes requested):

src/hooks/api/useProposalsBatch.js
vibros68: Thanks. I will fix it soon

vibros68:
@tiagoalvesdulce your way is right. I think I should do that way. Thank you

vibros68:
Hi all. I fix things after your review:
* Remove change on `onFetchProposalsBatch` func and use `useFetchMachine` func.
* Process can be described: 

At the first fetch to `inventory` endpoint. I send `status = 0` to fetch all the tokens of all statuses for the page 1st.  This is needed because at the `onRestartMachine` func (which is called when the user click on switch tab) I can verify: in that status, did we fetch all the needed data? while don't need to send to BE an `inventory` request.

(to @tiagoalvesdulce as we discussed that fetch only 1 status at the first request but after do coding I see this way is simpler way to dealt)

After all tokens of first page are fetched. I go to VERIFY state to fetch proposals batch (4 items at max) till we get the all needed data for that status. Then I go to RESOLVE state.

 At the RESOLVE state. I changed this func from empty func to check: If there is any status which is not handled: change the current status to that status and send back to START state. Here we do the same till there are no data. 

* About sort problem. I changed from sort by timestamp(desc) to sort by voteStatus(desc) and timestamp(desc)
* I tested and see all the requests looks fine without waste requests

src/actions/api.js
vibros68: The change of api was removed

lukebp:
@vibros68 Please don't force push changes.  Force pushing hides the changes that are being made.  The reviewer then has to go back and re-review code that might not have been changed.

vibros68:
@lukebp This change is redo because I changed the way of solution. So it almost is not related with the first time. But I don't think the problem is force push. The problem is I used `git rebase -i` and squash the previous commit. By the way the commits look cleaner. Any way I am sorry about that and will keep in mind this problem and take experience in the next time

lukebp:
Merge instead of rebasing and you will push able to update the PR without needing to force push.

src/containers/Proposal/Vetted/Vetted.jsx
```diff
@@ -37,7 +36,8 @@ const VettedProposals = ({ TopBanner, PageDetails, Sidebar, Main }) => {
   } = useProposalsBatch({
     fetchRfpLinks: true,
     fetchVoteSummaries: true,
-    status: statusByTab[tabLabels[index]]
+    statuses: statusByTab[tabLabels[index]],
+    proposalPageSize: 4
```
tiagoalvesdulce, line 40: Just to endorse @lukebp's comment. We removed this param here: https://github.com/decred/politeiagui/pull/2447/files#diff-52c024866c95acc478004c9be1d44fd6f39aa09128f1179c35cda0a2f7828830L40, defaulting it to 5. You force-pushed the conflict resolution and now we can't see how you solved the merge conflicts.

src/hooks/api/useLegacyVettedProposals.js
```diff
@@ -27,17 +27,26 @@ const mapStatusToString = {
   [PROPOSAL_VOTING_INELIGIBLE]: "abandoned"
 };
 
-export default function useLegacyVettedProposals(shouldReturn = false, status) {
+export default function useLegacyVettedProposals(
```
tiagoalvesdulce, line 30: I don't get the need to change this hook. We only have approved, rejected and abandoned legacy proposals (from old pi archive, you can see they are hardcoded json structures https://github.com/decred/politeiagui/blob/master/src/legacyproposals.json and will be removed soon). I think this part of code shouldn't be touched on this PR.

tiagoalvesdulce (changes requested):

src/hooks/api/useLegacyVettedProposals.js
tiagoalvesdulce: This could be the reason why the sorting of approved proposals is getting messed up. See new proposals after legacy proposals:

Screen Shot 2021-06-21 at 09 37 34

tiagoalvesdulce: ps: the same happens once you load all the proposals on the rejected tab

vibros68: Because the entry changed from a number (status) to array. If I do not change the code inside that func. I will need change the input to `statuses[0]`. What is your opinion?

tiagoalvesdulce: Your changes shouldn't affect `useLegacy`. Notice it's creating a collateral as I mentioned above. Please format the parameter before passing it to the function, don't change an already proven to work function that's been used in other parts of the code.

src/containers/Proposal/Vetted/Vetted.jsx
vibros68: Hi. May be there are mistake when I solve the conflict. I was remove that variable in the next commit

tiagoalvesdulce:
The issue I mentioned about proposals showing after legacy is still happening.

image

vibros68:
Hi @tiagoalvesdulce . I fixed the issue of `the sorting of approved proposals is getting messed up`. I make the sort be flexible. It will be voteStatus and timestamp in `Under Review` tab and be timestamp with other tabs as original

src/containers/Proposal/Vetted/helpers.js
```diff
@@ -56,9 +53,19 @@ export const mapProposalsTokensByTab = (tabLabels, proposalsTokens) => {
...
+  [tabValues.REJECTED]: [PROPOSAL_VOTING_REJECTED],
+  [tabValues.INELIGIBLE]: [PROPOSAL_VOTING_INELIGIBLE]
+};
+
+export const sortByTab = {
+  [tabValues.UNDER_REVIEW]: {
+    fields: ["voteStatus", "timestamp"],
+    order: ["desc", "desc"]
+  }
+  // For other tabs = undefined (get the default)
```
victorgcramos, line 70: what is the default?

src/hooks/api/useProposalsBatch.js
```diff
@@ -165,7 +173,25 @@ export default function useProposalsBatch({
...
+            proposals,
+            uniq(
+              allByStatus[
+                getProposalStatusLabel(
+                  currentStatuses[newIndex],
+                  isByRecordStatus
+                )
+              ] || []
+            )
+          );
```
victorgcramos, line 189: I think you could improve the readability in this case... Maybe use a new variable/func?

src/hooks/api/useProposalsBatch.js
```diff
@@ -176,29 +202,32 @@ export default function useProposalsBatch({
...
+        proposals,
+        uniq(allByStatus[statusLabel] || [])
+      );
+      if (unfetchedTokens.length > 0) {
+        index = i;
+        return true;
+      }
+      return false;
+    });
+    if (!foundStatus) {
```
victorgcramos, line 221: If the status is not found, then we have an error. In this case, the machine should be sent to the `FAILURE` state (`REJECT`action), or handle the error properly.

victorgcramos (changes requested):
Hey @vibros68, noticed that the multiple statuses by tab (as you've done with Under Review tab) should work for any multiple status I pass on the `statusByTab` object. Noticed that if I wanted to add another status into the Approved tab, the list doesn't render it. IMO, since you opted to go for a multi-status array approach, the machine should behave the same for any status array you pass to it.

The expected behavior on this case should be:
```javascript
// if I wanted to render another status on APPROVED list, for example:

export const statusByTab = {
  [tabValues.UNDER_REVIEW]: [
    PROPOSAL_VOTING_ACTIVE,
    PROPOSAL_VOTING_NOT_AUTHORIZED
  ],
  [tabValues.APPROVED]: [
    PROPOSAL_VOTING_NOT_AUTHORIZED,
    PROPOSAL_VOTING_APPROVED
  ],
  [tabValues.REJECTED]: [PROPOSAL_VOTING_REJECTED],
  [tabValues.INELIGIBLE]: [PROPOSAL_VOTING_INELIGIBLE]
};
```

And the list displays the unauthorized proposals first, then the approved ones. When I do it, only the approved ones are rendered.
Screen Shot 2021-06-29 at 7 51 49 PM

This might sound minor, but since the fetch machine presented the same behavior for every singular status, it should also present the same behavior for any multi-status array.

Also left some inline comments regarding the code readability

vibros68:
@victorgcramos 
> 
> The expected behavior on this case should be:
> 
> ```js
> // if I wanted to render another status on APPROVED list, for example:
> 
> export const statusByTab = {
>   [tabValues.UNDER_REVIEW]: [
>     PROPOSAL_VOTING_ACTIVE,
>     PROPOSAL_VOTING_NOT_AUTHORIZED
>   ],
>   [tabValues.APPROVED]: [
>     PROPOSAL_VOTING_NOT_AUTHORIZED,
>     PROPOSAL_VOTING_APPROVED
>   ],
>   [tabValues.REJECTED]: [PROPOSAL_VOTING_REJECTED],
>   [tabValues.INELIGIBLE]: [PROPOSAL_VOTING_INELIGIBLE]
> };
> ```

You need to add the value on here also. This is the old way master did
```
export const getProposalTokensByTabOption = (tabOption, proposalsTokens) => {
  if (!proposalsTokens) return [];
  const { pre, started, approved, rejected, ineligible } = proposalsTokens;
  switch (tabOption) {
    case tabValues.UNDER_REVIEW:
      return [...started, ...pre];
    case tabValues.APPROVED:
      return [...approved, ...pre];
    case tabValues.REJECTED:
      return rejected;
    case tabValues.INELIGIBLE:
      return ineligible;
    default:
      return [];
  }
};
```

victorgcramos:
@vibros68 I think the development experience in this case is a little confusing. In order to avoid confusion like the one I just did, I'd suggest to change the `getProposalTokensByTabOption` to get proposals tokens from given statuses, instead. It accepts an array of statuses, and returns the tokens according to the statuses on this array. For example, let's call it `getRecordsTokensByStatusList`:

```javascript
const tokens = getRecordsTokensByStatusList(statusByTab[tabValues.APPROVED], inventory);
//  returns an array of tokens.
```

This should also standardize the two methods used by [src/containers/Proposal/Admin/helpers.js](https://github.com/decred/politeiagui/blob/518000960a50423bcecd0a0ea5aac042b16b2223/src/containers/Proposal/Admin/helpers.js#L37) and [src/containers/Proposal/Vetted/helpers.js](https://github.com/decred/politeiagui/blob/master/src/containers/Proposal/Vetted/helpers.js#L25), which are doing the same, aren't they?


You can use the helper below to help you: https://github.com/decred/politeiagui/blob/518000960a50423bcecd0a0ea5aac042b16b2223/src/containers/Proposal/helpers.js#L375-L392

Give me your thoughts on this one. What do you think?

vibros68:
Thank you @victorgcramos . I think it is a good idea. I got the same problem as you encounter when be starting work on this issue. But because I am a new contributor so I just do as the same as the old way. 

>This should also standardize the two methods used by [src/containers/Proposal/Admin/helpers.js](https://github.com/decred/politeiagui/blob/518000960a50423bcecd0a0ea5aac042b16b2223/src/containers/Proposal/Admin/helpers.js#L37) and [src/containers/Proposal/Vetted/helpers.js](https://github.com/decred/politeiagui/blob/master/src/containers/Proposal/Vetted/helpers.js#L25), which are doing the same, aren't they?

About this comment. Do you want to merge 2 function into 1. Right? If it is right I think it is better too

victorgcramos (changes requested):
Hey, PR looking really good. I'd suggest some few minor changes:

1. When you start a voting for some proposal, the list should be updated according to the status changes as well. Noticed that the proposal remains on the same place on the Under Review list, when it should go to the top of the list, as we previously moved the proposal to the top of `voting` tab.

2. When you submit a new vetted proposal, the proposal is put on top of the list, even if there are proposals under vote. The first proposals should be the voting ones, and then we display the unreviewed ones.

These are all minor, but since you are working on a feature that comprises the most important part of the UX of our platform, those previous pattern should be kept.

lukebp:
I agree with the comments victor made.  Just to clarify in case this is not already being done, the proposals listed in the `Under Review` tab should be grouped by vote status.

Ongoing votes are first.
Authorized votes are second.
Unauthorized votes are third.

vibros68:
I  am working on it now

src/hooks/api/useProposalsBatch.js
```diff
@@ -176,29 +201,35 @@ export default function useProposalsBatch({
...
       return send(RESOLVE);
```
victorgcramos, line 224: Not found statuses are still being handled as "resolved". Shouldn't it return an error?

src/containers/Proposal/Vetted/helpers.js
```diff
@@ -7,58 +7,29 @@ import {
...
-    return {
-      ...map,
-      [tab]: getProposalTokensByTabOption(tab, proposalsTokens)
-    };
-  }, {});
+export const statusByTab = {
+  [tabValues.UNDER_REVIEW]: [
+    PROPOSAL_VOTING_ACTIVE,
+    PROPOSAL_VOTING_NOT_AUTHORIZED
+  ],
```
victorgcramos, line 22:
```suggestion
  [tabValues.UNDER_REVIEW]: [
    PROPOSAL_VOTING_ACTIVE,
    PROPOSAL_VOTING_AUTHORIZED,
    PROPOSAL_VOTING_NOT_AUTHORIZED
  ],
```
Once you fix the inventory action on [src/actions/api.js#L590-593](https://github.com/decred/politeiagui/blob/595f99ac95d555d09973407333b2c4c44393c4e8/src/actions/api.js#L590-L593), you can add the `authorized` status here

victorgcramos (changes requested):
Hey @vibros68, noticed some stuff you could also improve in your PR. We currently merge the inventory object by two statuses `authorized` and `unauthorized` ones, in order to show both statuses on the same list. It's called `PRE_VOTE`. See below:

https://github.com/decred/politeiagui/blob/595f99ac95d555d09973407333b2c4c44393c4e8/src/actions/api.js#L590-L593

But once you adapted the fetch machine to handle status arrays, I think you can remove the `PRE_VOTE` status and add the `unauthorized` and `authorized` separately, just like you handled `voting` and `pre` ones. See: https://github.com/decred/politeiagui/blob/595f99ac95d555d09973407333b2c4c44393c4e8/src/containers/Proposal/helpers.js#L385

src/hooks/api/useProposalsBatch.js
vibros68: Hi. I have  a comment on code about this problem: 
```
// onRestartMachine called when the user clicking on a new tab
// The function will find what is the status has been loading in the previous session.
// If there are not status found. That mean all the data are loaded in this tab and go to RESOLVE state
```

victorgcramos: hmmm got it. Makes sense! However, the `foundStatus` naming is confusing... It means that there are unfetched tokens for some status, but the `foundStatus` naming makes us believe that some status was found, not the tokens itself. Status will exist, the resolution is correct, but in this case I'd just change the variable naming.

lukebp:
@vibros68 I've already asked you once to not force push.  Why are you continuing to force push?

vibros68:
> @vibros68 I've already asked you once to not force push. Why are you continuing to force push?

I make a mistake with the last commit. I forgot to remove a line of console.log so I fixed the last commit by squash commit

vibros68:
I am sorry if it is bad behavior here. In the past I used squash commit to make the commits be cleaner. The last team I work recommended me to produce clean commits so that is my habit when I make some ugly commits.

lukebp:
Please do not force push.  You can squash a group of commits before pushing them, but once they've been pushed they should not be overwritten.

vibros68:
Yes. I will remember this

lukebp (changes requested):
I'm seeing some loading issues with flickering when testing this PR that is not present on master.

Master

https://user-images.githubusercontent.com/9039877/125647780-ed19b4cb-d83e-4916-b861-b348f32b2f84.mp4

PR 2431

https://user-images.githubusercontent.com/9039877/125647828-ce391108-1217-4627-a738-7fb64792b7dd.mp4

lukebp (changes requested):
I found quite a few more issues.  This PR needs some work.  You should be testing this stuff manually and verifying that it works before asking for future reviews.  Do not waste the reviewers time by submitting PRs that you have not thoroughly tested yourself.

**The loading graphics are not being shown while the "Under Review" list is being populated**


https://user-images.githubusercontent.com/9039877/125651661-95dd102d-bd6c-417c-86a1-eb9d410ae457.mp4

**Pagination is completely broken in the Under Review tab.**  

I started 30 proposal votes using the command `pictl votetestsetup [adminEmail] [adminPassword] --votes=30 --duration=16`.  There are only 5 proposal votes showing with no way to load more and its displaying the authorized and unauthorized proposals before it has finished displaying all the proposals being voted on.

Pagination on this tab should work just like the other tabs, except it has a hierarchy of which proposals should be shown first.  It should work like this:

1. Shows the first page of proposal with the `started` vote status. 
1a. If there are not a full page of proposal votes then it continues to fetch the first page of proposals in the next vote status in the heirarchy (`authorized`).
1b. If there are a full page of proposal votes then it pagniates the proposals like normal until it reaches the end of the `started` proposals list. You will need to keep fetching the vote inventory for the `started` vote status until the page of returned tokens is less than the inventory page size. Once all of the `started` proposals have been displayed in a paginated fashion, continue to the `authorized` proposals.
2. Fetch the `authorized` procedure by repeating the procedure in step 1.
3. Fetch the `unauthorized` procedure by repeating the procedure in step 1.


https://user-images.githubusercontent.com/9039877/125653171-5966e78d-743b-4613-a774-946a59897647.mp4

victorgcramos:
> The loading graphics are not being shown while the "Under Review" list is being populated

@lukebp we have an issue for that: #2471

lukebp:
Ok, thanks for the info @victorgcramos.  Fixing that will be out of scope for this PR so that comment can be skipped.  The intial load flickering and the pagination issues are still applicable to this PR.

vibros68:
https://user-images.githubusercontent.com/84569563/125728746-9e724a43-0872-41e8-aecf-22c39aecd1e5.mp4

In my side pagination still be working well. About the flicker. I am working on it

vibros68:
> I'm seeing some loading issues with flickering when testing this PR that is not present on master.

Because there are more than 1 status in this tab. So at the first time the app will load `started` proposals, if there are no `started` proposal, a new cycle of loading will be started. So that is the cause you see the flickering

lukebp:
@vibros68 verify the pagination when there are more than 5 proposals being voted on and more than 5 proposals that are authorized.  I think that's when you'll see the issue.

lukebp:
This is getting better, but its still not right.  If you compare the pagination to master you'll see the issue.

master: list view is populated with the first page.  Additional pages are not populated until you scroll down.  The loading icon is being shown while the page is being retrieved.

https://user-images.githubusercontent.com/9039877/126078763-4056078d-f829-4c5d-8cfe-ce26498b1517.mp4


this branch: all pages are retrieved without the user scrolling down


https://user-images.githubusercontent.com/9039877/126078820-c35b30fc-0ac9-48b3-a3ed-39c4b6335eed.mp4

lukebp (changes requested):
See pagination and scrolling issues from previous comment.

lukebp (changes requested):
This is looking much better.  The pagination UX is correct now.  I did find one edge case bug though.  When there are no proposals being voted on, but there are proposals that have not started voting yet, the proposals that are not being voted on do not load initially.  See the video below.  You have to switch tabs in order to get them to load.

https://user-images.githubusercontent.com/9039877/126167619-3b7a1d6f-2c55-4646-825f-4898b0f9a188.mp4

lukebp (approved):
This looks correct from a UX standpoint.  Nice work.

@tiagoalvesdulce will need to approve the code before this can be merged.

src/containers/Proposal/helpers.js
```diff
@@ -409,3 +411,38 @@ export const getTokensForProposalsPagination = (
   tokens,
   pageSize = PROPOSAL_PAGE_SIZE
 ) => [take(pageSize)(tokens), takeRight(tokens.length - pageSize)(tokens)];
+
+/**
+ * Returns the proposal tokens by status from an inventory of tokens
+ * @param {*} statuses
+ * @param {*} isByRecordStatus
+ * @param {*} inventory
```
tiagoalvesdulce, line 419: When using jsdoc, please inform the type. https://jsdoc.app/tags-type.html

src/containers/Proposal/helpers.js
```diff
@@ -409,3 +411,38 @@ export const getTokensForProposalsPagination = (
...
+    const label = getProposalStatusLabel(status, isByRecordStatus);
+    return [...tokens, ...(inventory[label] || [])];
+  }, []);
+
+/**
+ * Returns a map of proposals tokens with the key is tab name
+ * the tokens are taken from the inventory, filtered by status
+ * @param {*} statusByTab
+ * @param {*} inventory
+ * @param {*} isByRecordStatus
```
tiagoalvesdulce, line 436: Same as above

tiagoalvesdulce (commented):
Code is looking good. I left a few comments. The only thing that is bothering me about UX is that when the proposal goes down the proposal list (e.g. goes from authorized to unauthorized) it can move when more proposals are loaded. See:

https://user-images.githubusercontent.com/13955303/126534836-8ead249e-acb5-4a3f-bd6b-54d079cbc4ef.mp4

In my opinion we shouldn't re-sort the list on status changes. Thoughts @lukebp ?

lukebp:
@tiagoalvesdulce are you refering to the view of the person who made the state change?  Not re-sorting it for them makes sense.  The list item moving around like that can be confusing. 

If the page is being loaded from scratch then I think the sorting is fine.

tiagoalvesdulce:
Yes, I am talking about the view of the person who made the state change.

vibros68:
So. Do we decide to not re-sorting it after the user made the state change?

src/actions/api.js
```diff
@@ -1322,16 +1323,19 @@ export const onFetchVotesDetails = (token) =>
...
-      .then((response) =>
+      .then((response) => {
         dispatch(
           act.RECEIVE_AUTHORIZE_VOTE({ ...response, token, success: true })
-        )
-      )
+        );
+        if (typeof doneCb === "function") {
+          doneCb();
+        }
```
victorgcramos, line 1337: Is this used elsewhere? If not, please remove it

victorgcramos (changes requested):
Left one minor inline comment

victorgcramos (changes requested):
Code looks good! Please make sure all tests are running smoothly. #2473 added tests for the proposals list, so please make sure they are okay.

vibros68:
Screenshot from 2021-07-30 15-21-08
You can see. The test is expecting 5 items. While it got 8 items. But we can see on the inventory response. There are 3 items in `started` and 20 in `unauthorized`. With the logic of Under Review tab. It should fetched 3 items of started fisrt. Then because still there are space in the bottom. So it will fetch the next page (that is the first page of unauthorized). So the list should have 8 items and that is correct value

lukebp:
@tiagoalvesdulce can you give this a final review when you get a chance.

vibros68:
@victorgcramos I fixed the e2e test. So it is running without bugs now

teste2e/cypress/e2e/records/recordsList.js
```diff
@@ -1,6 +1,7 @@
+import {shortRecordToken} from "../../../../src/helpers";
```
amass01, line 1: hrmm this looks like not formatted/prettified code we should include the e2etest files to prettify and lint commands.

vibros68: I tried import like `"src/helpers"` but it does not work

victorgcramos: there's the same function on e2e utils! You can import from there!

src/hooks/api/useProposalsBatch.js
```diff
@@ -165,7 +197,29 @@ export default function useProposalsBatch({
           .catch((e) => send(REJECT, e));
         return send(FETCH);
       },
-      done: () => {}
+      done: () => {
+        // If there are not remaining tokens. Check there are unscanned status or not.
```
tiagoalvesdulce, line 201:
```suggestion
        // If there are not remaining tokens, check if have unscanned status.
```

src/hooks/api/useProposalsBatch.js
```diff
@@ -165,7 +197,29 @@ export default function useProposalsBatch({
...
+            uniq(
+              allByStatus[
+                getProposalStatusLabel(newStatus, isByRecordStatus)
+              ] || []
+            )
+          );
+          setRemainingTokens(unfetchedTokens);
+          setStatusIndex(newIndex);
+          if (!values(proposals).length) {
+            // In the case there are no proposals fetched yet.
```
tiagoalvesdulce, line 217:
```suggestion
            // If no proposals are fetched yet.
```

src/hooks/api/useProposalsBatch.js
```diff
@@ -165,7 +197,29 @@ export default function useProposalsBatch({
...
+            // Continue the fetching cycle to fetch the first page at user view
```
tiagoalvesdulce, line 218:
```suggestion
            // Continue the fetching cycle to fetch the first page.
```

src/hooks/api/useProposalsBatch.js
```diff
@@ -176,29 +230,37 @@ export default function useProposalsBatch({
...
-      isByRecordStatus
-    );
-    const unfetchedTokens = getUnfetchedTokens(
-      proposals,
-      uniq(allByStatus[newStatusLabel] || [])
-    );
-
-    // machine stop condition: inventory loaded, but no tokens to fetch
-    if (isEmpty(remainingTokens) && isEmpty(unfetchedTokens) && !page) {
+  // onRestartMachine called when the user clicking on a new tab
```
tiagoalvesdulce, line 233:
```suggestion
  // onRestartMachine is called when the user clicks on a new tab
```

src/hooks/api/useProposalsBatch.js
```diff
@@ -176,29 +230,37 @@ export default function useProposalsBatch({
...
+  // The function will find what is the status has been loading in the previous session.
```
tiagoalvesdulce, line 234:
```suggestion
  // The function will find out whether the status has been loading in the previous session.
```

src/hooks/api/useProposalsBatch.js
```diff
@@ -176,29 +230,37 @@ export default function useProposalsBatch({
...
+  // If there are not status found. That mean all the data are loaded in this tab and go to RESOLVE state
```
tiagoalvesdulce, line 235:
```suggestion
  // If no status is found, all the data is loaded in this tab. Go to the RESOLVE state.
```

teste2e/cypress/e2e/records/recordsList.js
```diff
@@ -37,37 +38,37 @@ describe("Records list", () => {
       });
       cy.wait("@records.records");
       // each proposal should be rendered accordingly to inventory response
-      cy.assertListLengthByTestId("record-title", RECORDS_PAGE_SIZE) // first records batch
+      cy.assertListLengthByTestId("record-title", RECORDS_PAGE_SIZE + 3) // first records batch
```
tiagoalvesdulce, line 41: Is it ok to hardcode this? Looks like it won't work in some cases. @victorgcramos

tiagoalvesdulce (commented):
Code is looking good. I tested and it works as expected. Left a few suggestions and a comment.

teste2e/cypress/e2e/records/recordsList.js
vibros68: I explained this problem in [here](https://github.com/decred/politeiagui/pull/2431#issuecomment-889745223)

victorgcramos: I guess you should get that info from the inventory you just passed to the middleware. Maybe create a const and use it both on the middleware and the assertion.

Since the middleware controls how the inventory should look like, we'll only see issues if our middleware inventory changes.

vibros68: As I explained. If there are 3 items in `started` status. Will have a space in the bottom. So the fetch machine will fetch more a page to fill up the space. But if there are more than 5 items, the first page will fill up the space of the body. And we need to scroll down to load the next page. So I think hard code in this case is fine. 
@victorgcramos @tiagoalvesdulce

teste2e/cypress/e2e/records/recordsList.js
amass01: oh right, what @victorgcramos said!

amass01: but that's not what my comment is about, this code isn't fomatted it should be `import { shortRecordToken }...` note the missing whitespaces.

amass01: actually that's not in the scope of this PR, please ignore my comment i will open a new issue and work on it.

teste2e/cypress/e2e/records/recordsList.js
tiagoalvesdulce: I get your issue @vibros68, just wondering if the solution work for all cases.

vibros68: For the other case. If the screen size of the browser is smaller than the current one. The fetch machine will stop at the first page with 3 items. But the screen size seem be fixed at that size. May be @victorgcramos know more than me at this point

victorgcramos: I think this is fine. I agree with @vibros68

teste2e/cypress/e2e/records/recordsList.js
vibros68: Okay. I changed the import path to the function you mentioned. But for the prettify work. It should not be in this pr. I think that too. Thank you @amass01

teste2e/cypress/e2e/records/recordsList.js
vibros68: Okay. Now everything seem be okay. I think the pr should be reviewed last one and get merged

tiagoalvesdulce (approved):
tACK

PR #2434, opened 2021-06-17
feat: create a button style flat mode switch and align it horizontally with the sorting options dropdown

tiagoalvesdulce:


lukebp (approved):
Before

image

After
image
image

Looks great!

PR #2441, opened 2021-06-18
fix: user proposals pagination

tiagoalvesdulce:
This PR fixes a rare but critical bug that the cache was masking. The User page wasn't properly paginated e.g it was attempting to do a request with `all user proposals - cached proposals`, if that number was bigger than 20, the `records` request would return `400` and the user proposals list would be empty.

### Solution description

I implemented pagination for this route.


lukebp (approved):
ACK

The pull request that introduced this bug is unkown.  The bug appears to have been present for a long time.

Future prevention
- Code reviewers need to fully consider all edge cases and assume the system needs to scale to a large number of records
- e2e tests that confirm the correct behavior for user record pagination (#2443)

src/actions/types.js
```diff
@@ -52,8 +52,10 @@ export const REQUEST_LOGOUT = "API_REQUEST_LOGOUT";
 export const RECEIVE_LOGOUT = "API_RECEIVE_LOGOUT";
 export const RECEIVE_CMS_LOGOUT = "API_RECEIVE_CMS_LOGOUT";
 
-export const REQUEST_USER_PROPOSALS = "REQUEST_USER_PROPOSALS";
-export const RECEIVE_USER_PROPOSALS = "RECEIVE_USER_PROPOSALS";
+// export const REQUEST_USER_PROPOSALS = "REQUEST_USER_PROPOSALS";
+// export const RECEIVE_USER_PROPOSALS = "RECEIVE_USER_PROPOSALS";
```
victorgcramos, line 56: I think this should leave 👍

src/containers/Proposal/User/Submitted.jsx
```diff
@@ -20,8 +20,8 @@ const Proposals = (props) => {
   const [hasMoreToLoad, setHasMore] = useState(false);
 
   const { userID } = props;
-  const { proposals, loading, numOfUserProposals, onFetchUserProposals } =
-    useUserProposals(userID);
+  const { proposals, loading, numOfUserProposals, onFetchMoreProposals } =
+    useUserProposals({ userID });
```
victorgcramos, line 24: Code formatting stuff. I think you should use this approach here
```suggestion
  const {
    proposals,
    loading,
    numOfUserProposals,
    onFetchMoreProposals
  } = useUserProposals({ userID });
```

src/containers/Proposal/User/hooks.js
```diff
@@ -67,102 +61,151 @@ export function useDraftProposals() {
...
+        if (hasRemainingTokens) return send(VERIFY);
+        onFetchUserProposals(userID)
+          .catch((e) => send(REJECT, e))
+          .then((res) => {
+            const resTokens = [...res.unvetted, ...res.vetted];
+            if (!resTokens) return send(RESOLVE);
+            setRemainingTokens(resTokens);
+            return send(VERIFY);
+          });
+        return send(FETCH);
```
victorgcramos, line 132: Perfect 🚀

victorgcramos (approved):
LGTM! It's working as expected! Left some inline comments with some minor suggestions

src/containers/Proposal/User/Submitted.jsx
tiagoalvesdulce: Let's let prettify take care of this.

Issue #2445, opened 2021-06-18
Paywall error message when paywall is off


lukebp:
I'm still seeing the paywall error message when the paywall is turned off.

![image](https://user-images.githubusercontent.com/9039877/122600427-5bfbba00-d035-11eb-9852-1b91f63be87a.png)

PR #2446, opened 2021-06-18
fix: do not show error msg when paywall is off

tiagoalvesdulce:
Closes https://github.com/decred/politeiagui/issues/2445

This PR also removes the `user/<userid>?tab=credits` when the paywall is off.


lukebp (approved):
ACK

victorgcramos (approved):
tACK!

PR #2449, opened 2021-06-18
fix: do not fetch summaries and count for rfp submissions details + improve initial fetch condition

tiagoalvesdulce:
This PR does 3 simple things:

1. Remove the **need** of initial fetching if proposal is a RFP. Notice it will still fetch if it's missing details but it was forcing a fetch every time before.
2. Improve proposal details loading state using the state machine.
3. Do not fetch `summaries` and `count` for the RFP proposal of a RFP submission in the proposal details.

## UI Changes

https://user-images.githubusercontent.com/13955303/122621377-bb74bc80-d06b-11eb-8851-4e84f9f56f26.mp4

### DIFF
--- a/src/containers/Proposal/Detail/hooks.js
+++ b/src/containers/Proposal/Detail/hooks.js
@@ -92,7 +92,7 @@ export function useProposal(token, threadParentID) {
   const isMissingVoteSummary = !(
     voteSummaries[tokenShort] && voteSummaries[tokenShort].details
   );
-  const needsInitialFetch = (token && isMissingDetails);
+  const needsInitialFetch = token && isMissingDetails;
 
   const [remainingTokens, setRemainingTokens] = useState(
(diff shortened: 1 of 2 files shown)


src/containers/Proposal/Detail/hooks.js
```diff
@@ -158,11 +166,13 @@ export function useProposal(token, threadParentID) {
       },
       done: () => {
         // verify proposal on proposal changes
+        // TODO: improve this in the future so we don't need to verify once it should be done
```
tiagoalvesdulce, line 169: That initial flicking occurs because of this. It's returning to the verify state after being done. I tried some approaches bu they end up generating collaterals. Will investigate later.

lukebp (approved):
ACK

victorgcramos (approved):
LGTM! Good job!

Issue #2450, opened 2021-06-18
Pagination issues on Abandoned tab


tiagoalvesdulce:
We are seeing multiple unnecessary requests being fired on this tab. We have to make sure that:

1. List item cards should be the same size so we don't have different behaviors between tabs
2. A page of proposals should be sending at most 4 requests. The first 3 requests are the records batch, comments count, and vote summaries. The last request will only be when the parent RFP name is needed and should be a records batch for the RFP names.

lukebp:
I can't find the exact PR, but 1 has been fixed.  I believe 2 has also been fixed by #2447.  Closing this issue.  Please re-open it if I am mistaken.

PR #2454, opened 2021-06-23
feat: Improve censored proposals UX

### LINKED ISSUE #2440: Censored record UX improvement
**Record details**
Censored records currently do not display any information about why they were censored.  The gui already has access to this information without needing to send any additional requests, its part of the returned status change metadata, but its not being displayed anywhere.  This should be displayed on the record details page.

- Admin that censored the record
- Reason for censoring the record
- Censored timestamp (like we do for record edits)

**List view**

The list view card for a censored record is currently just a blank card with the author and record token.

image

This should be made to match the abandoned record UX.  
- Grayed out background color.
- Record status tag displayed in the upper right corner. Make this red for censored records (like we do for the `rejected` vote status. 
- Censored at timestamp (like the `abadoned 2 months ago` timestamp)

There are certain scenerios where the client may want to display both abandoned and censored records in the same list, such as the `ineligible` vote status, and using the same backround color with different record status tag colors will allow it to flow nicely while still being able to differentiate between censored and abandoned.

thi4go:
Closes #2440

### Solution description

Include new censored proposal render conditions on `Proposal.jsx`. Add fetch machine to control fetching user data by public key, and include the data needed into the parsed raw proposals object, which we use to render proposals list and details.

### UI Changes Screenshot

Proposal List
image

Proposal Details
image

### DIFF
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -111,8 +111,7 @@ const parseProposalMetadata = (proposal = {}) => {
   return metadata ? JSON.parse(atob(metadata.payload)) : {};
 };
 
-// parseVoteMetadata accepts a proposal object parses it's metadata
-// and returns it as object of the form { linkto, linkby }
+// parseVoteMetadata accepts a proposal object and parses its votes metadata.
 //
 // censored proposals won't have metadata, in this case this function will
 // return an empty object
@@ -123,8 +122,8 @@ const parseVoteMetadata = (proposal = {}) => {
   return metadata ? JSON.parse(atob(metadata.payload)) : {};
 };
 
-// parseUserMetadata accepts a proposal object parses it's metadata
-// and returns it as object of the form { userid, token }
+// parseUserMetadata accepts a proposal object and parses its user plugin
+// metadata.
 //
 // proposals without any user metadata will return an empty object
 const parseUserPluginMetadata = (proposal = {}) =>
@@ -135,8 +134,14 @@ const parseUserPluginMetadata = (proposal = {}) =>
         const parsedPayload = JSON.parse(payload);
         return parsedPayload;
       } catch (e) {
-        // parses metadata payload manually
+        // Parses metadata payload that may be nested without a json parent
+        // on the payload response. This happens when there were more than
+        // one status change to the proposal. Returns a key-value object
+        // where the key is the status for that status change, and the value
+        // is their parsed json payload.
         return compose(
+          reduce((acc, curr) => ({ ...acc, [curr.status]: curr }), {}),
+          filter((md) => Object.keys(md).length),
           map((parsed) => JSON.parse(`{${parsed}}`)),
           splitFp(/\{(.*?)\}/)
         )(payload);
@@ -146,8 +151,8 @@ const parseUserPluginMetadata = (proposal = {}) =>
     get("metadata")
   )(proposal);
 
-// parseProposalIndexFile accepts a proposal object parses it's metadata
-// and returns it as object of the form { description }
+// parseProposalIndexFile accepts a proposal object and parses its index.md
+// file.
 //
 // censored proposals won't have metadata, in this case this function will
 // return an empty object
@@ -158,28 +163,29 @@ const parseProposalIndexFile = (proposal = {}) => {
 };
 
 // parseRawProposal accepts raw proposal object received from BE and parses
-// it's metadata & status changes
+// its metadata & status changes
 export const parseRawProposal = (proposal) => {
-  // Parse metdata
-  // Censored proposal's metadata isn't available
+  // Parse metadatas
   const { name } = parseProposalMetadata(proposal);
   const { linkby, linkto } = parseVoteMetadata(proposal);
   const { description } = parseProposalIndexFile(proposal);
-  const { userid, timestamp } = parseUserPluginMetadata(proposal);
+  const usermds = parseUserPluginMetadata(proposal);
+  const statuschangemsg = usermds[proposal.status]?.message;
   // get prop timestamps
   const { publishedat, censoredat, abandonedat } = getProposalTimestamps(
     proposal,
-    timestamp
+    usermds.timestamp
   );
   return {
     ...proposal,
     description: description || proposal.description,
     name: name || proposal.name,
     linkby,
-    userid: userid || proposal.userid,
+    userid: usermds.userid || proposal.userid,
     linkto,
-    publishedat,
     commentsCount: proposal.commentsCount || 0,
+    statuschangemsg,
+    publishedat,
     censoredat,
     abandonedat
   };
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -189,7 +189,7 @@ export const parseRawProposal = (proposal) => {
     proposal,
     usermds.timestamp
   );
-  
+
   return {
     ...proposal,
     description: description || proposal.description,
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -123,12 +123,19 @@ const parseVoteMetadata = (proposal = {}) => {
 };
 
 // parseUserMetadata accepts a proposal object and parses its user plugin
-// metadata.
+// metadata. Returns a key-value object where the key is the status for
+// that status change, and the value is their parsed json payload. If
+// the metadata was not a status change one, the data will be inserted
+// plainly in the object.
 //
 // proposals without any user metadata will return an empty object
 const parseUserPluginMetadata = (proposal = {}) =>
   compose(
-    reduce((acc, curr) => ({ ...acc, ...curr }), {}),
+    reduce(
+      (acc, curr) =>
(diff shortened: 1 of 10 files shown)


lukebp:
The list item cards appear to still be smaller than an abandoned list item, is that accurate?  If so, we want to make them the same size.  All list items cards should be the same size regardless of what list they're displaying.

thi4go:
@lukebp indeed, the comments icon on the abandoned proposals is giving a larger bottom padding versus the censored proposals. Will fix this before undrafting the PR

lukebp (approved):
ACK

UX looks good.

src/hooks/api/useProposalsStatusChangeUser.js
```diff
@@ -0,0 +1,101 @@
...
+  const onSearchUser = useAction(act.onSearchUser);
+  const resultsByPk = useSelector(sel.queryResultsByPublicKey);
+  const resultsByID = useSelector(sel.searchResultsByID);
+
+  // The public keys for the users we need to search for
+  const unfetchedPublicKeys = uniq(
+    Object.values(proposals).flatMap((prop) =>
+      prop?.status === status ? prop?.statuschangepk : []
+    )
+  ).filter((pk) => pk && !resultsByPk[pk]);
```
victorgcramos, line 34:
```suggestion
  const unfetchedPublicKeys = flow(
    values,
    flatMap((prop) => (prop.status === status ? prop.statuschangepk : [])),
    uniq,
    filter((pk) => pk && !resultsByPk[pk])
  )(proposals);
```

I think you should go for a functional approach here

src/hooks/api/useProposalsStatusChangeUser.js
```diff
@@ -0,0 +1,101 @@
...
+    },
+    initialValues: {
+      status: "idle",
+      loading: true,
+      verifying: true
+    }
+  });
+
+  // Restart machine if there are unfetched public keys left
+  if (!isEmpty(unfetchedPublicKeys) && state.status === "success") {
```
victorgcramos, line 70:
```suggestion
  if (hasUnfetchedPublicKeys && state.status === "success") {
```

src/helpers.js
```diff
@@ -123,20 +122,37 @@ const parseVoteMetadata = (proposal = {}) => {
...
 //
 // proposals without any user metadata will return an empty object
 const parseUserPluginMetadata = (proposal = {}) =>
   compose(
-    reduce((acc, curr) => ({ ...acc, ...curr }), {}),
+    reduce(
+      (acc, curr) =>
+        curr.status ? { ...acc, [curr.status]: curr } : { ...acc, ...curr },
+      {}
+    ),
```
victorgcramos, line 138: same here :)

src/hooks/api/useProposalsStatusChangeUser.js
```diff
@@ -0,0 +1,101 @@
...
+      verify: () => {
+        if (!hasPublicKeys) return send(RESOLVE);
+        const pk = head(publicKeys);
+        onSearchUser({ publickey: pk })
+          .then(() => {
+            setPublicKeys(publicKeys.filter((v) => v !== pk));
+            return send(VERIFY);
+          })
+          .catch((e) => send(REJECT, e));
+        return send(FETCH);
```
victorgcramos, line 58: good job. Very clean!

victorgcramos (changes requested):
Hey @thi4go thanks for the PR. The censored proposals UX looks really good. Abandoned tab is now standardized, and all proposals cards from it are giving the user the correct highlight. Good job!

One thing I'd ask you to take a deeper look into is that the `useProposalsStatusChangeUser` should be loaded once the `proposals` are loaded properly.

Left some inline comments! Thank you.

src/hooks/api/useProposalsStatusChangeUser.js
thi4go: It is using a functional composition approach, but yeah, I suppose we can use lodash's `flow` here to improve readability

src/helpers.js
```diff
@@ -123,20 +122,37 @@ const parseVoteMetadata = (proposal = {}) => {
...
+        // on the payload response. This happens when there were more than
+        // one status change to the proposal.
         return compose(
+          reduce(
+            (acc, curr) =>
+              curr.status
+                ? { ...acc, [curr.status]: curr }
+                : { ...acc, ...curr },
+            {}
+          ),
```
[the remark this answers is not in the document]
thi4go: Function composition and functional approach is the standard of the repo. I don't think we gain declaritiveness by adding a `byStatus` in this case, like we do in our redux store, a global state store which is accessed by a lot of different parts of the code. The function in scope is a helper to yet another helper, `parseRawProposal`, and is only ever used there. Even if we ever come to use the function elsewhere for parsing user metadata decoupled from a proposal, imo it is more readable and clean to write:

```
const usermds = parseUserPluginMetadata(proposal)
const statuschangemd =  usermds[proposal.status]
or
const censoredmd = usermds[PROPOSAL_STATUS_CENSORED]
```
instead of:

```
const usermds = parseUserPluginMetadata(proposal)
const statuschangemd =  usermds.byStatus[proposal.status]
or
const censoredmd = usermds.byStatus[PROPOSAL_STATUS_CENSORED]
```

Seems like added verbosity without much to gain for a simple helper function return, and not a state control object etc. @tiagoalvesdulce would be nice if you drop your thoughts on this before moving forward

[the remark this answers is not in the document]
victorgcramos: > Even if we ever come to use the function elsewhere for parsing user metadata decoupled from a proposal, imo it is more readable and clean to write

I disagree. If you use this function to parse any user metadata, you'll need to read the whole function to know that this "2" is the record status. How am I supposed to know that these numeric keys are about the status?

src/hooks/api/useProposalsStatusChangeUser.js
thi4go: done

src/hooks/api/useProposalsStatusChangeUser.js
```diff
@@ -0,0 +1,101 @@
...
+            : prop;
+        })
+        .reduce(
+          (acc, prop) => ({
+            ...acc,
+            [shortRecordToken(prop?.censorshiprecord?.token)]: prop
+          }),
+          {}
+        )
+    : proposals;
```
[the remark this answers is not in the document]
thi4go: Good match mate, no need for this check here

src/helpers.js
thi4go: Not sure I follow. You mean using `flow` here?

victorgcramos: comment went out of order. This is about the "byStatus" suggestion.

src/hooks/api/useProposalsStatusChangeUser.js
victorgcramos: it's a matter of taste for me. Both will work!

thi4go: checked and we can't use flow in this situation because `flatMap` and `filter` are native prototype Array methods, so they need to be called by an object of type Array, or we get `flatMap` and `filter` are not defined.

thi4go:
@victorgcramos comments addressed. The `useProposalsStatusChangeUser` is already doing this actually. It only starts the fetch machine once we have `unfetchedPublicKeys` set up, which only happens when the proposals batch are properly loaded and passed down to this hook. Otherwise, the fetch machine stays on the resolved state since there are no pk's to search for and no work to be done.

victorgcramos (approved):
tACK!

tiagoalvesdulce:
Seeing this error in the proposal details page:

Screen Shot 2021-07-08 at 09 35 38

tiagoalvesdulce (changes requested):
See comment above.

victorgcramos:
my fault. Forgot to check for undefined values on my code suggestion. Sorry guys.

thi4go:
@tiagoalvesdulce fixed mate

tiagoalvesdulce (changes requested):
As listed on https://github.com/decred/politeiagui/issues/2450, the abandoned and censored cards should have the same height. Currently, they are different. See:

Screen Shot 2021-07-14 at 09 44 29

Screen Shot 2021-07-14 at 09 44 17

victorgcramos:
@tiagoalvesdulce I think that's because proposal title has this line break due to a small screen width, while censored proposals will only display its token, which is shorter. I don't think there's a way to avoid it consistently... Maybe the only thing that could be done in this case, is to have the same title boundaries for every proposal regardless its status tag. Any thoughts?

tiagoalvesdulce:
You're right, I picked a bad example. It's consistent between 1 line and 2 lines titles. We can discuss if we want to break the line or just put a `...` to it in a new issue if the difference in height create issues in the future. This PR LGTM.

tiagoalvesdulce (approved):

Issue #2451, opened 2021-06-19
Flat mode comment reply bug


lukebp:
1. Turn flat mode on.
2. The replies will have a `see in context` link.  Click this link.  It doesn't appear to be working.

Example: https://test-proposals.decred.org/record/89dd08b

PR #2455, opened 2021-06-23
fix: only apply flat mode if comments are not on single thread mode

tiagoalvesdulce:
Closes https://github.com/decred/politeiagui/issues/2451

### Solution description

Easy fix. We shouldn't apply flat mode when on single thread.

### UI Changes Screenshot

**Before**

Screen Shot 2021-06-23 at 11 08 39

**After**
Screen Shot 2021-06-23 at 11 09 02


lukebp (approved):
ACK

Make sure to add the `pi-not-deployed` tag to the PR and issue once merged.

PR #2469, opened 2021-07-08
Proposal submission: Add additional proposal metadata.

### LINKED ISSUE #2459: Add additional proposal metadata.
Update the proposal submission workflow to accommodate the proposal metadata fields added in decred/politeia#1442.

amass01:
This diff adds new proposal metadata fields to the `new proposal` &
`edit proposal` flows.

- Adds the new fields to the proposal form in both create and edit
  proposal flows.
- Displays new fields above the proposal body in the proposal details page.
- Uses pi policy new values to validate new fields and to set start &
  end dates range.
- Adds unit tests.

---

Closes #2459.
Depends on: decred/pi-ui#352.
Requires: decred/politeia#1444.


src/components/ProposalForm/ProposalForm.jsx
```diff
@@ -115,13 +119,23 @@ const ProposalForm = React.memo(function ProposalForm({
     [setFieldValue]
   );
 
-  const selectOptions = useMemo(() => getProposalTypeOptionsForSelect(), []);
+  console.log({ amountmin, amountmax, enddatemax, domains });
```
amassarwi, line 122: grrrrr

src/components/ProposalForm/validation.js
```diff
@@ -52,6 +63,34 @@ export const proposalValidation =
...
+    // start date validation
+    if (!values.sDate) {
+      errors.sDate = "Required";
+    }
+
+    // end date validation
+    if (!values.eDate) {
+      errors.eDate = "Required";
+    }
+
```
amassarwi, line 88: if both start & end dates are provided, ensure start date is smaller.

package.json
```diff
@@ -17,7 +17,7 @@
     "localforage": "^1.9.0",
     "lodash": "^4.17.20",
     "normalize.css": "^8.0.1",
-    "pi-ui": "https://github.com/decred/pi-ui",
+    "pi-ui": "https://github.com/amassarwi/pi-ui#datepickerange",
```
amassarwi, line 20: will be reverted when https://github.com/decred/pi-ui/pull/352 is in.

yarn.lock
```diff
@@ -9788,9 +9788,9 @@ performance-now@^2.1.0:
   resolved "https://registry.yarnpkg.com/performance-now/-/performance-now-2.1.0.tgz#6309f4e0e5fa913ec1c69307ae364b4b377c9e7b"
   integrity sha1-Ywn04OX6kT7BxpMHrjZLSzd8nns=
 
-"pi-ui@https://github.com/decred/pi-ui":
+"pi-ui@https://github.com/amassarwi/pi-ui#datepickerange":
```
amassarwi, line 9791: same

lukebp (changes requested):
The input box should start with $.  A $ should also be added to the invalid amounts.  Numbers should not be displayed without a unit.

image

What's the purpose of putting the start date and end date in the same input box?  This seems like a worse UX for the date picker.  It doesn't appear that you can even choose an end date on mobile.

image

image

amassarwi:
@lukebp you need to run `yarn install` to get latest datepicker version, all this weird bugs were fixed in https://github.com/decred/pi-ui/pull/352

src/actions/api.js
```diff
@@ -832,25 +832,45 @@ export const onSubmitProposal = (
   userid,
   username,
   name,
+  amount,
+  sDate,
+  eDate,
```
victorgcramos, line 837: can we go for a more declarative naming? This is very very minor, but I think that replacing it with the full name will bring a better readability.

victorgcramos (changes requested):
Hey @amassarwi, thanks for the PR. This is working as expected regarding the data submission. Metadata is submitted correctly, editing is nice. I think we could improve the UX in 2 aspects:
1. The `Amount` field label should describe that the amount is in USD like `Amount (USD)` or something, and it shouldn't accept non-numeric values. See:
    Screen Shot 2021-07-20 at 6 22 35 PM

2. I think we could display the metaadata info on the proposals detail page in order to bring a full transparency regarding it. We can only notice the new metadata value changes using the `pictl` CLI. Maybe below the proposal description field, what do you think?
    Screen Shot 2021-07-20 at 6 25 39 PM

lukebp:
I agree about displaying the metadata on the proposal details page.  Good point.

src/components/RecordWrapper/RecordWrapper.jsx
```diff
@@ -358,6 +360,57 @@ export const LinkSection = ({ children, className, title }) => (
...
+    year: startYear
+  } = (startDate && formatUnixTimestampToObj(startDate)) || {};
+  const {
+    day: endDay,
+    month: endMonth,
+    year: endYear
+  } = (endDate && formatUnixTimestampToObj(endDate)) || {};
+
+  return (
+    !!amount &&
```
amassarwi, line 384: this condition is wrong here - currently RFPs not displaying the domain metadata but they should. pushing a fix in sec.

amassarwi: fixed 👍🏽

lukebp (changes requested):
I'm getting this error when I click the "New Proposal" button.

image

lukebp:
Also, please update the original PR message with more details about what this PR does.  I should be able to fully understand all the changes being made by reading the message.

lukebp (changes requested):
This is looking good from a UX standpoint.  Nice work.  I only have a few minor UX nits.

1. Can we add commas to the amount anytime its displayed to the user (amount error message, proposal details amount field)?  Displaying large amounts without commas make them hard to read. 

2. Can the `$` in the amount text box show up all the time instead of only after the first number is input.

3. Can we display the proposal metadata before the proposal text instead of after.  The metadata gives high level details that you would want to know before diving into the specifics of the proposal text.

amassarwi:
@lukebp 

Thanks for the review will do, regarding 2 not sure if that's a good ux as it would replace the field label and users will see
`$` as value instead of the `Amount (USD)` label we currently have. another option would be changing the label to
`Amount ($)` what do you think ?

lukebp:
@amassarwi that makes sense.  The part that find weird is the transition from when the cursor has the input box selected (no $ at this point) to when the user inputs a number and the $ gets added.  There's movement in there that is not smooth.  Try it out and see if you can spot it.

Do you think it would make sense to add the $ once the cursor selects the input box instead of when the user inputs the first number?

amassarwi:
@lukebp I see what you mean, adding the `$` unit on focus is a good i can give it a try and we see how it looks like.

victorgcramos (approved):
LGTM! Good job! 🚀

amassarwi:
@lukebp 
 
1 & 3 addressed; working on 2.

lukebp:
UX looks good.  One last edge case.  How do you think this should be handled?  User inputs the amount with commas.
image

lukebp:
tACK

UX looks good.

src/components/DatePickerField/DatePickerField.jsx
```diff
@@ -1,52 +1,85 @@
...
-const DatePickerField = ({ name, placeholder, years, className }) => {
-  const [isOpen, openPicker, closePicker] = useBooleanState(false);
-  const togglePicker = () => {
-    if (!isOpen) {
-      openPicker();
-    } else {
-      closePicker();
-    }
-  };
+const DatePickerField = ({
```
lukebp, line 9: A minimum start date policy is being added to the backend.  The date picker will need to take this into account for the start and end dates.

lukebp (changes requested):

src/components/DatePickerField/DatePickerField.jsx
amass01: done 👍🏽

lukebp (approved):
tACK

Needs a final code review from @tiagoalvesdulce.

teste2e/cypress/utils.js
```diff
@@ -141,39 +145,18 @@ export const makeProposal = (
...
-      digest: objectToSHA256({
-        name,
-        linkby:
-          type === PROPOSAL_TYPE_RFP
-            ? convertObjectToUnixTimestamp(rfpDeadline)
-            : undefined,
-        linkto: type === PROPOSAL_TYPE_RFP_SUBMISSION ? rfpLink : undefined
-      })
-    }
-  ]
```
tiagoalvesdulce, line 176: Why can we get rid of this?

tiagoalvesdulce (commented):
Looking good. Great job. Left a few comments.

teste2e/cypress/utils.js
amass01: I think this is legacy code, the new proposal route accepts only files array and no metadata - see the one in api.js.

tiagoalvesdulce: ok, cool.

tiagoalvesdulce (approved):

Issue #2471, opened 2021-07-09
Loading placeholders not displayed before initial records batch


victorgcramos:
Noticed that when working on e2e tests for records lists. On the first records batch, the loading placeholders should be displayed in order to provide a good UX, so users can notice the data loading without facing a plank page.

![loading-placeholders-issue](https://user-images.githubusercontent.com/22639213/125139131-5fd7a500-e0e6-11eb-8fd6-d47d5e70aeec.gif)
(slow 3g network)

PR #2473, opened 2021-07-12
e2e: tests for proposals list

victorgcramos:
This diff adds e2e tests for proposals list, including records and inventory paginations assertions, tab switching, and inventory ordering. It also adds the possibility to add middlewares before launching tests, which send mocked data responses for API requests, in order to make a more standardized testing process.

#### Middlewares
A middleware can be interpreted as a request interceptor, described by the `intercept` cypress command. Each middleware is responsible for handling requests and its respective responses, so you can manage to return mocked data instead of relying on the API to have the expected data.

Each plugin/tool on our application can have its own custom middleware, just like this PR does for inventory and records requests.

1. `records.records`: intercepts the `records/v1/records` request and returns mocked proposal data for given proposal tokens and status, instead of server data. 
2. `ticketvote.inventory`: intercepts the `ticketvote/v1/inventory` request and returns random inventory tokens according to the given status list and amount. This can be used to handle the inventory tokens amount without needing to populate the backend with proposals.

Testing a proposal list, specially for pagination, requires the backend to have the required amount of tokens, so we can only test whether the list renders proposals correctly and makes the expected `records` call if our inventory is standardized. With the mocked data introduced by the middlewares, we can manage to do it without needing to have the backend requirements for  the tests to be executed.


teste2e/cypress/e2e/records/recordsList.js
```diff
@@ -0,0 +1,115 @@
...
+      cy.findByTestId("sidebar").should("be.hidden");
+      cy.viewport(1001, 500);
+      cy.findByTestId("sidebar").should("be.visible");
+    });
+
+    // The test below is currently broken. Check issue #2471
+    // it("can render loading placeholders properly", () => {
+    //   cy.visit(`/`);
+    //   cy.assertListLengthByTestId("loading-placeholders", 5);
+    // });
```
tiagoalvesdulce, line 113: #2471 is fixed, can we re-add this?

tiagoalvesdulce (approved):

teste2e/cypress/e2e/records/recordsList.js
victorgcramos: sure!

PR #2479, opened 2021-07-17
fix: Persist public data on the proposals state tree on logout

### LINKED ISSUE #2474: Logout on proposal details triggers infinite loading
This is a minor bug, but when a user logs out from his account when visiting a proposal details page, the UI triggers an inifinite loading. It is fixed by a page refresh, but the user should not need to do that. The expected behaviour is that the UI behaves like the page refresh when the user logs out, by just removing the actions that needs a logged in user.

https://user-images.githubusercontent.com/2729122/125496856-d8d8bfd1-0deb-4f11-bb21-f22d3712d61a.mp4

thi4go:
This diff persists public data from the proposals state tree on user logout event.

closes #2474

### DIFF
--- a/src/reducers/models/proposals.js
+++ b/src/reducers/models/proposals.js
@@ -17,7 +17,8 @@ import {
   CENSORED,
   ARCHIVED,
   INELIGIBLE,
-  PROPOSAL_STATE_UNVETTED
+  PROPOSAL_STATE_UNVETTED,
+  PROPOSAL_STATE_VETTED
 } from "../../constants";
 import {
   shortRecordToken,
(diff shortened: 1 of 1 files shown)


tiagoalvesdulce (changes requested):
You have some lint issues.

tiagoalvesdulce (changes requested):
You also have an error:

Screen Shot 2021-07-08 at 09 35 38

To repro: login > go to vetted props > logout > login again > go to admin props

thi4go:
Hey @tiagoalvesdulce, pushed the lint fix. I think your local repo might be out of sync with this branch/masters. The `unfetchedPublicKeys` on this branch/master contains the correct undefined error handling for that case. It is as follows:

```
const unfetchedPublicKeys = flow(
    Object.values,
    flatMap((prop) => (prop?.status === status ? prop.statuschangepk : [])),
    uniq,
    filter((pk) => pk && !resultsByPk[pk])
  )(proposals);
```

The error path you described tho did indeed error out, good call. I removed an unnecessary clean for the proposals state tree that caused this error.

tiagoalvesdulce (approved):
tACK

PR #2484, opened 2021-07-20
fix: display placeholders while loading initial records batch

### LINKED ISSUE #2471: Loading placeholders not displayed before initial records batch

victorgcramos:
This diff fixes #2471 by adding the initial value for placeholders according to given `pageSize` argument. Default is set to `PROPOSAL_PAGE_SIZE`, which is 5.

### UI Changes Screenshot
**before:**
Screen Shot 2021-07-20 at 7 27 59 PM

**after:**
Screen Shot 2021-07-20 at 7 25 51 PM


tiagoalvesdulce (approved):
ACK

PR #2486, opened 2021-07-21
Plugin architecture

tiagoalvesdulce:
As said in the [Politeia: Pi 2021 Q3 (Proposal System)](https://proposals.decred.org/record/91cfcc8) proposal the frontend should:

> Adapt politeiagui to match the backend plugin architecture

This PR is the core of the repo restructure for the plugins architecture.

# Differences

## Toolchain

Politeiagui currently uses [create-react-app](https://create-react-app.dev/). Create-react-app is great because it doesn't require configuration to get started but it's not very flexible. We even use [react-app-rewired](https://github.com/timarney/react-app-rewired) to do some more advanced stuff but that wouldn't do for a composable plugins architecture.

The new app structure uses a custom config with the webpack bundler. It is also using [yarn workspaces](https://classic.yarnpkg.com/en/docs/workspaces/) and [lerna](https://lerna.js.org/) to manage the packages in our monorepo.

## Folder Structure

Politeiagui currently uses a traditional react/redux folder structure:

```
actions/
    ...
components/
    ...
containers/
    ...
reducers/
    ...
```

This has some big cons:

1. Need to edit multiple files/directories to add a new function.
2. When the codebase grows, a lack of inner structure makes it hard to maintain.
3. The division between dumb components and smart containers doesn't make sense sometimes.

The new app structure uses a `plugin-based structure` as such:

```
plugin-A/
    ...
plugin-B/
    ...
records/
    ...
shared/
    ...
shared-hooks/
    ...
```

The pros are:

1. All plugins are independent.
2. Codebase is easier to scale and maintain.

## State library

We decided to stick with `redux`. Being extremely verbose was a big con of redux in the past but now with [redux-toolkit](https://redux-toolkit.js.org/) we can create slices of state and async thunks without any boilerplate. The `useDispatch` and `useSelector` hooks from `react-redux` allow us to interact with the state anywhere in the component tree. If not using `react` the plugin author can import the `store` from `@politeiagui/core` and use all its functions like `dispatch()` and `getState()` because `redux` is framework agnostic. Redux DevTools is also a big plus.

## Router

The implemented router works using the browser history and location APIs and regex.

When writing a plugin the developer can just export routes in the following format:

```
{
path: string,
view: function that will have the info on how to render the plugin UI,
cleanup (optional): to cleanup the UI from the DOM on unmount
}
```

Then they can be imported in the app shell and concatenated. Something like:

```
import { routes as statisticsRoutes } from "@politeiagui/statistics";
const routes = [...coreRoutes, ...statisticsRoutes];
```

## Creating new state slices

To create new slices in the redux store, the developer can use the injectReducer function exported with the store.

```
  store.injectReducer = (key, asyncReducer) => {
    store.asyncReducers[key] = asyncReducer;
    store.replaceReducer(createReducer(store.asyncReducers));
  };
```

With this util the plugin developer can create a new reducer, attach it to the store and create a whole new state branch.

## Troubleshooting

**Uncaught ReferenceError: regeneratorRuntime is not defined**

Make sure you are using [@babel/plugin-transform-runtime](https://babeljs.io/docs/en/babel-plugin-transform-runtime).

In the last case you can import the regenerator-runtime to the top of the file triggering the error:

`import "regenerator-runtime";`

For more info, see: https://github.com/facebook/regenerator

**SyntaxError: /Users/tiagoalvesdulce/Desktop/stuff/new-pi/packages/core/src/routes/routes.js: Support for the experimental syntax 'jsx' isn't currently enabled:**

Make sure you are using [@babel/preset-react](https://babeljs.io/docs/en/babel-preset-react) correctly.
I am looking into ways to not make plugins dependent on @babel/preset-react. It works nice for
plugins in the monorepo but when using symlinks the exports from `core` will break because some
use `jsx` even if I don't import the code that's using it. 

If you are using react for your plugin, make sure to tell your bundler to expect it to be loaded externally.
This can be done via externals in webpack.

```
 externals: {
    react: 'react',
    "react-dom": 'react-dom',
}
```


plugins-structure/package.json
```diff
@@ -0,0 +1,18 @@
+{
+  "private": true,
+  "workspaces": {
+    "packages": [
+      "records",
+      "plugin-vote",
+      "shared",
+      "shared-hooks"
+    ]
+  },
```
tiagoalvesdulce, line 10: All plugins should be listed here so `lerna` and `yarn` know they should run commands on them.

plugins-structure/plugin-vote/src/components/Votes.js
```diff
@@ -0,0 +1,76 @@
+import React, { useState, lazy, Suspense } from "react";
+import { useVoteInventory } from "../hooks/useVoteInventory";
+const RecordsList = lazy(() => import("records/RecordsList"));
```
tiagoalvesdulce, line 3: Import chunk from remote app

plugins-structure/plugin-vote/src/components/Votes.js
```diff
@@ -0,0 +1,76 @@
...
+
+function VoteStatusList({ status, setListAsFinished, enabled, finished }) {
+  const {isLoading, isIdle, data, fetchNextPage, hasNextPage} = useVoteInventory({ status, enabled });
+  const allRecords = data?.pages.flatMap(page => page.vetted[status] ? page.vetted[status] : []);
+  return isIdle ? null : isLoading ? "Loading..." : (
+    <Suspense fallback={<div>Loading...</div>}>
```
tiagoalvesdulce, line 9: Suspense to wait for lazy chunk

plugins-structure/plugin-vote/webpack.config.js
```diff
@@ -0,0 +1,71 @@
...
+      },
+    ],
+  },
+  plugins: [
+    new ModuleFederationPlugin({
+      name: "plugin-vote",
+      filename: "remoteEntry.js",
+      remotes: {
+        records: "records@https://localhost:3001/remoteEntry.js",
+      },
```
tiagoalvesdulce, line 44: register remote `records` to `plugin-vote`. This means that `plugin-vote` has access to everything that `records` exposes.

plugins-structure/records/webpack.config.js
```diff
@@ -0,0 +1,71 @@
...
+      },
+    ],
+  },
+  plugins: [
+    new ModuleFederationPlugin({
+      name: "records",
+      filename: "remoteEntry.js",
+      exposes: {
+        "./RecordsList": "./src/components/RecordsList.js",
+      },
```
tiagoalvesdulce, line 44: Exposes components/functions/objects to all apps that registered its remote.

victorgcramos (commented):
This is such a great improvement for our current system. It brings more scalability and an easier maintainable code base. Great job.

One thing I'd like to discuss here is about the mutations on our global states. Here's some example:

> we want to get records from our cache by token, which is something we do a lot on our standard architecture. For example, we have a records detail page, and I need to get the record I just fetched from list.

The current draft is storing the records info on our react-query cache by pages, but when we are on details page, we don't have the current page, so querying the data is not possible (is it?). IMO, we should have some state management tool that indexes the records by tokens instead of page, and gets incremented as long as new pages gets fetched.

Any thoughts on this one? How can we customize the way we handle our cache?

I've taken a look at `useMutation` but here's an example of how can we increment the lists on an optimistic way: https://react-query.tanstack.com/examples/optimistic-updates. This is too complicated to handle every time we want to perform some "state" mutation.

Do we have a better and a more readable solution for this? Thanks!

tiagoalvesdulce:
* The complexity on optimistic mutations can be moved to a function.

* When we are querying pages of records we can manually update the cache to store a record by id. For example:

fetching `['records', 'authorized']` > fetched record `b4e3f7f0ddad1212 ` > create `['records', 'b4e3f7f0ddad1212']` and save the info we have.

Now, when I hit the `details` page, I can already show some data and the `record/details/b4e3f7f0ddad1212` will update in the background. You sure can access the cache and reference that record by id. Please notice that there's a difference in `react-query` between the first fetch and the fetches that come when the `useQuery` key is already stale in the cache. The second one is more like an `updating`.

* Also, it is possible to query all the pages if you want. Just a `flatMap`.

IMO a better discussion about react-query is: we want politeiagui to be framework agnostic. Of course we will still use react for our internal plugins but we don't want to force people to use that to build a custom plugin. A client state management library like redux doesn't depend on any framework. That's a +1 for redux. But the way I see it, we would still create `getters` and `setters` and an API to people to interact with state because we don't want to force them to use _any_ library at all. `react-query` has some useful functions we can expose to people to interact with. See [QueryClient](https://react-query.tanstack.com/reference/QueryClient), [QueryCache](https://react-query.tanstack.com/reference/QueryCache) and [MutationCache](https://react-query.tanstack.com/reference/MutationCache). I'm still testing this out but it's looking good. I'll let you know about my new findings.

victorgcramos:
Update: the Webpack module federation plugin is dedicated for run-time applications. However, the run-time requires a multi-deploy infrastructure, since we need to run all plugins on separate ports. This is not a good option in terms of development  experience, because if someone tries to extend the platform, he/she will have trouble configuring the deploy to fetch plugins from a specific domain.

We decided to go for a build-time application with Lerna + Yarn Workspaces, which can coordinate the bundling locally and use the plugins just by adding it into the correct workspace.

Also, we are discussing whether the react-query is suitable for our needs, but we are intended to go for a regular Redux approach, since most devs are familiar with it, and react-query is a little complex (specially for beginners) when you need to manage your state to fit our design.

tiagoalvesdulce:
# Important updates

- Remove `react-query` in favor of `redux` and `redux-toolkit`.
Redux is better suited to us because it's framework agnostic.
A plugin can import the redux store and read state, dispatch actions, etc.
With that we can implement core state and all plugins can access and interact with it.
Redux-toolkit was a nice surprise. I personally didn't like redux because it's verbose.
Redux-toolkit reduces a lot of code boilerplate and repetition.

- Implement vanilla js router.
Important because we want to be framework agnostic.
When writing a plugin the developer can just export routes in the following format:

```
{
path: string,
view: function that will have the info on how to render the plugin UI,
cleanup (optional): to cleanup the UI from the DOM on unmount
}
```

Then they can be imported in the app shell and concatenated. Something like:

```
import { routes as statisticsRoutes } from "@politeiagui/statistics";
const routes = [...coreRoutes, ...statisticsRoutes];
```

Right now, be careful to not initialize the router twice. E.g, you are writing a plugin and
then you define some routes and initializes the router both in the app shell and your plugin.
This might lead to some unexpected behaviors. I would advise to create a `dev` folder with
the router initialization for the plugin that will not export anything and the webpack entrypoint
won't know about (see `core` package for example). I am looking to change the router to a singleton
so that isn't an issue anymore.

- Implement records fetch queue.
One issue we have with the old architecture is managing the inventory/records fetching.
To solve that I created a `recordsFetchQueue` in the `recordsSlice`.

```
   recordsFetchQueue:
   {
    [state]: {
        [status]: [...tokens]
      }
    }
````

The list will be populated every time an `inventory` is fetched:

```
dispatch(
        setRecordsFetchQueue({
          recordsState,
          status,
          records: res[recordsState][status],
        })
      );
```

Then the developer can choose when to fetch slices of that queue limited by the records max page size.
How I am doing it:

https://user-images.githubusercontent.com/13955303/134917098-83641cd1-2094-414f-a97d-b936c0592848.mp4

The inventory has 2 success states. `succeeded/isDone` and `succeeded/hasMore` depending on the size 
of the inventory (if it has length == 20 we know we have more to fetch, otherwise we are done fetching it.)

- The information fetched can be cached among plugins.
Right now, `core` and `statistics` are completely independent plugins that work on their own.
But if I put them in an application shell, the cache can be shared. See:

https://user-images.githubusercontent.com/13955303/134918872-87f6ee72-43f6-4bd5-8338-cbd2053e3271.mp4

Here when I go to statistics, it fetches all records for the first inventory.
Then, when I go back, records already loads them. When I go to statistics again it knows all the records.

## Troubleshooting

**Uncaught ReferenceError: regeneratorRuntime is not defined**

Make sure you are using [@babel/plugin-transform-runtime](https://babeljs.io/docs/en/babel-plugin-transform-runtime).

In the last case you can import the regenerator-runtime to the top of the file triggering the error:

`import "regenerator-runtime";`

For more info, see: https://github.com/facebook/regenerator

**SyntaxError: /Users/tiagoalvesdulce/Desktop/stuff/new-pi/packages/core/src/routes/routes.js: Support for the experimental syntax 'jsx' isn't currently enabled:**

Make sure you are using [@babel/preset-react](https://babeljs.io/docs/en/babel-preset-react) correctly.
I am looking into ways to not make plugins dependent on @babel/preset-react. It works nice for
plugins in the monorepo but when using symlinks the exports from `core` will break because some
use `jsx` even if I don't import the code that's using it. 

If you are using react for your plugin, make sure to tell your bundler to expect it to be loaded externally.
This can be done via externals in webpack.

```
 externals: {
    react: 'react',
    "react-dom": 'react-dom',
}
```

**Review by victorgcramos — COMMENTED** [review]
Nice job, mate! This is looking really good! Using redux + toolkit was a good alternative to make it more generic to other JS frameworks.

One thing i'd like to suggest is to use a better prefix for the plugin slice actions. Instead of `selectRecordsInventoryStatus` use `InventorySelectStatus`, for example... I mean, use the plugin/slice prefix instead of the action.

Will add more comments updating our discussion on matrix, so other devs can participate in the conversation.

**Comment by tiagoalvesdulce** [comment]
I like that, will change. Thanks!

**Review by amass01 — COMMENTED** [review]
Nice work, looks pretty good, I like the plugin based file structure.

One observation:
Could you please update the _state library_ section in the PR description ?

**Comment by tiagoalvesdulce** [comment]
Updated!

**Comment by tiagoalvesdulce** [comment]
I updated the todo list with "Add route UI composability example". Talking with victor, it wasn't clear how that would work. For more context... How would someone compose UI pieces from different plugins in a single route? For example, compose Records Info from the records plugin, Comments section from the comments plugin and TicketVoteBar from the ticketvote plugin into the `/records` route.

We have two options:

1. Compose views in the `app-shell`. Each plugin will export functions telling how to render the UI pieces and the `app-shell` will call them accordingly to compose the view.

2. Compose views in the plugin that defines the route. A plugin will create the view and adapt the view according the configs it will receive.

At first I was more towards the second  but I think the first is more flexible and can keep plugins completely isolated.

**Comment by victorgcramos** [comment]
I'd go for 1 too.

**Comment by lukebp** [comment]
Great progress.  Allowing the plugin implementations to be framework agnostic is a huge plus.  Taking into account the developer and sysadmin experience like you guys are doing (runtime vs buildtime complexity) is also nice to see.

I do have one question.

Do plugins have the ability to extend html elements with custom html/css?  For example, let's say the default configuration uses the record API with no plugins.  The default home page would be a list view of all records.  Then I enable the comments plugin.  Does the comments plugin have the ability to add the comment anchor element onto the list item?  How do you see this working?

image

**Comment by tiagoalvesdulce** [comment]
You are asking about adding a piece of UI to a view defined by other plugin, right?

When we want to just compose views (like adding the comments section below the record details) the application shell can import both views and compose it. But if we want to tell a plugin to render a piece of UI from another plugin that's not that different. That's how I see it:

- Comments plugin exports the `anchor element`;
- The application shell will be responsible to import it and tell the Records plugin it has a piece of UI for it to render;
- The Records plugin receives the element and renders it.

This is the ideal flow, the app-shell coordinating any type of relationship between plugins.

Of course someone could still traverse the DOM and append a html somewhere, but that is discouraged.

**Comment by lukebp** [comment]
On the backend, there are really two different types of plugins.

https://github.com/decred/politeia/blob/master/docs/plugins.md

Does this model make sense for the frontend as well or do you see it working differently?

**Comment by tiagoalvesdulce** [comment]
I updated the `core` exports with [node subpath exports](https://nodejs.org/api/packages.html#packages_subpath_exports) and improved the naming of the exported APIs according was discussed on matrix. By lukebp:

> one thing that go hammers on that I do like is to not use the package name in exported variables from that package, since when you call it other code you're already going to be qualifying the variable name using the package reference
https://talks.golang.org/2014/names.slide#12
strings.StringsReader is redundant. strings.Reader is simpler and easier to read.

With that we could do this change:

```diff
- import {
-  store,
-  selectRecordsInventoryStatus,
-  fetchRecordsInventory,
-  selectRecordsInventoryLastPage,
-  selectRecordsInventoryByStateAndStatus,
-  setRecordsFetchQueue,
-  fetchRecordsNextPage,
-  selectHasMoreRecordsToFetch,
-  selectRecordsByStateAndStatus,
- } from "@politeiagui/core";
+ import { store } from "@politeiagui/core";
+ import { records } from "@politeiagui/core/records";
+ import { recordsInventory } from "@politeiagui/core/records/inventory";
```

And we use, for example, `records.fetchNextPage` instead of `fetchRecordsNextPage`.

tiagoalvesdulce:
> Does this model make sense for the frontend as well or do you see it working differently?

Yes, it makes sense. What I'm calling `application shell` has exactly the same principle of the BE `application plugins`. It's the application shell job to: import all the plugin specific routes, create custom routes composing views from different plugins, pass down custom behavior to plugins via arguments, initialize the router, deal with navigation, deal with theming. This is basically the app that will be bundled and deployed as a final product.

victorgcramos (commented):
### App-Shell Setup Feedback

I tried to create an app that extends the core records plugin and converts the records into proposals (parses its metadata, index.md file...) and I faced some issues when doing it. Most of them was due to the lack of tutorial/docs, but, in terms of development experience, even if the dev doesn't code following the docs, the dev should be able to figure out what's missing and what is he/she doing wrong.

Politeiagui "users" won't be just proposals or cms users anymore, so what we call "user-friendly" on our current design, should be read now as "dev-friendly". Only app-shell devs should worry about a user-friendly interface for the "average Joe".

Here's my example:
```javascript
import React from "react";
import { recordsInventory } from "@politeiagui/core/records/inventory";

const Home = () => {
  const inv = recordsInventory.useFetch({ recordsState: 2, status: 4 });
  console.log(inv);
  return (
    <div>
      HOME PAGE
    </div>
  );
};

export default Home;
```

I'm trying to access the inventory using the `recordsInventory.useFetch` hook, but I got an error saying I need to setup my redux store to make it work.

Here's my `App.js`:
```javascript
import React from "react";
import { Provider as ReduxProvider } from "react-redux";
import Router from "./router";
import { RoutesPublic } from "./pages/Routes";
import { store } from "@politeiagui/core";

const App = () => {
  return (
    <ReduxProvider store={store}>
      <Router>
        <RoutesPublic />
      </Router>
    </ReduxProvider>
  );
};

export default App;
```

But I'm still getting this error which is not well described so I'm not able to debug them easily:
```
recordsInventorySlice.js:97 Uncaught TypeError: Cannot read properties of undefined (reading '4')
```

Also, when I tried to fetch the inventory using just the `recordsInventory.fetch()` method, I noticed that I need to wrap the actions with a `store.dispatch` from the core plugin. I think this wasn't a trivial thing to figure out, so I guess the exported actions should be already wrapped with the regarded providers and HoFs, and work on other plugins without those wrappers.

```javascript
const Home = () => {
  const onLoadProposals = async () => {
    const inv = await store.dispatch(recordsInventory.fetch());
    console.log(inv);
  };

  return (
    <div>
      <button onClick={onLoadProposals}>Load</button>
    </div>
  );
};

export default Home;
```

Getting this response from the action when I try to fetch an inventory without passing any argument:
```javascript
{
    "type": "recordsInventory/fetch/rejected",
    "meta": {
        "requestId": "Scofti2ZQFHErE-5R7Jlj",
        "rejectedWithValue": false,
        "requestStatus": "rejected",
        "aborted": false,
        "condition": false
    },
    "error": {
        "name": "TypeError",
        "message": "Cannot read properties of undefined (reading 'recordsState')",
        "stack": "TypeError: Cannot read properties of undefined (reading 'recordsState')..."
    }
}
```

IMO, since we are creating a composable architecture, we should be using something that would work on a self contained environment, with everything configurable and accessible by interface methods.

I think it's also important to define default values for our core methods, or at least write a good documentation for all methods, including what are the expected values. Error messages also need some improvement in order to help dev debugging.

If you have any doubts, you can DM me on Matrix and we can discuss it.

plugins-structure/packages/core/src/records/inventory/recordsInventorySlice.js
```diff
@@ -0,0 +1,104 @@
...
+// Reducer
+const recordsInventorySlice = createSlice({
+  name: "recordsInventory",
+  initialState,
+  reducers: {},
+  extraReducers(builder) {
+    builder
+      .addCase(fetchRecordsInventory.pending, (state, action) => {
+        const { recordsState, status } = action.meta.arg;
+        state[recordsState][status].status = "loading";
```
victorgcramos, line 69: This would throw an error on our reducer if `state[recordsState]` or `state[recordsState][status]` is undefined. We should prevent this from happening, or at least indicate the correct error message instead of:
`Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'vetted')`

IMO, we should return default values and have a default reducer status because `state`, `status` and `page` are optional, hence they should also work when you don't pass any arg to the reducer, see:

Screen Shot 2021-09-30 at 8 28 46 PM

victorgcramos (commented):
Hey also, I figured out that the inventory doesn't work if you don't pass any arguments to the `fetchRecordsInventory` thunk. We are making an assumption that all inventory calls will include the `recordsState` and `status` arguments, otherwise they will throw an error.

See my inline comment for an example.

plugins-structure/packages/core/src/client/client.js
```diff
@@ -0,0 +1,75 @@
...
+      const error = Error("state is a required parameter");
+      console.error(error);
+      throw error;
+    }
+    const csrf = await getCsrf(state);
+    const body = {
+      requests: records.map((token) => ({
+        token,
+        filenames: ["proposalmetadata.json", "votemetadata.json"],
+      })),
```
victorgcramos, line 26: I know it's a draft, but it's good to keep in mind that those filenames are parameters that are passed by app-shell or plugin, and can't be hardcoded on the core.

Writing it as a reminder 😁

plugins-structure/packages/core/src/records/inventory/recordsInventorySlice.js
tiagoalvesdulce: I decided to make `recordsState` and `status` required so we can make our reducers more consistent. When trying to use `recordsInventory.fetch()` without them, the request will not occur and an error message will be printed in the console. I can't think of a situation when the developer will have to fetch the recordsInventory without knowing the `status` and `recordsState` and if they need to fetch all, they can fire different requests.

tiagoalvesdulce:
@victorgcramos About this: `recordsInventorySlice.js:97 Uncaught TypeError: Cannot read properties of undefined (reading '4')`. That's because `recordsInventory.fetch()` is expecting strings for `recordsState` and `status`. I use the utils `getRecordStateCode` and `getRecordStatusCode` to get the appropriate code. We have two options:

1. Fire an error and log a message if the dev send a number.
2. Accept both numbers and strings.

Of course both option will be documented. What do you prefer?

victorgcramos (commented):
Very nice improvements! Thanks for adding an app-shell example that extends routes and consumes the records core package. Much easier to understand and replicate! Good job! I have one minor feedback regarding the developer experience:

Tried to add records fetch into the politeia app-shell, by using the following on `apps/politeia/src/index.js`:

```javascript
const App = () => {
  const inv = recordsInventory.useFetch({
    recordsState: "vetted",
    status: "public",
  });
  const inventory = recordsInventory.selectByStateAndStatus(store.getState(), {
    recordsState: "vetted",
    status: "public",
  });
  records.useFetch({ inventory, inventoryStatus: inv.status });
  const proposals = records.selectAll(store.getState());
  console.log(inv, inventory, proposals);
  return <h1>Hey</h1>;
};
```

Notice that every time I want to use a store selector, such as `recordsInventory.selectByStateAndStatus` and `records.selectAll`, it's required to pass the `store.getState()` argument for it. Although I think that this approach is more customizable, i.e., you can use the selector for your own state tree, I guess we will only use that feature for edge cases, hence, there would be many pieces of code we would have to pass the same `store.getState()` param to the selector, and it's kinda redundant. Do you agree?

I would suggest you to add a custom `createSelector` method so we can extend plugin selectors and add our custom state, and if we want to add a custom selector for proposals from our own app-state, we should use something like:



```javascript
// you decide the file location 😁 
import { createSelector } from "@politeiagui/core" 
const customSelectorByStateAndStatus = createSelector(
  customState.getState(), records.selectByStateAndStatus
);

// Usage:

const App = () => {
  const customRecords = customRecordsSelectorByStateAndStatus({
    recordsState: "unvetted", status: "public"
  })

  const myRecords = records.selectByStateAndStatus({ recordsState: "unvetted",  status: "public" })
}
```

This way you can also standardize the arguments format for our selectors, where both arguments are required, but due to the object destructuring, the `Uncaught TypeError: Cannot read properties of undefined` error is thrown when you pass no arguments to the selector. (this is minor, but the state and status validation only works if you pass an empty object as a parameter to the selector).

victorgcramos (commented):
This is looking really good, mate! Left some inline comments.

plugins-structure/packages/core/src/router/router.js
```diff
@@ -0,0 +1,152 @@
...
+    if (window.location.pathname === url) return;
+    // Update the page title
+    // Some browsers ignore the second param on pushState
+    if (settings.title) document.title = settings.title;
+    // We are not sending a URL state (first param)
+    // because we don't need it. All info we need in
+    // the route is in the redux store or sent via
+    // query params. All query params are sent to the
+    // view.
+    window.history.pushState(null, settings.title, url);
```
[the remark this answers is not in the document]
tiagoalvesdulce: `data` param? You mean the `state` param? Not using that because we pass route params in the URL and other important state in the store. I can add it but it can be confusing (values conflicting, etc).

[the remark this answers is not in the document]
victorgcramos: Yes, `state`! I'm okay with not adding the `state` there. I just said that because of the hisotry manipulation docs. It could be useful for managing different navigation states without needing to manipulate the store.

tiagoalvesdulce:
@victorgcramos mentioned some issues with importing `@politeiagui/core` methods. We use `exports` to alias our exports, which is not supported by `jest` `resolver` yet: https://github.com/facebook/jest/issues/9771. Should be included on jest 28.

We shouldn't be relying on methods from `@politeiagui/core` on plugins unit tests. Those should be mocked instead. The integration can be tested by cypress on `integration` and `e2e` tests.

plugins-structure/packages/core/src/records/records/recordsSlice.js
```diff
@@ -0,0 +1,229 @@
...
+  { recordsState, status }
+) => {
+  if (validateRecordStateAndStatus(recordsState, status)) {
+    // We have valids record state and status.
+    return Object.values(state.records.records).filter((record) => {
+      return (
+        record.state === getRecordStateCode(recordsState) &&
+        record.status === getRecordStatusCode(status)
+      );
+    });
```
victorgcramos, line 190: After working on records manipulation, I noticed that it will return records alphabetically ordered by tokens. I think the correct approach for getting tokens from inventory is to use the `selectRecordsByTokensBatch` defined below.

Any thoughts on this?

plugins-structure/apps/politeia/src/index.js
```diff
@@ -0,0 +1,94 @@
...
+function handleApi() {
+  const state = store.getState();
+  const status = api.selectStatus(state);
+  if (status === "loading") {
+    document.querySelector("#root").innerHTML = "<h1>Loading api...</h1>";
+  }
+  if (status === "succeeded" && !routerInitialized) {
+    routerInitialized = true;
+    router.init({ routes });
+  }
```
victorgcramos, line 84: I don't think we should add loading handlers here, because we'll need to fetch the API endpoint to renew the CSRF token before login.

tiagoalvesdulce: hmm, fair point. Any idea on where we should move it?

plugins-structure/packages/core/src/records/records/recordsSlice.js
tiagoalvesdulce: fixed!

plugins-structure/apps/politeia/src/index.js
victorgcramos: I think we should only display the `loading api...` message if the transition is `idle -> loading`. Maybe use a flag to control it.

Also, it's worth to mention that the initial successful case should behave differently from other successful cases, because of the same CSRF renewing issue mentioned above. The "app ready" condition is when the api is fetched at least once, right? So I'd expect it to be ready if the state transition is `idle -> loading -> succeeded`. If the api is fetched again, the transition would be `succeeded -> loading -> succeeded`, which shouldn't call `router.init` again.

plugins-structure/packages/core/src/client/client.js
```diff
@@ -0,0 +1,89 @@
...
+    const recordsInfo = await parseResponse(response);
+    return recordsInfo.records;
+  },
+  async fetchApi() {
+    const response = await fetch("/api");
+    const csrf = response.headers.get("X-Csrf-Token");
+    const api = await parseResponse(response);
+    return { api, csrf };
+  },
+  async fetchPolicy(state) {
```
victorgcramos, line 47: what about `fetchRecordsPolicy`?

tiagoalvesdulce: 👍

tiagoalvesdulce:
TODO before merging:

- [ ] Implement `lastTokenPos` on ticketvote (required to finish the Home page)
- [ ] Implement Home page on pi app (https://github.com/tiagoalvesdulce/politeiagui/pull/17)
- [ ] Finish and merge docs (https://github.com/tiagoalvesdulce/politeiagui/pull/5)

tiagoalvesdulce:
Quick explanation why we decided to move from recordsQueue to pointers to the inventory list:

With this approach we had to manage multiple queues and had duplicated data structures
image

With the new approach we have pointers to the inventory list so we don't have to manage multiple queues nor duplicated data.
image

lukebp (approved):
ACK

Let's get @victorgcramos's review on this as well before we merge.

plugins-structure/apps/politeia/src/routes/routes.js
```diff
@@ -0,0 +1,54 @@
...
+  //     ]);
+  //     return ReactDOM.render(
+  //       <Provider store={store}>
+  //         <Details {...params} />
+  //       </Provider>,
+  //       root
+  //     );
+  //   },
+  //   cleanup: () => ReactDOM.unmountComponentAtNode(root),
+  // },
```
victorgcramos, line 53: should we clear this piece of code?

plugins-structure/packages/core/src/client/client.js
victorgcramos: up!

victorgcramos (commented):
Looking really good. Just pushed a quick fix for some common-ui package errors: https://github.com/tiagoalvesdulce/politeiagui/pull/29

Also, left some inline comments :D

plugins-structure/apps/politeia/src/index.js
```diff
@@ -0,0 +1,93 @@
...
+      <Header />
+      <div id="root"></div>
+    </>
+  );
+}
+
+function getRoutes() {
+  let routes = mergeRoutes(proposalsRoutes, coreRoutes, "core");
+  routes = mergeRoutes(routes, statisticsRoutes, "statistics");
+  routes = mergeRoutes(routes, ticketvoteRoutes, "ticketvote");
```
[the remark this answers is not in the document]
tiagoalvesdulce: Yeah! Our plugins are exporting routes so we can dev/test them more easily. In prod we will implement all of our routes on app-shell.

victorgcramos (approved):
LGTM!!

PR #2494, opened 2021-07-22
feat: upgrade react-markdown and block comments headers rendering

### LINKED ISSUE #2476: Comment text size should be consistent
image

The ability to make certain text in your comment appear bigger than normal should not be allowed.  Syntax like making text bold or italic is fine, but the size of the text should be the same for all comment text.

Example: https://proposals.decred.org/record/ae609f1/comments/15

The gui should ignore markdown sytax that changes the text size.

victorgcramos:
This diff closes #2476, an issue that reported some unexpected header rendering on comments section, by adding the `disallowedElements` prop into the `Markdown` component, which blocks a list of elements from being rendered.

### Solution description

In order to add the `disallowedElements` prop, the `react-markdown` package needed to be updated to version 6.0.2. Before updating it, I've researched for possible XSS vulnerabilities [see this page on synk.io](https://snyk.io/test/npm/react-markdown/6.0.2), and I couldn't find any possible vulnerability that could be exploited.

So, after updating it, I also tested the following payload from [cujanovic's Mardown XSS payloads](https://github.com/cujanovic/Markdown-XSS-Payloads/blob/master/Markdown-XSS-Payloads.txt), and also no  vulnerabilities were found.

### UI Changes Screenshot
**before**:
Screen Shot 2021-07-22 at 7 38 51 PM

**after**:
Screen Shot 2021-07-22 at 7 39 58 PM

### DIFF
--- a/src/components/Markdown/helpers.js
+++ b/src/components/Markdown/helpers.js
@@ -1,14 +1,7 @@
 import React from "react";
-import htmlParser from "react-markdown/plugins/html-parser";
 import ModalExternalLink from "../ModalExternalLink";
 import useModalContext from "src/hooks/utils/useModalContext";
 import xssFilters from "xss-filters";
-
-export const htmlParserRules = htmlParser({
-  isValidNode: (node) => {
-    return node.type !== "script";
-  }
-});
 
 export const traverseChildren = (el, cb) => {
   const filterChildren = (c) =>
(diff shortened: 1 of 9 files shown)


src/components/Markdown/helpers.js
```diff
@@ -32,8 +25,29 @@ export const traverseChildren = (el, cb) => {
...
+  if (typeof el === "string") return el;
+  const props = el.props;
+  if (!props) {
+    return el;
+  }
+  const newProps = {
+    ...props
+  };
+  if (newProps.url) {
+    console.log("aparece aqui");
```
tiagoalvesdulce, line 38: log

tiagoalvesdulce (approved):
tACK

src/components/Markdown/helpers.js
victorgcramos: damn, sorry.

PR #2500, opened 2021-07-27
feat: add tooltip to warn user about draft issues

### LINKED ISSUE #2496: Add warning to markdown editor
When you create a proposal draft the data is saved locally to the user's browser until they submit the proposal.  If something goes wrong with the user's browser or computer, it is not uncommon for the proposal draft to be deleted and unrecoverable.  Issues like these are tricky from a dev standpoint because there are variables specific to the user's local setup that are likely contributing to the issue and make it hard to identify the root cause or if there's even something that can be done about it.

A warning message should be added to the markdown editor says:

"Drafts are saved locally to your browser and are not recoverable if something goes wrong.  We recommend drafting the content offline then using the editor to submit the final version."

victorgcramos:
This diff closes #2496 by adding the warning on `DraftSaver` component.

### UI Changes Screenshot
Screen Shot 2021-07-27 at 6 23 55 PM

### DIFF
--- a/src/components/ProposalForm/DraftSaver.jsx
+++ b/src/components/ProposalForm/DraftSaver.jsx
@@ -1,5 +1,5 @@
 import React, { useEffect, useState } from "react";
-import { Button } from "pi-ui";
+import { Button, Tooltip } from "pi-ui";
 import { FormikConsumer } from "formik";
 import { useDraftProposals } from "src/containers/Proposal/User/hooks";
 import { getQueryStringValue, setQueryStringValue } from "src/lib/queryString";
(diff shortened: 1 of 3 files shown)


lukebp:
I don't think this should be a tooltip.  People won't see it until they go to save the draft, which is too late a lot of the time.  I think this should be some type of banner.

@tiagoalvesdulce thoughts?

victorgcramos:
what about that, @lukebp?
Screen Shot 2021-07-27 at 7 12 01 PM

lukebp:
That's not really going to be noticable.  We want to prevent the user from writing their proposal in the markdown editor.  It needs to be very apparent to somebody starting to write a new proposal.

victorgcramos:
maybe this?

Screen Shot 2021-07-27 at 7 18 17 PM

lukebp:
I like that a lot.

tiagoalvesdulce (approved):
LGTM!

PR #2520, opened 2021-08-07
feat: Comment vote optimistic updates.

### LINKED ISSUE #2517: Comment vote highlighting
When you upvote/downvote a comment, the icon does not become highlighted until a succesful response is received from the server.  This creates the appearance of lag, especially when there is larger than normal latency in the server reply.

The way this should work is that the upvote/downvote icon should become highlighted on click.  If the request fails for some reason then it should be removed.

Failed comment votes are the edge case and we should design the UX to reflect that.

victorgcramos:
This diff closes #2517, an issue that requested the implementation of optimistic updates on comment vote actions.

### Solution description
In order to apply the optimistic updates on comment vote actions with redux, the state tree must be updated before the response from server gets returned. So, to handle possible errors, the previous comment count should also be stored on the state tree, so we can undo the changes we made. Therefore, when the server returns an error, the state is replaced by the backup, then the error is displayed.

### UI Changes Screenshot

Preview for offline mode errors:
2517-comment-vote-highlight


lukebp (approved):
ACK

tiagoalvesdulce (changes requested):
https://user-images.githubusercontent.com/13955303/128732308-c0aee873-7873-4807-964b-48179f316f89.mp4

1. IMO we can remove the error message only if the next action succeeds. This flicker is a bit annoying. What do you think?
2. Since we are optimistically updating the button highlighting, we should do it for the number of votes as well. You can see that if we have a lot of comments, the delay is hurting the UX.

lukebp:
> This flicker is a bit annoying.

Good point.  Agreed.

> we should do it for the number of votes as well

Another good catch.  Agreed.

victorgcramos:
I added a new cache layer to comments likes, because it gets updated more frequently, and the comments array updates were expensive in terms of performance. This way we get an O(1) complexity instead of O(n).

You also might notice some delay on the count update in some cases in comparison to the 👍 highlighting. This happens because the hovering action highlights it, but if you move your cursor outside the 👍 box, you can test whether it gets optimistically updated.

lukebp (changes requested):
I'm seeing this bug.  Upvote a comment.  Downvote the same comment.  The score is messed up.  This happens when going the other way as well (downvote then upvote).

image

A user is only allowed to change their vote preference 5 times.  Attempting to change it again results in this error.  This behavior is correct, but the error message needs to be human readable.

image

victorgcramos:
Okay, noticed that too. Already fixing it. My bad.

victorgcramos:
Tried to avoid all listed edge-cases. Also, added e2e tests for comment vote actions that simulates a delayed network, and error responses. Comment data is mocked, so we don't need to set it up before running tests.

victorgcramos:
it prevents multi clicking, which causes many edge-cases regarding optimistic updates

lukebp:
You think that's an acceptable UX?  Really?  Do any other major websites do that?  No, because it's a terrible UX.  Come on, this obvious stuff.  Fix it.

lukebp:
And when I say fix it, I don't just mean take the icon off and introduce edge cases that cause bugs.  Find a solution that has a good UX and that handles the edge cases properly.

victorgcramos:
ok...

lukebp:
I understand that I may be coming off a bit harsh, but we need to raise the bar for politeiagui development.  We are getting crushed by bugs and UX improvements that should have ever made it into the codebase the first time.  There's going to be a higher standard moving foward.

A simple question you can ask yourselve regarding UX changes is, how does reddit or twitter or github or the other major platforms do it?  They employ tons of UX people and designers.  Just copy what they do.  If we want to deviate, we can decide that at a later date once its already been implemented using the industry norm UX.

src/components/Likes/Likes.jsx
```diff
@@ -17,22 +18,26 @@ const Likes = ({ upLikes, downLikes, onLike, onDislike, option, disabled }) => {
...
   }
 
-  const debouncedHandleDislike = debounce(handleDislike, 150);
-  const debouncedHandleLike = debounce(handleLike, 150);
+  // Avoid multi-clicking actions
+  const handleDebounceVote = (voteFn) => () => {
+    if (tempDisabled || disabled) return;
+    setTempDisabled(true);
+    debounce(voteFn, 150)();
+  };
```
tiagoalvesdulce, line 35: This is probably breaking the debounce making you add the not-allowed cursor to prevent multiple clicks. The debounced function should be passed to the eventListener.

tiagoalvesdulce (approved):
The code is looking good. Victor only missed a small detail about debounced functions (they are quite tricky).

src/components/Likes/Likes.jsx
victorgcramos: yeah you're right! Thanks, mate!

tiagoalvesdulce:
LGTM

lukebp (approved):
Much better.  The UX is looking good.

tACK

tiagoalvesdulce (approved):
LGTM

PR #2524, opened 2021-08-10
Allow user to select diff versions

### LINKED ISSUE #2466: Allow user to select diff versions
I think one way to improve the UX of the diff viewer would be to do the same thing github does for its diff viewer. 

image

On initial click of `version x` the diff viewer would do what it currently does, which is show the diff between the `selectedVersion` and the `selectedVersion - 1`.  The diff modal would have two dropdowns though that are populated with the versions being compared and that let the user choose what versions they want to compare.  This would improve the UX by explicity stating what versions are being compared and by also giving the user more fine grained control.

vibros68:
closed #2466 

* **The old behavior is**: Click on version selector. Choose the version, the selector become loading cycle. After loaded will open a modal. 
* **The new behavior**: Click on the version selector, open the modal immediately, the loading indicator is inside the modal. after the data is fetched, will show the data. And when the user change the version we show the loading indicator again.
* **Required**:  This pull request requires [pi-ui/pull/357](https://github.com/decred/pi-ui/pull/357) be merged
Preview:
Screenshot from 2021-08-19 22-37-52
Screenshot from 2021-08-19 22-38-08

### DIFF
--- a/src/components/ModalDiff/CompareVersionSelector.jsx
+++ b/src/components/ModalDiff/CompareVersionSelector.jsx
@@ -24,15 +24,16 @@ const CompareVersionSelector = ({
   const iconColor = isDarkTheme ? darkIconColor : undefined;
   const versionsOptions = useMemo(() => {
     const versions = [];
-    for (let index = latest; index >= 1; index--) {
+    for (let index = latest; index >= 0; index--) {
       versions.push(index);
     }
     return versions;
   }, [latest]);
+  const ZERO_VERSION_ALIAS = "----------";
   return (
     <div className={styles.versionSelectorContainer}>
       <Dropdown
-        title={`version ${base}`}
+        title={base ? `version ${base}` : ZERO_VERSION_ALIAS}
         className={styles.versionSelectorWrapper}
         itemsListClassName={className}>
         {versionsOptions.map((v) => {
@@ -43,7 +44,7 @@ const CompareVersionSelector = ({
                 onClick={() => {
                   onChange(BASE, v);
                 }}>
-                version {v}
+                {v ? `version ${v}` : ZERO_VERSION_ALIAS}
               </DropdownItem>
             )
           );
--- a/src/components/ModalDiff/CompareVersionSelector.jsx
+++ b/src/components/ModalDiff/CompareVersionSelector.jsx
@@ -34,20 +34,20 @@ const CompareVersionSelector = ({
   return (
     <div className={styles.versionSelectorContainer}>
       <Dropdown
-          title={`version ${base}`}
-          className={styles.versionSelectorWrapper}
-          itemsListClassName={className}>
+        title={`version ${base}`}
+        className={styles.versionSelectorWrapper}
+        itemsListClassName={className}>
         {versionsOptions.map((v) => {
           return (
-              compare > v && (
-                  <DropdownItem
-                      key={v}
-                      onClick={() => {
-                        onChange(BASE, v);
-                      }}>
-                    version {v}
-                  </DropdownItem>
-              )
+            compare > v && (
+              <DropdownItem
+                key={v}
+                onClick={() => {
+                  onChange(BASE, v);
+                }}>
+                version {v}
+              </DropdownItem>
+            )
           );
         })}
       </Dropdown>
@@ -57,20 +57,20 @@ const CompareVersionSelector = ({
         type="compare"
       />
       <Dropdown
-          title={`version ${compare}`}
-          className={styles.versionSelectorWrapper}
-          itemsListClassName={className}>
+        title={`version ${compare}`}
+        className={styles.versionSelectorWrapper}
+        itemsListClassName={className}>
         {versionsOptions.map((v) => {
           return (
-              base < v && (
-                  <DropdownItem
-                      key={v}
-                      onClick={() => {
-                        onChange(COMPARE, v);
-                      }}>
-                    version {v}
-                  </DropdownItem>
-              )
+            base < v && (
+              <DropdownItem
+                key={v}
+                onClick={() => {
+                  onChange(COMPARE, v);
+                }}>
+                version {v}
+              </DropdownItem>
+            )
           );
         })}
       </Dropdown>
--- a/src/components/ModalDiff/CompareVersionSelector.jsx
+++ b/src/components/ModalDiff/CompareVersionSelector.jsx
@@ -23,7 +23,7 @@ const CompareVersionSelector = ({
   const isDarkTheme = themeName === DEFAULT_DARK_THEME_NAME;
   const darkIconColor = getThemeProperty(theme, "text-color");
   const iconColor = isDarkTheme ? darkIconColor : undefined;
-  const versionsOptions = useMemo(() => rangeRight(latest), [latest]);
+  const versionsOptions = useMemo(() => rangeRight(latest + 1), [latest]);
 
   return (
     <div className={styles.versionSelectorContainer}>
--- a/src/components/ModalDiff/CompareVersionSelector.jsx
+++ b/src/components/ModalDiff/CompareVersionSelector.jsx
@@ -4,7 +4,6 @@ import {
   Dropdown,
   DropdownItem,
   getThemeProperty,
-  Spinner,
   useTheme
 } from "pi-ui";
 import styles from "./ModalDiff.module.css";
@@ -33,60 +32,48 @@ const CompareVersionSelector = ({
     return versions;
   }, [latest]);
   return (
-    <div>
-      {baseLoading ? (
-        <span className={styles.versionSelectorWrapper}>
-          {`version ${base}`} <Spinner invert />
-        </span>
-      ) : (
-        <Dropdown
+    <div className={styles.versionSelectorContainer}>
+      <Dropdown
           title={`version ${base}`}
           className={styles.versionSelectorWrapper}
           itemsListClassName={className}>
-          {versionsOptions.map((v) => {
-            return (
+        {versionsOptions.map((v) => {
+          return (
               compare > v && (
-                <DropdownItem
-                  key={v}
-                  onClick={() => {
-                    onChange(BASE, v);
-                  }}>
-                  version {v}
-                </DropdownItem>
+                  <DropdownItem
+                      key={v}
+                      onClick={() => {
+                        onChange(BASE, v);
+                      }}>
+                    version {v}
+                  </DropdownItem>
               )
-            );
-          })}
-        </Dropdown>
-      )}
+          );
+        })}
+      </Dropdown>
       <IconButton
         className={styles.versionCompareIcon}
         iconColor={iconColor}
         type="compare"
       />
-      {compareLoading ? (
-        <span className={styles.versionSelectorWrapper}>
-          {`version ${compare}`} <Spinner invert />
-        </span>
-      ) : (
-        <Dropdown
+      <Dropdown
           title={`version ${compare}`}
           className={styles.versionSelectorWrapper}
           itemsListClassName={className}>
-          {versionsOptions.map((v) => {
-            return (
+        {versionsOptions.map((v) => {
+          return (
               base < v && (
-                <DropdownItem
-                  key={v}
-                  onClick={() => {
-                    onChange(COMPARE, v);
-                  }}>
-                  version {v}
-                </DropdownItem>
+                  <DropdownItem
+                      key={v}
+                      onClick={() => {
+                        onChange(COMPARE, v);
+                      }}>
+                    version {v}
+                  </DropdownItem>
               )
-            );
-          })}
-        </Dropdown>
-      )}
+          );
+        })}
+      </Dropdown>
     </div>
   );
 };
--- a/src/components/ModalDiff/CompareVersionSelector.jsx
+++ b/src/components/ModalDiff/CompareVersionSelector.jsx
@@ -9,7 +9,8 @@ import {
 import styles from "./ModalDiff.module.css";
 import IconButton from "../IconButton";
 import PropTypes from "prop-types";
-import { COMPARE, BASE } from "./const";
+import { COMPARE, BASE, ZERO_VERSION_ALIAS } from "./constants";
+import rangeRight from "lodash/rangeRight";
 
 const CompareVersionSelector = ({
   onChange,
@@ -22,22 +23,16 @@ const CompareVersionSelector = ({
   const isDarkTheme = themeName === DEFAULT_DARK_THEME_NAME;
   const darkIconColor = getThemeProperty(theme, "text-color");
   const iconColor = isDarkTheme ? darkIconColor : undefined;
-  const versionsOptions = useMemo(() => {
-    const versions = [];
-    for (let index = latest; index >= 0; index--) {
-      versions.push(index);
-    }
-    return versions;
-  }, [latest]);
-  const ZERO_VERSION_ALIAS = "----------";
+  const versionsOptions = useMemo(() => rangeRight(latest), [latest]);
+
   return (
     <div className={styles.versionSelectorContainer}>
       <Dropdown
         title={base ? `version ${base}` : ZERO_VERSION_ALIAS}
         className={styles.versionSelectorWrapper}
         itemsListClassName={className}>
-        {versionsOptions.map((v) => {
-          return (
+        {versionsOptions.map(
+          (v) =>
             compare > v && (
               <DropdownItem
                 key={v}
@@ -47,8 +42,7 @@ const CompareVersionSelector = ({
                 {v ? `version ${v}` : ZERO_VERSION_ALIAS}
               </DropdownItem>
             )
-          );
-        })}
+        )}
       </Dropdown>
       <IconButton
         className={styles.versionCompareIcon}
@@ -62,11 +56,7 @@ const CompareVersionSelector = ({
         {versionsOptions.map((v) => {
           return (
             base < v && (
-              <DropdownItem
-                key={v}
-                onClick={() => {
-                  onChange(COMPARE, v);
-                }}>
+              <DropdownItem key={v} onClick={() => onChange(COMPARE, v)}>
                 version {v}
               </DropdownItem>
             )
--- a/src/components/ModalDiff/CompareVersionSelector.jsx
+++ b/src/components/ModalDiff/CompareVersionSelector.jsx
@@ -1,0 +1,102 @@
+import React, { useMemo } from "react";
+import {
+  DEFAULT_DARK_THEME_NAME,
+  Dropdown,
+  DropdownItem,
+  getThemeProperty,
+  Spinner,
+  useTheme
+} from "pi-ui";
+import styles from "./ModalDiff.module.css";
+import IconButton from "../IconButton";
+import PropTypes from "prop-types";
+import { COMPARE, BASE } from "./const";
+
+const CompareVersionSelector = ({
+  onChange,
+  className,
+  latest,
+  base,
+  compare,
+  baseLoading,
+  compareLoading
+}) => {
+  const { theme, themeName } = useTheme();
+  const isDarkTheme = themeName === DEFAULT_DARK_THEME_NAME;
+  const darkIconColor = getThemeProperty(theme, "text-color");
+  const iconColor = isDarkTheme ? darkIconColor : undefined;
+  const versionsOptions = useMemo(() => {
+    const versions = [];
+    for (let index = latest; index >= 1; index--) {
+      versions.push(index);
+    }
+    return versions;
+  }, [latest]);
+  return (
+    <div>
+      {baseLoading ? (
+        <span className={styles.versionSelectorWrapper}>
+          {`version ${base}`} <Spinner invert />
+        </span>
+      ) : (
+        <Dropdown
+          title={`version ${base}`}
+          className={styles.versionSelectorWrapper}
+          itemsListClassName={className}>
+          {versionsOptions.map((v) => {
+            return (
+              compare > v && (
+                <DropdownItem
+                  key={v}
+                  onClick={() => {
+                    onChange(BASE, v);
+                  }}>
+                  version {v}
+                </DropdownItem>
+              )
+            );
+          })}
+        </Dropdown>
+      )}
+      <IconButton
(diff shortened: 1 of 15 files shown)


lukebp:
It's not just flickering.  The loading icons are also messed and the transitions are not smooth at all.

vibros68:
https://user-images.githubusercontent.com/84569563/129439932-068b4def-6140-4378-9379-0088ad3bce36.mp4

The bug of flickering is exist in the master before. The cause is when change a state, react will re-render the modal instead of re-render the component inside the modal

vibros68:
The bug of flickering is exist in the master before. The cause is when change a state, react will re-render the modal instead of re-render the component inside the modal

https://user-images.githubusercontent.com/84569563/129440100-2d8b91dc-a00d-445d-8438-3faf2cedbe3d.mp4

vibros68:
The flickering is exist on master. I have opened a new issue for it: #2539 . I think it should be fixed on a separated pull request

lukebp (changes requested):
Very nice.  The UX of this pull request is looking much better.

The one comment that I have is that I think we should have a header in the modal that explains what the modal does.  It might not be obvious to first time users.  I would add something similar to the github compare page.

image

```
Compare versions
Compare changes across different versions of the record.
```

lukebp (changes requested):
1. Can you add some more spacing in between the version dropdowns and the proposal title?  It's too close together and looks weird.

image

2. I get the error `TypeError: baseProposal is undefined` when I select `version 1` from the proposal details dropdown (not the modal).  I would expect it to pull up the model with a `-` in the first version dropdown and `version 1` in the second version dropdown.  The diff would look like the existing behavior for a `version 1` on test-proposals.decred.org where the entire record is highlighted in green.

image

3. I think that the `-` option that I listed in the comment above should be one of the options for the first version dropdown in the modal.  Example, when viewing the diff between `-` and `version 3` it will show the full record contents of `version 3` as being new (highlighted in green).  This will allow for users to not only view the diff between two separate versions, but also view the entire contents of a specific version if they want to.

lukebp (commented):
tACK

UX is looking good.  This is ready for a code review @tiagoalvesdulce.

src/components/ModalDiff/CompareVersionSelector.jsx
```diff
@@ -0,0 +1,88 @@
...
+    return versions;
+  }, [latest]);
+  const ZERO_VERSION_ALIAS = "----------";
+  return (
+    <div className={styles.versionSelectorContainer}>
+      <Dropdown
+        title={base ? `version ${base}` : ZERO_VERSION_ALIAS}
+        className={styles.versionSelectorWrapper}
+        itemsListClassName={className}>
+        {versionsOptions.map((v) => {
```
amass01, line 39:
```suggestion
        {versionsOptions.map((v) => 
```

src/components/ModalDiff/CompareVersionSelector.jsx
```diff
@@ -0,0 +1,88 @@
...
+          return (
```
amass01, line 40:
```suggestion
           (
```

src/components/ModalDiff/CompareVersionSelector.jsx
```diff
@@ -0,0 +1,88 @@
...
+      <Dropdown
+        title={`version ${compare}`}
+        className={styles.versionSelectorWrapper}
+        itemsListClassName={className}>
+        {versionsOptions.map((v) => {
+          return (
+            base < v && (
+              <DropdownItem
+                key={v}
+                onClick={() => {
```
amass01, line 67:
```suggestion
                onClick={() => 
```

src/components/ModalDiff/DiffProposal.jsx
```diff
@@ -0,0 +1,168 @@
...
+    changedVersion,
+    baseProposal,
+    compareProposal
+  } = useCompareVersionSelector(initVersion, latest, token);
+
+  const [activeTabIndex, setActiveTabIndex] = useState(0);
+  useEffect(() => {
+    setActiveTabIndex(0);
+  }, [props.show]);
+  return (
```
amass01, line 32: add an empty line above the return statement.

src/components/ModalDiff/DiffProposal.jsx
```diff
@@ -0,0 +1,168 @@
...
+            title={
+              <p className={styles.titleDiff}>
+                <DiffText
+                  oldText={baseProposal.title}
+                  newText={compareProposal.title}
+                />
+              </p>
+            }
+            subtitle={
+              <JoinTitle className={"margin-top-s"} separatorSymbol={":"}>
```
amass01, line 65:
```suggestion
              <JoinTitle className="margin-top-s" separatorSymbol=":">
```

amass01: no need for `{}` when passing a string as a param.

src/components/ModalDiff/DiffProposal.jsx
```diff
@@ -0,0 +1,168 @@
...
+                          <Event
+                            className="margin-left-s margin-right-s"
+                            event="edited"
+                            timestamp={baseProposal.details.timestamp}
+                          />
+                        )}
+                      {baseProposal.details.abandonedat && (
+                        <Event
+                          className="margin-left-s margin-right-s"
+                          event={"abandoned"}
```
amass01, line 86: same

src/components/ModalDiff/DiffProposal.jsx
```diff
@@ -0,0 +1,168 @@
...
+                />
+              </p>
+            }
+            subtitle={
+              <JoinTitle className={"margin-top-s"} separatorSymbol={":"}>
+                <Author
+                  username={compareProposal.details.username}
+                  url={`/user/${compareProposal.details.userid}`}
+                />
+                <JoinTitle separatorSymbol={"-"}>
```
amass01, line 70: same

src/components/ModalDiff/DiffProposal.jsx
```diff
@@ -0,0 +1,168 @@
...
+                        <Event
+                          className="margin-left-s margin-right-s"
+                          event="edited"
+                          timestamp={compareProposal.details.timestamp}
+                        />
+                      )}
+                    {compareProposal.details.abandonedat && (
+                      <Event
+                        className="margin-left-s margin-right-s"
+                        event={"abandoned"}
```
amass01, line 106: same

src/components/ModalDiff/hook.js
```diff
@@ -0,0 +1,104 @@
...
+        setError(e);
+        setBaseLoading(false);
+      });
+  }, [
+    token,
+    baseVersion,
+    fetchProposalVersions,
+    onFetchProposalDetailsWithoutState
+  ]);
+  useEffect(() => {
```
amass01, line 72: add empty line

src/components/ModalDiff/hook.js
```diff
@@ -0,0 +1,104 @@
...
+  );
+  const changedVersion = (versionType, v) => {
+    if (versionType === COMPARE) {
+      setCompareVersion(v);
+    }
+    if (versionType === BASE) {
+      setBaseVersion(v);
+    }
+  };
+  useEffect(() => {
```
amass01, line 50: add empty line

src/components/ModalDiff/hook.js
```diff
@@ -0,0 +1,104 @@
...
+      return {
+        details: proposal,
+        files: getAttachmentsFiles(proposal.files),
+        text: description,
+        title: name
+      };
+    },
+    []
+  );
+  const changedVersion = (versionType, v) => {
```
amass01, line 42: add empty line

src/components/ModalDiff/hook.js
```diff
@@ -0,0 +1,104 @@
...
+  const [compareVersion, setCompareVersion] = useState(initVersion);
+  const [baseLoading, setBaseLoading] = useState(false);
+  const [compareLoading, setCompareLoading] = useState(false);
+  const [compareProposal, setCompareProposal] = useState({});
+  const [baseProposal, setBaseProposal] = useState({});
+  const [error, setError] = useState();
+  const onFetchProposalDetailsWithoutState = useAction(
+    act.onFetchProposalDetailsWithoutState
+  );
+  const fetchProposalVersions = useCallback(
```
amass01, line 18: add empty line

src/components/ModalDiff/hook.js
```diff
@@ -0,0 +1,104 @@
...
+
+export function useCompareVersionSelector(initVersion, latestVersion, token) {
+  const [baseVersion, setBaseVersion] = useState(initVersion - 1);
+  const [compareVersion, setCompareVersion] = useState(initVersion);
+  const [baseLoading, setBaseLoading] = useState(false);
+  const [compareLoading, setCompareLoading] = useState(false);
+  const [compareProposal, setCompareProposal] = useState({});
+  const [baseProposal, setBaseProposal] = useState({});
+  const [error, setError] = useState();
+  const onFetchProposalDetailsWithoutState = useAction(
```
amass01, line 15: i'd add also add empty line before declaring a function which is not a oneliner.

src/components/ModalDiff/hook.js
```diff
@@ -0,0 +1,104 @@
...
+        setError(e);
+        setCompareLoading(false);
+      });
+  }, [
+    token,
+    compareVersion,
+    fetchProposalVersions,
+    onFetchProposalDetailsWithoutState
+  ]);
+  return {
```
amass01, line 94: also here empty line before return

amass01 (commented):
Left some minors, code looks good over all.

src/components/ModalDiff/CompareVersionSelector.jsx
```diff
@@ -0,0 +1,88 @@
...
+  const darkIconColor = getThemeProperty(theme, "text-color");
+  const iconColor = isDarkTheme ? darkIconColor : undefined;
+  const versionsOptions = useMemo(() => {
+    const versions = [];
+    for (let index = latest; index >= 0; index--) {
+      versions.push(index);
+    }
+    return versions;
+  }, [latest]);
+  const ZERO_VERSION_ALIAS = "----------";
```
victorgcramos, line 32: You can put this outside the component. It's a constant, and this will cause the attribution to happen every render.

src/components/ModalDiff/hook.js
```diff
@@ -0,0 +1,104 @@
+import { useState, useCallback, useEffect } from "react";
```
victorgcramos, line 1: This file should be named `hooks.js` instead of `hook`. All hooks files are named in plural, so we should follow this standard.

src/components/ModalDiff/const.js
```diff
@@ -0,0 +1,2 @@
+export const COMPARE = "compare";
```
victorgcramos, line 1: Again, naming should follow our standards. Use `constants.js` instead of `const.js`

src/components/ModalDiff/hook.js
```diff
@@ -0,0 +1,104 @@
...
+import { COMPARE, BASE } from "./const";
+import * as act from "src/actions";
+import { useAction } from "../../redux";
+import { getAttachmentsFiles, parseRawProposal } from "../../helpers";
```
victorgcramos, line 5:
```suggestion
import { getAttachmentsFiles, parseRawProposal } from "src/helpers";
```

src/components/ModalDiff/hook.js
```diff
@@ -0,0 +1,104 @@
...
+
+export function useCompareVersionSelector(initVersion, latestVersion, token) {
```
victorgcramos, line 7: `latestVersion` is never used. Either use it, or delete it.

src/components/ModalDiff/hook.js
```diff
@@ -0,0 +1,104 @@
...
+  const [baseLoading, setBaseLoading] = useState(false);
+  const [compareLoading, setCompareLoading] = useState(false);
+  const [compareProposal, setCompareProposal] = useState({});
+  const [baseProposal, setBaseProposal] = useState({});
+  const [error, setError] = useState();
+  const onFetchProposalDetailsWithoutState = useAction(
+    act.onFetchProposalDetailsWithoutState
+  );
+  const fetchProposalVersions = useCallback(
+    async (onFetchProposalDetailsWithoutState, token, version) => {
```
victorgcramos, line 19: I think you don't to pass the `onFetchProposalDetailsWithoutState` function to the `fetchProposalVersions` func? I think you should pass the function to the `useCallback` deps instead of passing it every time you need to fetch a proposal version.

src/components/ModalDiff/hook.js
```diff
@@ -0,0 +1,104 @@
...
+      .catch((e) => {
+        setError(e);
+        setCompareLoading(false);
+      });
+  }, [
+    token,
+    compareVersion,
+    fetchProposalVersions,
+    onFetchProposalDetailsWithoutState
+  ]);
```
victorgcramos, line 93: The purpose of using the fetch machine is to avoid this multiple side effects handling. You have 4 possible states here, which could be used on the machine to improve the side-effects control. But, it's just my opinion. The way you did works fine.

src/components/ModalDiff/DiffProposal.jsx
```diff
@@ -0,0 +1,168 @@
...
+        onChange={changedVersion}
+        base={baseVersion}
+        compare={compareVersion}
+        baseLoading={baseLoading}
+        compareLoading={compareLoading}
+      />
+      {!!baseProposal.details &&
+      !!compareProposal.details &&
+      !baseLoading &&
+      !compareLoading ? (
```
victorgcramos, line 53: use a new variable that is more declarative when you have lots of conditions to handle. Maybe `isDiffAvailable` or something like this.

src/components/ModalDiff/DiffProposal.jsx
```diff
@@ -0,0 +1,168 @@
...
+                  username={compareProposal.details.username}
+                  url={`/user/${compareProposal.details.userid}`}
+                />
+                <JoinTitle separatorSymbol={"-"}>
+                  {baseVersion > 0 && (
+                    <JoinTitle>
+                      {baseProposal.details.timestamp !==
+                        baseProposal.details.publishedat &&
+                        baseProposal.details.timestamp !==
+                          baseProposal.details.abandonedat && (
```
victorgcramos, line 76: same here for the long conditions. Move it to a separate variable so we can have a cleaner code.

src/components/ModalDiff/DiffProposal.jsx
```diff
@@ -0,0 +1,168 @@
...
+                          timestamp={baseProposal.details.abandonedat}
+                        />
+                      )}
+                    </JoinTitle>
+                  )}
+                  <JoinTitle>
+                    {compareProposal.details.timestamp !==
+                      compareProposal.details.publishedat &&
+                      compareProposal.details.timestamp !==
+                        compareProposal.details.abandonedat && (
```
victorgcramos, line 96: long conditions

src/components/ModalDiff/DiffProposal.jsx
```diff
@@ -0,0 +1,168 @@
...
+                    {compareProposal.details.abandonedat && (
+                      <Event
+                        className="margin-left-s margin-right-s"
+                        event={"abandoned"}
+                        timestamp={compareProposal.details.abandonedat}
+                      />
+                    )}
+                  </JoinTitle>
+                </JoinTitle>
+              </JoinTitle>
```
victorgcramos, line 112: these two children do the same stuff. The only difference is that one is for `baseProposal` and other is for `compareProposal`. Since it's always (and should always be) the same code, please move it to another separate component, so we can improve scalability and declarativeness of our code.

src/components/ModalDiff/CompareVersionSelector.jsx
```diff
@@ -0,0 +1,88 @@
...
+  const isDarkTheme = themeName === DEFAULT_DARK_THEME_NAME;
+  const darkIconColor = getThemeProperty(theme, "text-color");
+  const iconColor = isDarkTheme ? darkIconColor : undefined;
+  const versionsOptions = useMemo(() => {
+    const versions = [];
+    for (let index = latest; index >= 0; index--) {
+      versions.push(index);
+    }
+    return versions;
+  }, [latest]);
```
victorgcramos, line 31: let's use this https://lodash.com/docs/4.17.15#rangeRight to be more declarative. Try to use lodash or lodash/fp methods as always as you can. Only use `for` loops when it requires a huge performance improvement, or when you are dealing with long and complex arrays.

victorgcramos (changes requested):
Hey @vibros68, thanks for the PR. This is working as expected. I have only one comment regarding the UX. I think the proposal title should be bigger, because it does not look like a title and looks like a simple bold text. Also, there should be something indicating that these two selectors are about comparing one **against** the other.

Screen Shot 2021-08-19 at 2 46 06 PM

Left come inline comments regarding the code readability.

lukebp:
> I think the proposal title should be bigger, because it does not look like a title and looks like a simple bold text.

This is something that I ok'd, but I can see your point.  To keep things simple, this formatting should match the list item formatting.

> Also, there should be something indicating that these two selectors are about comparing one against the other.

@victorgcramos is this a change that was introduced by this PR?  Seems like it's out of scope.

victorgcramos:
> @victorgcramos is this a change that was introduced by this PR? Seems like it's out of scope.

It's just a simple indicator, like an icon between these two selectors. Having an empty space between them bothers me a little. Should be an easy task. See:
130118453-7c6e10fa-57bf-47da-8b7b-e628a99b57b2
(did this using the markup tool on Mac. Of course it can be improved, but it's just to illustrate my point)

What do you think? The images on the PR description included this icon, and it looks good IMO.

src/components/ModalDiff/hook.js
victorgcramos: Also, the `../../redux` can be replaced by `src/redux`

lukebp:
@victorgcramos oh I misread your comment.  I thought you were refering to changing how the red/green diff is shown.  Yeah, the indicator is already there.  It's just requires decred/pi-ui#357 to be merged.

src/components/ModalDiff/DiffProposal.jsx
victorgcramos: I mean, use a new generic component for both, instead of `crtl+c` and `ctrl+v`. Always avoid doing it. This is a bad practice in terms of maintainability.

vibros68:
>I think the proposal title should be bigger, because it does not look like a title and looks like a simple bold text. 
> 

Hmm. I changed the proposal title to be smaller because now there the modal title. If there are 2 titles in a frame, it looks a little 
weird. So I think I should change it to be bigger than normal size but smaller than the main title. How do you think @victorgcramos

src/components/ModalDiff/hooks.js
```diff
@@ -0,0 +1,102 @@
...
+      })
+      .catch((e) => {
+        setError(e);
+        setBaseLoading(false);
+      });
+  }, [
+    token,
+    baseVersion,
+    fetchProposalVersions,
+    onFetchProposalDetailsWithoutState
```
victorgcramos, line 70: You forgot to remove it.

src/components/ModalDiff/hooks.js
```diff
@@ -0,0 +1,102 @@
...
+      })
+      .catch((e) => {
+        setError(e);
+        setCompareLoading(false);
+      });
+  }, [
+    token,
+    compareVersion,
+    fetchProposalVersions,
+    onFetchProposalDetailsWithoutState
```
victorgcramos, line 89: same.

victorgcramos (approved):
LGTM! You forgot to remove some `useEffect` deps.

vibros68:
@amass01 Can you do the further review?

tiagoalvesdulce (changes requested):
I'm seeing some weird behavior when I play a little bit with the versions selected. In a proposal with 2 versions, if I select to compare the "base version" (empty) against the version 1, I can't go back and compare it against the version 2 if I don't close the modal.

Here is a video with steps to repro:

https://user-images.githubusercontent.com/13955303/130456259-994ccfbe-a9a6-4161-8c5e-244fd7cdfc06.mp4

Please ignore the comparison icon missing, my `pi-ui` version is outdated.

src/components/ModalDiff/CompareVersionSelector.jsx
```diff
@@ -0,0 +1,78 @@
...
+  className,
+  latest,
+  base,
+  compare
+}) => {
+  const { theme, themeName } = useTheme();
+  const isDarkTheme = themeName === DEFAULT_DARK_THEME_NAME;
+  const darkIconColor = getThemeProperty(theme, "text-color");
+  const iconColor = isDarkTheme ? darkIconColor : undefined;
+  const versionsOptions = useMemo(() => rangeRight(latest), [latest]);
```
tiagoalvesdulce, line 26: I think this is causing the issue in my previous comment. When you select a version other than the latest version, the real latest is being cropped out the `versionsOptions`. I suspect that when you are selecting a `version < than the latest` in the version comparison, the `latest` prop is being updated. For example: you have a proposal with 3 versions, I select to compare the version 1 against version 3, all works. Now, if I select to compare version 1 against version 2, latest is being updated to 2. This way 3 is left out of the `versionsOptions` array and I can't select it.

vibros68:
> I'm seeing some weird behavior when I play a little bit with the versions selected. In a proposal with 2 versions, if I select to compare the "base version" (empty) against the version 1, I can't go back and compare it against the version 2 if I don't close the modal.

Hi @tiagoalvesdulce . Your problem because the left selector is for base version and the right is for compare version. The compare version is always older than the base version. So if

> I'm seeing some weird behavior when I play a little bit with the versions selected. In a proposal with 2 versions, if I select to compare the "base version" (empty) against the version 1, I can't go back and compare it against the version 2 if I don't close the modal.
> 
Hi @tiagoalvesdulce . Thank you for your review. The bug is appeared because I used wrong way with the `lodash/rangeRight` function. I fixed it and just pushed the code

tiagoalvesdulce (approved):
LGTM. Good job!

lukebp (approved):
tACK

PR #2527, opened 2021-08-10
fix: add files to new proposal and edit proposal

### LINKED ISSUE #2529: Inline image attachments bug
Inline image attachments are not being submitted as part of the file payload.  The inline image renders correctly in the markdown preview, but then is not included in the request payload.

**Attribution**

This bug was introduced in #2469.

**Prevention**

This bug highlights the need for more comprehensive unit tests.  It was a tricky one because it was only tangentially related to the changes in 2469, so was unlikely to be caught in manual testing of the PR.  Unit tests would have caught it though.

tiagoalvesdulce:
Screen Shot 2021-08-10 at 15 20 33

Closes #2529

### DIFF
--- a/src/actions/api.js
+++ b/src/actions/api.js
@@ -869,7 +869,7 @@ export const onSubmitProposal = (
         markdown: description,
         linkby: rfpDeadline,
         linkto: rfpLink,
-        files
+        attachments: files
       })
     )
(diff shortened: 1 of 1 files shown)


victorgcramos (approved):
tACK

lukebp (approved):
tACK

PR #2542, opened 2021-08-17
fix: Render markdown tables correctly.

### LINKED ISSUE #2540: Markdown tables are broken
I'm pretty sure it rendered correctly in the [translations proposal](https://proposals.decred.org/record/af9942a) previously, but now it shows:

tables

I just pasted the code into gist and it renders as a table so I think the syntax is correct.

victorgcramos:
This diff Closes #2540 by adding `remark-gfm` plugin to markdown renderer. Also, adds unit tests for the markdown renderer.

### UI Changes Screenshot
**Before:**
Screen Shot 2021-08-18 at 9 13 23 PM

**After:**
Screen Shot 2021-08-18 at 9 13 01 PM


tiagoalvesdulce (approved):
This PR LGTM. Great job!

Only one thing...

With the `remark-gfm` plugin we also get `autolink literals`, `strikethrough ` and `tasklist`.

For example, with this markdown:

```
## Autolink literals

www.example.com, https://example.com, and contact@example.com.

## Strikethrough

~one~ or ~~two~~ tildes.

## Table

| a | b  |  c |  d  |
| - | :- | -: | :-: |

## Tasklist

* [ ] to do
* [x] done
```

We get the following result on test-proposals:

Screen Shot 2021-08-20 at 16 22 27

While on this PR we see:

Screen Shot 2021-08-20 at 16 24 07

I personally like the `strikethrough` and `task list` but I'm not a fan of the `autolink literals`. Can we choose the options we would like to apply to our markdown renderer?

What do you think @lukebp ?

victorgcramos:
Totally agree @tiagoalvesdulce! I think autolinks is okay too, because it's also replaced by our `LinkRenderer`. It's a matter of UX, but it brings us closer to the Github's markdown UX.

lukebp:
I don't have a strong opinion on this.  @tiagoalvesdulce can make the call.

lukebp:
Typically, we would want to match github's UX.  This case is slightly different though because of the security implications.  Highlighting the URLs can allow someone to potentially misdirect a user to a malicious site.  Example:
```
[https://docs.decred.org](https://someothersite.com)
```

This would appear to the user like they were navigating to dos.decred.org, correct?

victorgcramos:
For any external links, the modal is displayed for the user indicating the URL he/she is attempting to access. But yes, if the user doesn't pay attentions to the modal, this might happen.

lukebp:
The modal certainly helps, but the ability to misrepresent links in the markdown is something that should still be prevented imo.  I agree with @tiagoalvesdulce on this.  We should not use the autolink literals.

lukebp:
Now that I think about it, what I referenced before is not actually the issue at hand.

```
[https://docs.decred.org](https://someothersite.com)
```

That will be rendered as a markdown link regardless of whether we choose to use autolink literals.  The autolink literals only apply to text that is not already formatted as a markdown link.  Example:
```
https://somesite.com
```

I would probably lean towards matching the github UX, but I don't feel strongly either way so I'll leave this decision up to @tiagoalvesdulce.

victorgcramos:
I agree. Will fix this. But if someone wants to misrepresent links, he could just `[https://url1.com](https://malicious.com)`, and this will render [https://url1.com](https://malicious.com). I don't think we can prevent it from happening. But this will display the modal telling the real URL, see:
Screen Shot 2021-08-23 at 5 46 59 PM

tiagoalvesdulce:
I talked with victor on matrix and the [API](https://github.com/remarkjs/remark-gfm#api) doesn't allow us to turn off only autolinks. This PR LGTM.

PR #2549, opened 2021-08-26
comments: Add proposal author updates.

### LINKED ISSUE #2547: comments: Add proposal author updates.
**Spec**

Proposal authors need to be able to give periodic updates on the status of their proposal. These updates will be given as comments. The Politeia users will have the ability to interact with the author and ask further questions.

Currently, once a proposal vote has finished the proposal is locked and no additional comments can be made. The new behavior will be the following.

  - Once a proposal vote has finished, all existing comment threads are locked.
    
  - When a proposal author wants to give an update they will start a new comment thread. The author is the only user that will have the ability to start a new comment thread once the voting period has finished. Each update will have an author provided title.
    
  - Anyone can reply to any comments in the thread and can cast upvotes/downvotes for any comments in the thread.
    
  - The comment thread will remain open until either the author starts a new update thread or an admin marks the proposal as closed/completed.

**implementation**

  - Each update will be shown as its own comment section. The updates will be displayed from newest to oldest.
    
  - Each update will have a title.

### LINKED ISSUE #2570: useScrollTo: cleanup setTimeout on unmount.
The following `useScrollTo` hook is very useful and handy, sometimes when navigating away from
the proposal details page where this hook is used an error thrown to the console: 
`document.getElementById(element) is undefined` 

This because the component is not in the DOM when the setTimeout callback is called so we will 
need to cleanup the setTimeout on component unmount by returning a cleanup function from 
the `useLayoutEffect` hook.

```
import { useLayoutEffect } from "react";

const scrollToElement = (element) =>
  setTimeout(() => {
    document.getElementById(element).scrollIntoView();
  }, 100);

function useScrollTo(element, shouldScroll) {
  useLayoutEffect(() => {
    shouldScroll && scrollToElement(element);
  }, [element, shouldScroll]);
}

export default useScrollTo;
```

amass01:
**Spec**

Proposal authors need to be able to give periodic updates on the status
of their proposal. These updates will be given as comments. The Politeia
users will have the ability to interact with the author and ask further
questions.

Before this commit, once a proposal vote has finished the proposal is 
locked and no additional comments can be made. This commit changes
that to the following:

  - Once a proposal vote has finished, all existing comment threads are
    locked.
    
  - When a proposal author wants to give an update they will start a new
    comment thread. The author is the only user that will have the
    ability to start a new comment thread once the voting period has
    finished. Each update will have an author provided title.
    
  - Anyone can reply to any comments in the thread and can cast
    upvotes/downvotes for any comments in the thread.
    
  - The comment thread will remain open until either the author starts a
    new update thread or an admin marks the proposal as
    closed/completed.

**implementation**

  - Each update is shown as its own comment section. The updates will be
    displayed from newest to oldest.
    
  - Each author update has a title - comment form updated & title
    displayed if the comment in an author update.

  - **This diff refactors/normalizes the comments redux state, by
    storing the comments as a map vs as an array before this. 
    This improves the performance and avoids extra expensive
    re-renders on the containers level where they need to process the
    comments array to split the comments to different sections in case
    the proposal has author updates which are separated threads from 
    the main comments thread.**
    
---

Closes #2547.
Closes #2570.
Depends on decred/politeia#1491.


src/containers/Proposal/Detail/Detail.jsx
```diff
@@ -84,104 +101,176 @@ const ProposalDetail = ({ Main, match }) => {
...
+                  }
+                />
+              )}
+          </LoggedInContent>
+          {commentsError && (
+            <Message kind="error" className="margin-top-m">
+              {commentsError.toString()}
+            </Message>
+          )}
+        </Card>
```
amass01, line 234: This used to be in `container/Comments/Comments.jsx` but as it could be displayed now multiple time for each
author update section this logic moved to `container/Proposal/Detail/Details.jsx` as it should be displayed
once in the proposal details page.

amass01 (commented):
just found a bug when viewing single thread using comment specific link while having multiple author update threads

amass01:
> just found a bug when viewing single thread using comment specific link while having multiple author update threads

Fixed.

Another minor bug I just found, if proposal author posts a new author update it goes to the bottom while
it should show up as the first section.

amass01:
> Another minor bug I just found, if proposal author posts a new author update it goes to the bottom while
> it should show up as the first section.

Fixed.

lukebp (changes requested):
1. Comment plugin errors need human readable messages.  Note, this error was caused by me not having my comments `allowextradata` plugin setting set to true, so it was not caused by something in this PR.

image

2. The update title should be the title for the comments section, not in the comment itself.  Example: the `Comments` text from `Comments (1)` in the image below should be the update title.

image


3.  If I was a proposal author that was using this feature for the first time, I would not really know how to use it.  The rules of proposal updates are not clear from the existing UX.  Here are some things I think a first time user would struggle with.

- How am I suppose to know that the comment markdown editor is suppose to be used for proposal updates?  It looks like I'm leaving a normal comment.  I probably wouldn't know that this is for author updates only and would likely assume that this is availble for every user.

- How would I know that creating a new update locks the old thread?  I would probably assume that I can use this like normal comments, which could lead to me creating multiple updates instead of replying to comments in the original thread.

We should talk through the UX for this in the pi-dev channel.

Everything else appears to be working nicely.  We just need to work through some of these UX details.

amass01:
>     1. Comment plugin errors need human readable messages.  Note, this error was caused by me not having my comments `allowextradata` plugin setting set to true, so it was not caused by something in this PR.
> 

Addressed.

amass01:
> The update title should be the title for the comments section, not in the comment itself. Example: the Comments text from Comments (1) in the image below should be the update title.

Addressed.

lukebp:
https://user-images.githubusercontent.com/9039877/132263200-2adb300c-3ef1-4339-8947-01bfad1b8b57.mp4

lukebp (changes requested):
1. Still seeing missing error codes. Also looks like the padding could be improved on the form title.

image

2. Submit the update, but exit out of the confirmation modal.  The update has been cleared out when it shouldn't be.

3. Info icon doesn't display description on hover when the browser width has been adjusted down.  It requires a click.  Is this expected?

image

4. Should we put the info icon on the confirmation modal?  I realize that this was part of the design spec on the login modal, but I think we should probably remove that as well.  We use the info icon throughout politeia to provide the user with additional information of hover or on click for mobile.  Having some info icons display additional descriptions and some that don't do anything is a confusing UX imo.  I think we should only use the info icon to display additional descriptions.

image

lukebp (changes requested):
UX looks good.  Needs test coverage.

victorgcramos (changes requested):
Hey @amass01, thanks for the PR! Really good feature! I have one minor observation regarding the comments votes:

As mentioned on Matrix, I guess the comments shouldn't be sorted after every vote action. This may cause the user to lose his/her comment in the middle of other comments. IMO, we should only sort the list when the page gets reloaded.

Also, I noticed some screen flickering when voting on a comment from a long list (150 comments, generated by the `pictl seedproposals` command). See:

2549-proposal-author-comments-ordering

victorgcramos (changes requested):
Hey @amass01 thanks for addressing my comments. I noticed two issues regarding the redux state refactor:

1. The comment votes are no longer getting optimistically updated. This was requested by #2517, and it was the solution I figured out to avoid these responses lags. Simulated on a 1kbit/s throttle:
    2549-proposal-author-comments-optimistic-updates

2. Comments are getting cleared after the logout action.
    2549-proposal-author-comments-logout-magic

Also, I have one question regarding the `useComments` hook inside the proposal detail container. From my PoV, comments should be handled regardless the proposals, because we can use comments for invoices, DCCs, pireddit posts and other incoming stuff. The implemented hook also contains information regarding Invoices and dccs, and it doesn't make much sense to move them inside the proposals container.

TLDR: comments is a plugin that extends records functionality, so it should be linked to records, not specifically linked to proposals, IMO.

What do you think?

lukebp:
> TLDR: comments is a plugin that extends records functionality, so it should be linked to records, not specifically linked to proposals, IMO.

yes, that's the correct way to think about it.  Generic records.  Generic comments.  Only the pi plugin knows what a proposal is.

amass01:
@victorgcramos optimistic votes logic in back in.

Regarding your second point theoretically I totally agree with you, but as you can see I call `useComments` not directly 
from from proposal detail container instead I call it from `useProposal` and uses only the list of the author update ids.
I need this information to display the multiple comments sections in case proposal has author updates, and for that we
need to call the `/comments` route which is being done in `useComments`.

I thought to add a new hook to fetch the author update ids separately but this requires moving all the comments fetching
logic to that new hook and I'm not sure if that's improve it, because you are still calling the `/comments` route behind the scene.

amass01:
Open TODOs:

- Comments are getting cleared after the logout action.

src/containers/Comments/Comments.jsx
```diff
@@ -5,38 +5,27 @@ import React, {
...
-import useModalContext from "src/hooks/utils/useModalContext";
-import WhatAreYourThoughts from "src/components/WhatAreYourThoughts";
+import { PROPOSAL_MAIN_THREAD_KEY } from "src/constants";
 import { commentsReducer, initialState, actions } from "./commentsReducer";
 import { getQueryStringValue } from "src/lib/queryString";
 import useLocalStorage from "src/hooks/utils/useLocalStorage";
 import { debounce } from "lodash";
-
-const COMMENTS_LOGIN_MODAL_ID = "commentsLoginModal";
+import { useComments } from "../Proposal/Detail/hooks";
```
victorgcramos, line 28: The hooks shouldn't have any proposal dependencies.

victorgcramos (commented):
Hmmm, can't you move the hook back to `src/containers/Comments/hooks.js` instead of `src/containers/Proposals/Detail/hooks.js`? Because if we needed to use the comments container anywhere else, you could just pass a prop to the `<Comments/>` component?

Maybe something like:

```javascript
const ProposalDetails = (props) => {
  // ...
  return (
    // ...
    <Comments
      token={recordToken}
      state={recordState}
      extraDataLabel="Author Updates"
      extraDataHint="proposalupdate"
    />
  );
};

// src/containers/Comments/Comments.jsx
import { useComments } from "./hooks";
const Comments = ({
  token,
  state,
  extraDataHint,
  extraDataLabel,
  threadParentId = 0,
  ...props
}) => {
  // ...
  const { comments, extraDataComments, ...commentsProps } = useComments(
    recordToken,
    ...otherArgs
  );
  // ...
  return (
    <>
      {/* render a section for extra data comments with the `extraDataLabel` title */}
      {extraDataComments &&
        extraDataComments.filter(
          ({ extradatahint }) => extradatahint === extraDataHint
        )}
      {/* render a section for unlabeled comments */}
      {comments}
    </>
  );
};

// src/containers/Comments/hooks.js

function useComments(token, ...otherArgs) {
  // just a simple indicator so you can select unlabeled comments
  const comments = useSelector(sel.comments(token));
  // select comments labeled as an extra data comment
  const extraDataComments = useSelector(sel.extraDataComments(token));
  // ...
  return { comments, extraDataComments, ...rest };
}
```
Left an inline comment regarding the dependency between those two containers I'm talking about.

src/containers/Comments/Comments.jsx
amass01: I can move useCommemts to our global hooks dir - src/hooks, as it's used in multiple places already, I had this in mind at some point.

amass01: > Hmmm, can't you move the hook back to `src/containers/Comments/hooks.js` instead of `src/containers/Proposals/Detail/hooks.js`? Because if we needed to use the comments container anywhere else, you could just pass a prop to the `<Comments/>` component?
> 
> Maybe something like:
> 
> ```js
> // src/containers/Proposal/Details.jsx
> const ProposalDetails = (props) => {
>   // ...
>   return (
>     // ...
>     <Comments
>       token={recordToken}
>       state={recordState}
>       extraDataLabel="Author Updates"
>       extraDataHint="proposalupdate"
>     />
>   );
> };
> 
> // src/containers/Comments/Comments.jsx
> import { useComments } from "./hooks";
> const Comments = ({
>   token,
>   state,
>   extraDataHint,
>   extraDataLabel,
>   threadParentId = 0,
>   ...props
> }) => {
>   // ...
>   const { comments, extraDataComments, ...commentsProps } = useComments(
>     recordToken,
>     ...otherArgs
>   );
>   // ...
>   return (
>     <>
>       {/* render a section for extra data comments with the `extraDataLabel` title */}
>       {extraDataComments &&
>         extraDataComments.filter(
>           ({ extradatahint }) => extradatahint === extraDataHint
>         )}
>       {/* render a section for unlabeled comments */}
>       {comments}
>     </>
>   );
> };
> 
> // src/containers/Comments/hooks.js
> 
> function useComments(token, ...otherArgs) {
>   // just a simple indicator so you can select unlabeled comments
>   const comments = useSelector(sel.comments(token));
>   // select comments labeled as an extra data comment
>   const extraDataComments = useSelector(sel.extraDataComments(token));
>   // ...
>   return { comments, extraDataComments, ...rest };
> }
> ```
> 
> I still don't agree with (or at least I couldn't understand 😅) the reason you moved the whole comments hook, which controls the comments container to inside the proposals container when actually, the only thing **required** from proposals (and also other records) is the record token. If you want some additional config, you can pass it by prop or context, but the comments container should work regardless the additional config.
> 
> For example, if you want to implement invoice comments in the future, you should use the same `<Comments/>`component, and just pass the props that differentiates it from other record types. You can allow comments votes or not, you can highlight record author comments or not, you can render the extra data labeled comments on a different section or not. These behaviors should be configured by props (or context) passed by the parent component, and the other comment plugin actions (vote, censor...) have the same behavior, because it interacts with the comments plugin.
> 
> Left an inline comment regarding the dependency between those two containers I'm talking about.

I'm not sure I follow, replied and addressed the inline comment.

amass01:
>     * Comments are getting cleared after the logout action.

Same on master, opening a separate PR with a fix.

amass01:
> > ```
> > * Comments are getting cleared after the logout action.
> > ```
> 
> Same on master, opening a separate PR with a fix.

Fixed on #2582.

amass01:
e2e tests added ✅

victorgcramos (approved):
Hey @amass01 thanks for the PR. Code is looking clean and good. Thanks for addressing my comments. Left minor inline comment regarding mocking test data.

Also, I think you could add more e2e tests to cover more actions over the author updates, for example:
- If I submit an author update for the first time, will the author updates be listed correctly?
- Submitting a new update locks the previous thread?
- Can I submit author updates on not approved proposals?
- Can I upvote/downvote an author update?
- Should I be able to reply author updates?

You can open an issue to track these e2e tests if you think it's more appropriate, but I think they would be a good addition :D

teste2e/cypress/e2e/comments/authorUpdates.js
```diff
@@ -0,0 +1,90 @@
...
+              endblockheight: 0,
+              eligibletickets: 0,
+              quorumpercentage: 0,
+              passpercentage: 0,
+              results: [],
+              bestblock: 767301
+            };
+            res.send(res.body);
+          });
+        }).as("votesummaries");
```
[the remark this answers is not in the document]
amass01: nice suggestion will do

amass01:
> 
> Also, I think you could add more e2e tests to cover more actions over the author updates, for example:
> 
>     * If I submit an author update for the first time, will the author updates be listed correctly?
> 
>     * Submitting a new update locks the previous thread?
> 
>     * Can I submit author updates on not approved proposals?
> 
>     * Can I upvote/downvote an author update?
> 
>     * Should I be able to reply author updates?
> 
> 
> You can open an issue to track these e2e tests if you think it's more appropriate, but I think they would be a good addition :D

I already cover the following two:

 >   If I submit an author update for the first time, will the author updates be listed correctly?
 >  Should I be able to reply author updates?
 
 But I also cover more:
 
 - Normal user can't post normal comment on approved proposal and only reply on author update
 thread
 - Author update can post only new author updates on approved proposal.
 
 I think this is more than enough for the scope of this PR and of course we can add more
 in the future.

lukebp:
Is this ready for a final review?

amass01:
@lukebp yep

lukebp (approved):
tACK

Nice work @amass01.  This is looking real nice.

Needs final code approval from @tiagoalvesdulce.  @victorgcramos can you also give this a final review?

victorgcramos (approved):
LGTM! Good job on e2e tests cleanup!

src/containers/Comments/CommentsList/CommentsListWrapper.jsx
```diff
@@ -1,13 +1,11 @@
 import React, { useState, useEffect } from "react";
 import CommentsList from "./CommentsList";
 
-const getChildren = (comments, commentId, lastTimeAccessed, currentUserID) => {
-  return (
-    comments.filter((comment) => +comment.parentid === +commentId) || []
-  ).map((comment) =>
-    createComputedComment(comment, comments, lastTimeAccessed, currentUserID)
+const getChildren = (comments, commentId, lastTimeAccessed, currentUserID) =>
+  (comments.filter((comment) => +comment.parentid === +commentId) || []).map(
```
tiagoalvesdulce, line 5: this `|| []` shouldn't be necessary since `Object.values({}).flat() == []` in the case `commentsMap` is empty.

src/containers/Comments/Comment/CommentWrapper.jsx
```diff
@@ -177,7 +185,8 @@ const CommentWrapper = ({
     loadingLikes ||
     readOnly ||
     (userLoggedIn &&
-      (identityError || paywallMissing || currentUser.username === username));
+      (identityError || paywallMissing || currentUser.username === username)) ||
+    (areAuthorUpdatesAllowed && !isInLatestUpdateCommentTree);
```
tiagoalvesdulce, line 189: can we come up with a meaningful name for this expression since it's reused?

src/containers/Comments/Download/hooks.js
```diff
@@ -42,7 +43,11 @@ export function useDownloadCommentsTimestamps(recordToken) {
     () => sel.makeGetRecordComments(recordToken),
     [recordToken]
   );
-  const comments = useSelector(commentsSelector);
+  const allCommentsBySection = useSelector(commentsSelector);
+  const comments = useMemo(
```
tiagoalvesdulce, line 47: why do we need this `useMemo` here? `allCommentsBySection` is an object, objects are compared by reference and you are creating `allCommentsBySection` above. The code inside `useMemo` is still executed on every render.

src/containers/Comments/helpers.js
```diff
@@ -8,6 +8,41 @@ export const commentSortOptions = {
...
+ * @param {Int} leafId leaf node id.
+ * @param {Map} comments array of comments.
+ */
+export const isInCommentTree = (rootId, leafId, comments) => {
+  if (leafId === rootId) {
+    return true;
+  }
+
+  // Convert comment array to a map
+  const commentsMap = comments.reduce((map, comment) => {
```
tiagoalvesdulce, line 25: 👍

src/helpers.js
```diff
@@ -583,3 +584,28 @@ export function getAttachmentsFiles(files) {
...
+
+/**
+ * calculateAuthorUpdateTree accepts an array of comments and an author update
+ * id. It calculates the author update thread comment tree then returns
+ * the tree as an sub-array of the original array.
+ * @param {String} authorUpdateId
+ * @param {Array} comments
+ * @returns {Array} array of author update thread comments.
+ */
+export const calculateAuthorUpdateTree = (authorUpdateId, comments) => {
```
tiagoalvesdulce, line 596: 👍 . I would suggest creating a variable to `uniq([...authorUpdateTree, ...children])`. Also, you are doing this functionality twice:

```
let children = comments
    .filter(({ parentid }) => authorUpdateTree.includes(parentid))
    .map(({ commentid }) => commentid);
```

I would create a `getChildrenComments` that takes a `parents` array as a parameter.

src/hooks/api/useComments.js
```diff
@@ -0,0 +1,180 @@
...
+
+  const getCommentVotes = useCallback(
+    (commentID) => commentsVotes && commentsVotes[commentID],
+    [commentsVotes]
+  );
+
+  // If displaying a single thread while having multiple author updates with
+  // different section, find to which tree the sub-tree we are displaying
+  // belongs to display only the relevant section.
+  const singleThreadRootId = useMemo(() => {
```
tiagoalvesdulce, line 143: I also think this `useMemo` is being useless because of `allCommentsBySection `. I'm not sure because now you are using selectors and I don't know if the selector memoization is helping here. Please investigate.

tiagoalvesdulce (changes requested):
The PR is looking pretty good. Great job! I left a few comments.

src/hooks/api/useComments.js
amass01: @tiagoalvesdulce 

Hrmm actually I added some console logs and tried without the `useMemo` and found that
it's useful in this case and the code ran twice on first load with useMemo vs tens of times without,
and i think that you are right about the ref comparison but as the variable `allCommentsBySection` is
actually the result of `useSelector` which dosn't use plain objects and some kind of `useState` so the ref shouldn't change between re-renders. what do you think ?

src/containers/Comments/Comment/CommentWrapper.jsx
amass01: good point, went with:  `notInLatestAuthorUpdateThread` which is way more readable.

src/containers/Comments/Download/hooks.js
amass01: hrmmm actually this is similar to the other comment with `allCommentsBySection`, it's ref shouldn't
change between re-renders thus the useMemo should be useful also here.

src/helpers.js
amass01: Great suggestions, thanks 💯

amass01: done

src/containers/Comments/Download/hooks.js
tiagoalvesdulce: hmm, I commented this pretty much because of https://react-hooks-cheatsheet.com/usememo and experience using `useMemo` with objects in the deps list before. I will take a closer look. Thanks!

tiagoalvesdulce (approved):
tACK

PR #2558, opened 2021-08-31
fix: Consistent error message for invalid identity

### LINKED ISSUE #2516: Improve identity error message.
The invalid identity error message that is displayed when attempting to authorize a vote is:

image

A more helpful error message is the one we display for proposal submissions since it tells the user how to fix it.

image

We should update the first error message to be more like the second error message.

thi4go:
closes #2516 

I decided to keep this simple and just match the error message for errors thrown in the `lib/pki` and for errors rendered with `components/IdentityErrorIndicator` components.

- `lib/pki` throws the identity error caused on runtime by methods that call the lib's functions.
- `components/IdentityErrorIndicator` is rendered when the components check beforehand if the identity is valid or not.

### DIFF
--- a/src/lib/pki.js
+++ b/src/lib/pki.js
@@ -5,8 +5,10 @@ import util from "tweetnacl-util";
 import get from "lodash/fp/get";
 
 export const STORAGE_PREFIX = "ed255191~";
-export const IDENTITY_ERROR =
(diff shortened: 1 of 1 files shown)


tiagoalvesdulce (approved):
LGTM

PR #2560, opened 2021-09-04
Edit proposal: validate initial start date.

### LINKED ISSUE #2546: Bad error message on proposal edit
When editing a RFP proposal, if I keep the start date older than the actual date, the GUI lets me submit the proposal but shows a bad error message.

Error message:
Screen Shot 2021-08-23 at 10 28 17

Error on www:
Screen Shot 2021-08-23 at 10 28 29

amass01:
This commit fixes two bugs in the edit proposal flow:

 - Editing a proposal with a start date which is not valid anymore
   resulted in an backend error and the message which was displayed to
   the user included only the BE error code, this commit maps the
   error codes to a human readable text messages which will be displayed
   to the user.

- GUI didn't validate start date on proposal edit, this commit ensures
  that the existing start date is still valid and smaller than minimum
  start date.  If the start date is invalid the form's submit is
  disabled and an error message is displayed next to the field, the
  error message text includes the formatted minimum start date.
  
This also adds e2e tests to ensure invalid field error codes are 
mapped to a human readable error message.

--- 

Closes #2546.


lukebp (approved):
ACK

tiagoalvesdulce (approved):
LGTM

PR #2566, opened 2021-09-07
fix: e2e tests.

### LINKED ISSUE #2564: e2e: fix failing tests.

amass01:
This diff fixes the following 3 e2e tests:

1. Proposal details test expected long token instead of short token
in the `/records` reply map.
2. Admin censored proposals list displays short token.
3. Comment vote test:
`should reset votes count on error` 
is now:  
`shouldn't change votes count on error`

---

Closes #2564.


tiagoalvesdulce (approved):
LGTM

PR #2568, opened 2021-09-08
fix: Avoid registration error after unpaid user logout.

### LINKED ISSUE #2562: User logout error when registration fee is not paid
This is only noticeable on `development` environment, but it's actually a bug caused by hitting the `/users/payment/registration` route after the logout action, and it ends up showing an "You must be logged in to perform this action" error.

### Steps to reproduce
1. Create a new user, verify it and  👉 **don't pay the registration fee** 👈 .
2. Logout
3. Wait until the error show up

unregistered-users-logout-error

victorgcramos:
Closes #2562.

This fixes an error triggered after an unpaid user logout. It happened
due to the payment polling call not being canceled on logout action.

It also adds E2E tests for user registration, login and logout.

### UI Changes Screenshot

**Before**
before

**After**
unregistered-users-logout-error-fix

### DIFF
--- a/src/actions/api.js
+++ b/src/actions/api.js
@@ -90,6 +90,10 @@ export const setPollingPointer = (paymentpolling) => {
 const POLL_INTERVAL = 10 * 1000;
 export const onPollUserPayment = () => (dispatch, getState) => {
   const userid = sel.currentUserID(getState());
+  if (!userid) {
+    clearPollingPointer();
+    return;
+  }
   return api
     .verifyUserPayment()
     .then((response) => response.haspaid)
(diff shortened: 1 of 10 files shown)


tiagoalvesdulce (approved):
ACK

Can you put the github keyword and PR description in different lines? Example:

```
Closes #issue_number

PR description....
```

lukebp:
@victorgcramos can you update your PR message to adhere to the 72 column limit so that it can be copy and pasted as the commit message?

https://gist.github.com/lukebp/5a034ba73646903669719b4724a1d945#commit-messages

PR #2571, opened 2021-09-11
fix: use full token in dcrdata charts link.

### LINKED ISSUE #2569: Use full length token for dcrdata link
The dcrdata voting charts link currently uses the shortened token, but it needs to use the full length token.

https://proposals.decred.org/record/58d9f46

Example:
https://dcrdata.decred.org/proposal/58d9f46 - this is broken
https://dcrdata.decred.org/proposal/58d9f4664777775d - this works

dcrdata button
image

amass01:
Closes #2569.


tiagoalvesdulce (approved):
ACK

PR #2608, opened 2021-09-25
fix: Comments timestamp download infinite loop

### LINKED ISSUE #2606: Timestamp downloads infinite loop
Attempting to download comment or vote timestamps results in infinite loops.

amass01:
This diff fixes the infinite loop when attempting to download comments
timestamps results.

The bug originates in the following `useMemo` usage discussion:
https://github.com/decred/politeiagui/pull/2549#discussion_r711424772

---

Closes #2606.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,4 @@ yalc.lock
 
 cypress/videos
 cypress/screenshots
+teste2e/cypress/downloads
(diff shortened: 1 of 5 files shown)


lukebp:
Are there tests we can add for this?

lukebp (approved):
tACK on firefox

It seems like the way to have prevent this bug is to have had e2e tests for all the downloads.  Is that accurate?

amass01:
@lukebp that's correct, I guess we didn't because it might need some changes in the Dropdown component. I can give it a try.

PR #2610, opened 2021-09-26
fix: Type error on comment bundle download

### LINKED ISSUE #2607: Download bundle TypeError bug
Originally reported by @vibros68 in #2597.

Go to the Under Review page. F5 to refesh the cache data. Click on a proposal. After redirect to the proposal, click on Available Downloads as fast as you can. This will lead a page error: `TypeError: Cannot convert undefined or null to object`.

thi4go:
This diff adds a check to see if comments are properly loaded on the
redux store before manipulating it as an object, which would result 
in the type error.

closes #2607


tiagoalvesdulce (approved):
ACK

PR #2623, opened 2021-09-30
Improve identity error message.

### LINKED ISSUE #2617: Improve identity error message.
This is a contiuation of #2516, which was closed prior to addressing the underlying issue.

There are two problems with the current identity error messages:

1. We have more then one of them.
2. Some of them don't link to the user's account page.  This is a bad UX.  If we tell the user they need to update their identity, we should provide them with the link to do so.

image
image
image

There should be a single identity error message that reads:

`Your identity is missing or is invalid.  You cannot submit any data until you update your identity on your account page.`

Where the `account` is a link to the user's account page.

amass01:
This adds a new react error boundary to handle global identity error.
The identity error can be thrown in various modals and error boundaries
is react's way to handle such errors.

**Implementation:**

- Add a new `IdentityErrorBoundary` to catch global identity error.
- Add a new `useCarsh` hook in order to crash inside the component tree
  when an identity error is thrown from an async action.
- Add a new utils function `isIdentityError` and use it cross the app.

--- 
[React error boundaries](https://reactjs.org/docs/error-boundaries.html)
Closes #2617.

### DIFF
--- a/src/components/ErrorBoundary.jsx
+++ b/src/components/ErrorBoundary.jsx
@@ -17,14 +17,9 @@ class ErrorBoundary extends Component {
   }
 
   static getDerivedStateFromError(error) {
-    console.log({ error });
     return {
-      error
+      error: error
     };
-  }
-
-  componentDidCatch(error, info) {
-    console.log({ error, info });
   }
 
   render() {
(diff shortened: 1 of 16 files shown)


lukebp (approved):
tACK of firefox

I really don't like the way react handles these global errors.  It trains the developer to just hit `esc` on errors in order to bybass the error overlay.  There's not much we can do about it though.  The react devs feel strongly about it and this is not something that is going to change.

https://stackoverflow.com/a/47400249

lukebp:
Updated error message.

image

PR #2628, opened 2021-10-08
ux: Add min loading time to list views.

### LINKED ISSUE #2586: Smooth out initial load UX
When there are not proposals in the list view, the loading indicator is cut off mid cycle.  I'm using the word "cycle" to describe the graphics completing a loading loop.  This creates a jarring effect that is not a great UX.

https://user-images.githubusercontent.com/9039877/133443369-e6144e7c-3083-4683-8f8e-a69220237804.mp4

Compare this the UX below where the loading indicator is given one full cycle before the list view is populated.

https://user-images.githubusercontent.com/9039877/133443592-272e2086-29b8-42e8-ad1f-976c9bf5083d.mp4

The way to improve this UX is to smooth it out by displaying the loading indicator for at least one full cycle before showing the results, regardless of the number of proposals that need to be displayed.

vibros68:
Closes #2586.

This diff adds a minimum timeout of 1.25s when loading the proposal
lists, if the list does not contain any proposals. This creates a
smoother UX since the loading animation is not abruptly cut off.

**Old UX**

https://user-images.githubusercontent.com/9039877/133443369-e6144e7c-3083-4683-8f8e-a69220237804.mp4

**New UX**

https://user-images.githubusercontent.com/9039877/138359346-8a2f66ca-675f-4114-a7d0-6d8840b41651.mp4

### DIFF
--- a/src/components/LazyList.jsx
+++ b/src/components/LazyList.jsx
@@ -20,23 +20,23 @@ const LazyList = ({
   emptyListComponent,
   renderItem,
   pageStart,
-  minimumInitialLoading
+  minInitialLoading
 }) => {
   const isEmpty = !items.length;
   const [initialLoading, setInitialLoading] = useState(true);
   useEffect(() => {
     let timer;
-    if (minimumInitialLoading > 0 && initialLoading) {
+    if (minInitialLoading > 0 && initialLoading) {
       timer = setTimeout(() => {
         setInitialLoading(false);
-      }, minimumInitialLoading);
+      }, minInitialLoading);
     }
     return () => {
       // clearTimeout when the component is unmounted.
       clearTimeout(timer);
     };
-  }, [minimumInitialLoading, initialLoading]);
-  const placeholdersLoading = isLoading || initialLoading;
+  }, [minInitialLoading, initialLoading]);
+  const arePlaceholdersLoading = isLoading || initialLoading;
   return (
     <div data-testid="lazy-list">
       <InfiniteScroll
@@ -46,7 +46,7 @@ const LazyList = ({
         hasMore={hasMore}>
         {items.map(renderItem)}
       </InfiniteScroll>
-      {placeholdersLoading ? (
+      {arePlaceholdersLoading ? (
         <div data-testid="loading-placeholders">{loadingPlaceholder}</div>
       ) : isEmpty ? (
         emptyListComponent
@@ -64,7 +64,7 @@ LazyList.propTypes = {
   loadingPlaceholder: PropTypes.node,
   hasMore: PropTypes.bool,
   pageStart: PropTypes.number,
-  minimumInitialLoading: PropTypes.number
+  minInitialLoading: PropTypes.number
 };
 
 LazyList.defaultProps = {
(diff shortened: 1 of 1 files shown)


vibros68:
@lukebp read for review

src/hooks/api/useProposalsBatch.js
```diff
@@ -174,7 +175,15 @@ export default function useProposalsBatch({
           );
           setStatusIndex(index);
           setRemainingTokens(tokens);
-          if (isEmpty(tokens)) return send(RESOLVE);
+          if (isEmpty(tokens)) {
```
lukebp, line 178: If I was reading this code I wouldn't know why there is an intial wait here.  It's not obvious.  Add some documentation explaining why we do this.

lukebp (changes requested):
tACK on firefox.  This UX is much smoother.

https://user-images.githubusercontent.com/9039877/136573298-bf3101e2-19c5-4612-a03a-cb3f2cd4717f.mp4

You need some documentation though.

src/hooks/api/useProposalsBatch.js
```diff
@@ -174,7 +175,17 @@ export default function useProposalsBatch({
...
+            // When the tokens is empty and the fetched proposals is empty too.
```
lukebp, line 179:
```suggestion
            // When the tokens are empty and the fetched proposals are empty too,
            // wait 1 second to avoid removing the loading indicator suddenly.
```

vibros68:
Thank you @victorgcramos . Your opinion is reasonable. I will change to that way

amass01 (commented):
Can we try this with 500ms ?

lukebp:
Post videos for 500ms and 1 second so that we can compare the difference.  I think 500ms will probably be too quick.

lukebp (changes requested):
This doesn't appear to be working anymore with the new implemenation.  Are you seeing the same thing?

https://user-images.githubusercontent.com/9039877/137918149-13027c13-e605-4879-b214-e7f268a19d5f.mp4

src/components/LazyList.jsx
```diff
@@ -19,9 +19,18 @@ const LazyList = ({
...
+  minimumInitialLoading = 1250
+  // 1250 allows the animation to complete one full cycle.
 }) => {
   const isEmpty = !items.length;
+  const [initialLoading, setInitialLoading] = useState(true);
+  if (minimumInitialLoading > 0 && initialLoading) {
+    setTimeout(() => {
+      setInitialLoading(false);
+    }, minimumInitialLoading);
+  }
```
victorgcramos, line 32: I think you should wrap this into a `useEffect` to avoid re-renders and improve performance.

```suggestion
  useEffect(() => {
    if (minimumInitialLoading > 0 && initialLoading) {
      setTimeout(() => {
        setInitialLoading(false);
      }, minimumInitialLoading);
    }
  }, [minimumInitialLoading, initialLoading]);
```

victorgcramos (changes requested):
Hey @vibros68 thanks for the PR 🚀 . Looking really nice. One thing I noticed is that for long lists, the timeout is delayed due to list re-render.

Regarding the timeout, if you take a look at [react-content-loader docs](https://github.com/danilowoz/react-content-loader#options), you'll see `speed={2}` on `src/components/Proposal/ProposalLoader`. So I think the ideal timeout is 2 seconds.

Also, you forgot to define the `PropType` for `minimumInitialLoading`, and I think you should move your default values to `PropType`'s defaults.

Check my inline comment for more info 😁

PR #2629, opened 2021-10-08
feat: multiple billing status changes.

### LINKED ISSUE #2627: Set Billing Status: Allow multiple billing status changes.
https://github.com/decred/politeia/pull/1528 adds a new pi API setting to limit the number of
possible billing status changes.

The gui should fetch that value and allow setting billing status while the number of billing status
changes not exceeding the new max setting.

amass01:
This diff consumes the new pi API setting `billingstatuschangesmax`,
which should be used as the maximum allowed billing status changes.

After this commit all approved proposals with less billing status
changes than the maximum will have the `Set Biiling Status` 
button enabled for admins.

This also uses the batched `/billingstatuschanges`  route by using
`useMachineFetch` in `useBillingStatusChanges` to fetch the 
billing status changes of all approved proposals page by page.

---

Closes #2627.


amass01:
needs a e2e test.

src/reducers/models/proposals.js
```diff
@@ -369,24 +370,57 @@ const proposals = (state = DEFAULT_STATE, action) =>
...
-              action.payload.reason
-            ),
-          [act.RECEIVE_BILLING_STATUS_CHANGES]: () =>
-            set(
+                : action.payload.billingStatus ===
+                  PROPOSAL_BILLING_STATUS_COMPLETED
+                ? PROPOSAL_SUMMARY_STATUS_COMPLETED
+                : PROPOSAL_SUMMARY_STATUS_ACTIVE
+            );
+            state = update(
```
amass01, line 384:
```suggestion
            return update(
```

src/constants.js
```diff
@@ -100,6 +101,7 @@ export const PROPOSAL_COMMENT_UPVOTE = 1;
 export const PROPOSAL_COMMENT_DOWNVOTE = -1;
 
 export const PROPOSAL_PAGE_SIZE = 5;
+export const BILLING_STATUS_CHANGES_PAGE_SIZE = 5;
```
lukebp, line 104: Shouldn't this use the policy reply?

lukebp:
1 makes sense.  Not much we can do about that.

For 2, I think the loading indicator should be displayed until all required data has been fetched.  I'd rather we display the loading indcator a little longer then necessary for proposals that don't need the button then to add elements to the DOM after the list views have already been rendered.

amass01:
sweet, will add an indicator

amass01:
- [x] In detail view fetch billing status changes if current user is admin and the proposal is approved
and wait for all information before hiding the placeholder animation.

Admin:
https://user-images.githubusercontent.com/10324528/137198150-8d215856-531e-45c2-b082-180a20dee2e8.mov


Normal user:
https://user-images.githubusercontent.com/10324528/137198123-298dbaa9-e893-48a7-8df2-026250ac9aaa.mov

amass01:
@lukebp 

> 1. The billing status changes request is being sent on initial load of the Under Review tab. It's requesting the approved proposals, but it's being sent on the wrong tab.

Addressed.

src/hooks/api/useProposalsBatch.js
```diff
@@ -1,3 +1,4 @@
+import Promise from "promise";
```
tiagoalvesdulce, line 1: Do you have to import from `"promise"`?

tiagoalvesdulce (approved):
ACK

PR #2638, opened 2021-10-15
fix: Comments navigation issues.

### LINKED ISSUE #2615: Bad page scrolling on external links.
Found a couple issues with external links in comments.  The screen recordings refer to the [GoDCR](https://proposals.decred.org/record/f7d9fc8) proposal.

1.  Clicking on an external link in the comments section causes the screen to reposition itself at the top of the comments section.  It should not do this.

https://user-images.githubusercontent.com/9039877/134992092-f9b6e120-09a4-4cab-ad08-07940a1d74f1.mp4

2. If the `scrollToComments=true` param is set ([GoDCR example](https://proposals.decred.org/record/f7d9fc8?scrollToComments=true)) then you also get a weird re-render flickering.

https://user-images.githubusercontent.com/9039877/134992438-9a6741e1-62af-4cbc-8e8d-388fb56943c0.mp4

3. Add a external link to the comment markdown editor, preview it, click on it in the preview.  It closed the editor and moves the page position back to the top of the comments section.  The comment markdown editor should not be closed and the screen should not be re-positions.

### LINKED ISSUE #2620: Record naviation bug.
The `Go back` button does not work after you navigate to this URL: 
https://proposals.decred.org/record/f7d9fc8?scrollToComments=true&sort=New

https://user-images.githubusercontent.com/9039877/135167313-c266c187-effa-405f-a2b4-09d70f2985c9.mp4

### LINKED ISSUE #2633: [bug] Proposal detail page fires duplicate comment votes requests.
image

### LINKED ISSUE #2636: Lost preferences during comment navigation.
Reproduction steps:
1. Enable flat mode and sort comments by "New".
2. Click "See in context" on a reply.
3. Navigate back to the full comment section by clicking the "View All" button.
4. Flat mode is still enabled, but the "New" filter has been lost.  The expected behavior is for your filtering preference to stay the same.

This if a common user flow if you're trying to catch up on new comments.  Losing your filtering preference makes it hard to use.

### LINKED ISSUE #2637: Comment navigation duplicate requests.
The full comment section is being retrieved from the backend everytime you use the "see in context" feature and when you navigate back to the full thread using the "View All" button.  There's no need for this.  politeiagui already has all of the comments loaded.

This creates a bad UX since it adds in a delay when navigating between the full comment section and individual comment threads.  The transition should be seamless.

The only time the full comment section should be fetched from the backend is if someone is navigating directly to the comment thread link and the comment section has not been previously loaded.

Example: https://proposals.decred.org/record/f7d9fc8/comments/4

victorgcramos:
This PR fixes some comments navigation issues.

- Removes duplicate comments and votes requests
- Fixes comment area scroll
- Fixes GoBack link and history past locations
- Store comments filter on localstorage instead of url query
- Fixes comments re-rendering caused by modal actions side-effects

Closes #2637 
Closes #2633 
Closes #2636 
Closes #2620 
Closes #2615

### DIFF
--- a/teste2e/cypress/e2e/proposal/detail.js
+++ b/teste2e/cypress/e2e/proposal/detail.js
@@ -152,11 +152,8 @@ describe("Proposal details", () => {
       cy.visit(`/record/${shortToken}`);
       cy.wait("@details");
       cy.wait(1000);
-      // cy.findByTestId("wayt-login-button").click();
-      // cy.typeLoginModal(admin);
       cy.login(admin);
       cy.visit(`/record/${shortToken}`);
-      // cy.wait("@details");
       cy.findByTestId("proposal-body").should("exist");
       cy.get("#commentArea").should("exist");
     });


src/containers/Comments/Comment/Comment.jsx
```diff
@@ -88,7 +91,7 @@ const Comment = ({
           </Link>
           <DateTooltip timestamp={createdAt} placement="bottom">
             {({ timeAgo }) => (
-              <Link className={styles.timeAgo} to={permalink}>
+              <Link className={styles.timeAgo} to={permalink + sortQuery}>
```
amass01, line 94: string template

src/components/Router/RouterProvider.jsx
```diff
@@ -16,7 +15,7 @@ const RouterProvider = ({ location, children, ...rest }) => {
   const [pastLocations, setPastLocations] = useState([]);
 
   useEffect(() => {
-    if (!isEqual(pastLocations[0], location)) {
+    if (pastLocations[0]?.pathname !== location.pathname) {
```
amass01, line 18: i would suggest having `const lastLocationPathname = pastLocations[0]?.pathname` & 
`const locationPathname = location.pathname;` **outside** the useEffect, with that you 
won't have `[location, pastLocations]` is the deps array which are objects and you will
have `[lastLocationPathname, locationPathname]` which are strings.

amass01: sweet, but we should still define those consts outside the useEffect right ?

victorgcramos: yes correct, but we would still have `[location, pastLocations]` as deps

amass01: that's fine with me

victorgcramos:
Screen Shot 2021-10-18 at 10 59 19 PM

PR #2640, opened 2021-10-26
change format date view

### LINKED ISSUE #2634: International date format for start and end dates
Proposal start and end dates use the `M/D/Y` format that takes extra time to disambiguate, e.g. "am I looking at Aug 9 or Sep 8"?

pi-date-format

It might be common in some regions, but on the global scale it is the [least used](https://en.wikipedia.org/wiki/Date_format_by_country) format after `DMY` and `YMD`.

Let's change it to either `2021-09-08` or the more user-friendly `08 Sep 2021` used in other Pi timestamps.

vibros68:
Closes #2634 

This diff change the date view format of proposal from `9/8/2021` to
`08 Sep 2021`
Screenshot from 2021-10-26 11-08-13


tiagoalvesdulce (approved):
LGTM

PR #2624, opened 2021-10-01
test: E2E Mock API structure

victorgcramos:
This commit adds a plugin extensible test support structure for e2e tests. It 
allows the developer to have a more scalable structure for mock api test data,
create custom commands, and generate custom data for tests.

- adds a plugin extensible test structure
- improve the proposals list test to work offline using the new structure
- improve docs for e2e testing including a tutorial of how to extend the core
    plugin to create new custom support plugins

****

In order to make our codebase more scalable, we organized the support commands
and files according to the [politeia](https://github.com/decred/politeia) plugin
structure. Each plugin will have its own support files, which can extend the
core (records) support files and create custom commands. You can use the custom
commands to test the application.

### Core Package

Let's take a deeper look at the `teste2e/cypress/support/core`:

```bash
teste2e/cypress/support/core
├── api.js # mock api repliers
├── commands.js # custom plugin commands
├── generate.js # mock data generators
└── utils.js # support utils and helpers
```

#### `api.js`:

Contains all repliers for the plugin api calls. Each replier will accept one
object with two parameters: `testParams` and `requestParams`. Both describe
respectively test configuration parameters, which can be used to configure
the response to match test pre-conditions, and request data sent to the API.

Let's build some replier for the `/records/v1/inventory` call:

```javascript
import { Inventory } from "./generate";
import { stateToString } from "./utils";

export function inventoryReply({
  testParams: { amountByStatus = {}, pageLimit = 20 },
  requestParams: { state = 2, page }
}) {
  const inventory = new Inventory(amountByStatus, {
    page,
    pageLimit
  });
  return { [stateToString(state)]: inventory };
}
//...
```

> Tip: all repliers names should be followed by the `Reply` suffix, such as
> `inventoryReply`, `recordsReply`, `policyReply`...

As you can see, we are receiving `{ amountByStatus, pageLimit }` from the test
case, which describe the status and state pre-conditions to run the tests.
`{ state, page } ` is the request body object for the `/inventory` call (check
the [records api docs](https://github.com/decred/politeia/blob/7eb46182bbe319aebfa7e9a0cdad45e305dd41e0/politeiawww/api/records/v1/v1.go#L514)).

#### `commands.js`

Contains all custom cypress commands that can be used for testing the plugin
contents. See the [Cypress Docs](https://docs.cypress.io/api/cypress-api/custom-commands)
for more information about how to create custom commands.

The core package contains the custom commands that can be used by other
support plugins, and core support methods, which can be used to create custom
[...]

PR #2641, opened 2021-10-27
[PoC] Run parallel E2E tests

victorgcramos:
Running e2e tests takes a very long time. If we add more tests, more time.
It doesn't sound like a good option for us in terms of scalability...

Fortunately, [sorry-cypress](https://docs.sorry-cypress.dev/) provides us a
good solution for avoiding the necessity to have a 
[Cypress Dashboard](https://www.cypress.io/pricing/) account.

The parallel run can't be used without the implementation of #2624, because
the way our tests are currently designed, calling the politeiawww api with two
parallel tests would trigger lots of user identities errors, making our tests fail.

For now, the parallel tests are supposed to be used on local development env
only, since it's an experimental feature.

----

### (Experimental) Run tests in parallel

> Note: currently on experimental mode. Use it only for local development to
> speed up your tests run.

> You need to have Docker running in order to make this parallel testing

You can run parallel tests using the [sorry-cypress](https://docs.sorry-cypress.dev/)
alternative, instead of regular Cypress in order to avoid using Cypress
Dashboard, which requires an account and other KYC stuff. Fortunately, this
alternative wraps cypress into one docker image and opens it for usage.

**Parallel tests will only work with Mocked APIs tests suites**

1. Setup your sorry-cypress docker image: `yarn && yarn test:e2e:parallel:setup`
2. Run the application on development mode: `yarn start`
3. Open a new terminal windows (yes, plural) and run `yarn cy:run:parallel` on
   each.


victorgcramos:
We should be able to add that on the new architecture, once #2803 gets merged. We aren't using the user layer on the new app, so we should be able to run e2e in parallel without any issues.

PR #2658, opened 2021-11-16
fix: Go Back link navigation

### LINKED ISSUE #2655: Incorrect 'Go Back' button behavior.
When you select the 'Go Back' button on a proposal details page, it takes you back to the "Under Discussion" tab of the home page, regardless of where you came from.  Here are some examples:

- When viewing unvetted proposals, I want to go back to the unvetted proposal list, not the home page.
- When viewing proposals in a list such as "Approved", I want to go back to the list that I came from, not the home page.

Previously, the "Go Back" button behavior would match the same behavior as the browser "Back" button.  This caused UX issues when navigating comment threads, so it was updated in #2638 to the current behavior.  This new behavior causes is own set of UX issues.

I think the correct behavior for the "Go Back" button is to go back to the previous page when ignorning all URL params and record sub-routes.  For example, if a user follows this navigation sequence:

1. https://test-proposals.decred.org/?tab=approved
2. https://test-proposals.decred.org/record/a6a5084
3. https://test-proposals.decred.org/record/a6a5084/comments/1
4. Go Back

This should take them back to step 1, not step 2.

Another example:

1. https://test-proposals.decred.org/?tab=approved
2. https://test-proposals.decred.org/record/a6a5084
3. https://test-proposals.decred.org/record/a6a5084?scrollToComments=true
4. Go Back

Again, this should take them back to step 1, not step 2.

victorgcramos:
Closes #2655.

This diff fixes the Go Back link navigation by adding the `breakpoint` prop,
where users can inform the route match path that should be considered a
navigation rollback point.

Fortunately, due to privacy matters, the browsers don't allow the previous
history states access. Therefore, the method used consists in storing the
previous navigated pages on our platform in a stack, and then getting the
index of the first route that does not match the current breakpoint. This way
we can restore the window navigation state, instead of pushing a new one.

### DIFF
--- a/src/components/Router/GoBackLink.jsx
+++ b/src/components/Router/GoBackLink.jsx
@@ -12,7 +12,7 @@ function isExactMatch(match) {
   return match?.isExact;
 }
 
-const GoBackLink = ({ label, hierarchy, rootLink, breakpoint }) => {
+const GoBackLink = ({ label, hierarchy, breakpoint }) => {
   const { themeName } = useTheme();
   const isDarkTheme = themeName === DEFAULT_DARK_THEME_NAME;
   const { history, navigationHistory } = useRouter();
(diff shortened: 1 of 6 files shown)


lukebp (approved):
tACK

Needs approval from @tiagoalvesdulce.

tiagoalvesdulce (approved):
LGTM

PR #2659, opened 2021-11-17
fix: Identity error on billing change modal

### LINKED ISSUE #2656: Incorrect identity error on modal.
The identity error is not being displayed correctly when the error is encountered on the billing status modal.

image

victorgcramos:
Closes #2656.

Fixes Identity error message on Set Billing Status modal.

### UI Changes

**Before:**
before

**After:**
2656-billing-status-identity-error

### DIFF
--- a/src/components/ModalSetBillingStatus/ModalSetBillingStatus.jsx
+++ b/src/components/ModalSetBillingStatus/ModalSetBillingStatus.jsx
@@ -14,6 +14,8 @@ import SelectField from "src/components/Select/SelectField";
 import { getProposalBillingStatusOptionsForSelect } from "./helpers";
 import { PROPOSAL_BILLING_STATUS_CLOSED } from "src/constants";
 import styles from "./ModalSetBillingStatus.module.css";
+import { useCrash, useModalContext } from "src/hooks";
+import { isIdentityError } from "src/utils";
 
 const StatusesExplanation = () => (
   <div>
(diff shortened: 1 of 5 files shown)


lukebp:
This is suppose to use the global identity error that was added in #2623.

victorgcramos:
Hmm okay, I thought the issue was with the message, not the error handling. Got it! Fixing it.

lukebp (changes requested):
This is working correctly if the user does not have an identity loaded in the browser at all, but is not working for when the user has the wrong identity loaded in the browser.  

image

You can reproduce this by logging in using pictl and updating the user's identity with the `pictl userupdatekey` command.  The identity saved to the browser will then be incorrect.

tiagoalvesdulce (approved):
This code works. But the error is outside the modal and the only way I can leave the Error page is returning to the Home page.
 
Screen Shot 2021-11-18 at 08 34 50

I think this is not the greatest UX. Can we keep the error inside the Modal? I'm not sure if ErrorBoundaries work with React Portal so the ideal solution might not be using them.

tiagoalvesdulce (changes requested):
Approved before by mistake.

victorgcramos:
@tiagoalvesdulce this was defined on #2623. Maybe it's outside the pr scope... Let's open an issue and discuss better solutions for it, what do you think?

tiagoalvesdulce:
oh, forgot about #2623. I'm fine with it but I think we should improve that in the next release. What do you think @lukebp ?

lukebp:
Let's use the global error for now so that we can get the release pushed then open an issue for further improving this UX.

lukebp (approved):
tACK

tiagoalvesdulce (approved):
LGTM. @victorgcramos please create the issue so we don't forget to improve it.

PR #2661, opened 2021-11-17
fix: Display 2FA on Login Modal

victorgcramos:
This diff fixes a bug introduced by #2638, where modal overlays weren't
triggering visual updates, hence, modals were not being replaced, and the
2FA digits input couldn't be seen:

The error can be reproduced when logging in with a 2FA active account
using the "What are your thoughts" login button.

2fa-login-modal

- Also, adds e2e tests to cover this case.

Looks like this now:
2fa-login-modal-after

### DIFF
--- a/src/components/ModalProvider.jsx
+++ b/src/components/ModalProvider.jsx
@@ -15,7 +15,7 @@ const ModalProvider = ({ children }) => {
   const modalToDisplay = useRef(initialState);
   // Modals are refs, and refs doesn't trigger component re-render. This
   // will force the state update, hence a modal update.
-  const [, setIsDisplaying] = useState(false);
+  const [, toggleDisplay] = useState(Date.now());
 
   const handleOpenModal = useCallback(function (
     modal,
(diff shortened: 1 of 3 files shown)


lukebp (approved):
tACK

Needs approval from @tiagoalvesdulce.

src/components/ModalProvider.jsx
```diff
@@ -15,7 +15,7 @@ const ModalProvider = ({ children }) => {
   const modalToDisplay = useRef(initialState);
   // Modals are refs, and refs doesn't trigger component re-render. This
   // will force the state update, hence a modal update.
-  const [, setIsDisplaying] = useState(false);
+  const [, toggleDisplay] = useState(Date.now());
```
tiagoalvesdulce, line 18: why are you using `Date.now()`?

tiagoalvesdulce (commented):
Left an inline comment

src/components/ModalProvider.jsx
victorgcramos: It's a method that I am sure that will return me different values for any call. The purpose here is to force a component re-render. Do you have a better idea?

tiagoalvesdulce: Yeah, you can use a dispatch function update. https://stackoverflow.com/questions/46240647/react-how-to-force-a-function-component-to-render/53837442#53837442

victorgcramos: nice! thanks!

tiagoalvesdulce (approved):
tACK

PR #2663, opened 2021-11-19
fix(Proposals List): do not reorder the list

### LINKED ISSUE #2654: List view re-ordering bug.
This bug was discovered on test-proposals.decred.org and can be reproduced there right now as well.  I was using firefox.

1. Navigate to the "Approved" list view.  The proposals are ordered correctly.
2. Navigate to the "Vote on Me" proposal details back.
3. Click your "Back" browser button to go back to the list views.
4. The gui is re-ordering the proposals for some reason and displaying them in the incorrect order now.  The "Vote on Me" proposal has been moved down in the list when it shouldn't have been.

https://user-images.githubusercontent.com/9039877/142009501-324bbad7-d951-4ef4-927b-21b075c431fe.mp4

tiagoalvesdulce:
The list was being reordered on RECEIVE_VOTES_INVENTORY because the
updateInventory function was inserting tokens into the front of the
allByVoteStatus[STATUS] array. With this commit now it inserts into
the back.

Closes #2654 

### Debugging

This was a very hard bug to reproduce. Thanks to redux devtools I could narrow the issue. Here's how I did it:

1. At first I couldn't repro the bug locally and thought it was related to proposals submissions. I was wrong.

2. As I couldn't repro the bug locally I went to https://test-proposals.decred.org/?tab=approved with the devtools network tab open and started to see what might be causing the bug. Then I saw something intriguing:

Screen Shot 2021-11-19 at 08 44 04
Screen Shot 2021-11-19 at 08 44 40

The first records were the ones from the second inventory call. I had a pretty good guess: **it's related to the `ticketvote/inventory` somehow.**

3. Now back to my local environment I just created a lot of approved proposals and noticed I could repro not only doing the steps described on #2654 but also when scrolling to the end of the list.

4. Then, if you can repro a bug locally it's pretty easy to spot the culprit with `redux-devtools`:

Here is the approved records token list before the `RECEIVE_VOTES_INVENTORY` action:

Screen Shot 2021-11-19 at 08 47 24

After the action:

Screen Shot 2021-11-19 at 08 41 03

So the tokens from the second inventory page was being inserted into the front of the `allByStatys["approved"]` list and the Proposals List in the UI is just a representation of this list.

5. Now the easy part. I went to that action and changed the state update function to insert new inventory tokens into the back. See a before and after the `RECEIVE_VOTES_INVENTORY` action now:

Screen Shot 2021-11-19 at 08 47 24
Screen Shot 2021-11-19 at 08 47 43

It's fixed! 🎉

### DIFF
--- a/src/reducers/models/proposals.js
+++ b/src/reducers/models/proposals.js
@@ -137,7 +137,10 @@ const updateProposalRfpLinks = (proposal) => (state) => {
   )(state);
 };
 
-const updateInventory = (payload) => (allProps) => ({
+const updateInventory = (payload) => (allProps) => {
+  console.log("payload", payload);
+  console.log("allProps", allProps);
+  return ({
   ...allProps,
   ...Object.keys(payload).reduce((res, status) => {
     const propsStatus = allProps[status] ? allProps[status] : [];
@@ -146,13 +149,14 @@ const updateInventory = (payload) => (allProps) => ({
       ...res,
       [status]: [
         ...new Set([
-          ...payloadStatus.map((token) => shortRecordToken(token)),
-          ...propsStatus
+          ...propsStatus,
+          ...payloadStatus.map((token) => shortRecordToken(token))
         ])
       ]
     };
   }, {})
 });
+};
 
 const onReceiveLogout = (state) =>
   compose(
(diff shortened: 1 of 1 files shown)


victorgcramos (approved):
ACK

amass01 (approved):
ACK

nice bug hunting

lukebp (approved):
ACK

Issue #2687, opened 2021-12-06
Quorum text improvement on mobile


lukebp:
A user provided feedback that quorum text on mobile is confusing.  `1997 votes left` does not mention that this is how many votes are left before a quorum is reached, which implies that there are only a total of 1997 votes left.

![image](https://user-images.githubusercontent.com/9039877/144881982-c2936bb6-bf38-4cb7-a00d-8b0bec656a05.png)

lukebp:
After some internal discussion, we decided to go with two lines on top of each other. The first line is the quorum requirement.  The second line displays the number of cast votes and the number of eligible votes.

```
8000 votes until quorum
1015/40960 votes
```

```
quorum reached!
9015/40960 votes
```

PR #2695, opened 2021-12-20
fix: quorum text

tiagoalvesdulce:
Closes https://github.com/decred/politeiagui/issues/2687

### Dependencies

Depends on https://github.com/decred/pi-ui/pull/412

### UI Changes Screenshot

**Desktop**
Screen Shot 2021-12-20 at 09 21 35

**Mobile**
Screen Shot 2021-12-20 at 09 48 59

### DIFF
--- a/src/components/Proposal/Proposal.jsx
+++ b/src/components/Proposal/Proposal.jsx
@@ -35,6 +35,7 @@ import {
   isActiveRfp,
   isEditableProposal,
   getQuorumInVotes,
+  getEligibleTickets,
   isVoteActiveProposal,
   isVotingFinishedProposal,
   getProposalToken,
@@ -430,6 +431,7 @@ const Proposal = React.memo(function Proposal({
                   renderStatusInfoComponent={
                     <VotesCount
                       isVoteActive={isVoteActive}
+                      eligibleVotes={getEligibleTickets(voteSummary)}
                       quorumVotes={getQuorumInVotes(voteSummary)}
                       votesReceived={getVotesReceived(voteSummary)}
                       onSearchVotes={!isLegacy ? openSearchVotesModal : null}
--- a/src/components/Proposal/Proposal.module.css
+++ b/src/components/Proposal/Proposal.module.css
@@ -54,6 +54,13 @@
 .voteCount {
   display: flex;
   user-select: none;
+  align-items: flex-end;
+}
+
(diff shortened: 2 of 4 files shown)


lukebp (changes requested):
image

This is not showing the total eligible votes.  This should say `2/5353 votes`.

tiagoalvesdulce:
Right, my bad.

I'm not feeling good about the UX in this PR tho. Here is a draw of our progress bar with the changes made by this PR:

Screen Shot 2022-01-17 at 17 42 24

While the progress bar max value is the number of votes to reach quorum, the text is displaying votes casted / eligible votes. Pretty confusing.

I suggest we go ahead and implement two bars, one for quorum and one for % of votes. Like @linnutee proposed here: https://github.com/decred/politeiagui/issues/2704. We can get rid of the quorum bar once its reached and show a `quorum reached` text. What do you think @victorgcramos @lukebp @amass01 ?

victorgcramos:
I like linutee's suggestion.

victorgcramos:
See how it looks if we split the view and place the status bars side by side:
Screen Shot 2022-01-17 at 6 50 25 PM

lukebp:
I think you guys are making this way more complicated than it needs to be.  The text from the orignal comment is simple and intuitive.

https://github.com/decred/politeiagui/issues/2687#issuecomment-986981529

A normal person is going to see `100/5000` as the number of votes that have been cast out of the total number of possible votes.  The number of votes until quorum is reached is also displayed.  This lines up with the status bar.

lukebp (approved):
tACK
image

linnutee:
I commented there in #2704 that adding that two separate bars makes only sense in a proposal specific view. That one has the real-estate to include all the details to make it clear enough. Otherwise much repetition in proposals list view, as that's much more packed already. 

I'd go for a more reductive approach in the prop. list view. Though this is quite dated https://voting.decred.org/ I think solving it with either some icon block as part of the status group, rather than yes/no voting would be optimal. Previous screenshot from Luke is in a sense simple, but adds  too much different type styles w/o alignment. 

Couple quick drafts to test and validate https://www.figma.com/file/igi4pdXlHhTiGyDIF9A2IR/dcr---design-guidelines---id-and-visual-comms?node-id=2%3A7429 :
Screenshot 2022-01-28 at 15 21 02

PR #2701, opened 2021-12-29
Show proposal vote block heights

### LINKED ISSUE #2699: Show proposal vote block heights.
The gui doesn't show the block start and end heights anywhere for proposal votes.  This is relevant info that should be displayed somewhere.  Maybe in the tooltip.

vibros68:
Closes #2699 

This diff add  the block start and end heights on the gui for proposal
votes. 

Screenshot from 2022-01-13 15-14-40
Screenshot from 2022-01-13 15-14-53

### DIFF
--- a/src/components/Proposal/Proposal.jsx
+++ b/src/components/Proposal/Proposal.jsx
@@ -159,6 +159,8 @@ const Proposal = React.memo(function Proposal({
     startDate,
     endDate
   } = proposal;
+  const { startblockheight, endblockheight } = voteSummary;
+  const votingBlocks = Number(endblockheight) - Number(startblockheight);
   const isAdmin = useSelector(sel.currentUserIsAdmin);
   const isVetted = state === PROPOSAL_STATE_VETTED;
   const isRfp = !!linkby || type === PROPOSAL_TYPE_RFP;
@@ -397,7 +399,10 @@ const Proposal = React.memo(function Proposal({
                     />
                   )}
                   {isVoteActive && (
-                    <>
+                    <Tooltip
+                      placement="bottom"
+                      content={`Block heights: ${startblockheight} - ${endblockheight}`}
+                      contentClassName={styles.quorumTooltip}>
                       <Text
                         className={classNames(
                           "hide-on-mobile",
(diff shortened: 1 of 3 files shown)


vibros68:
Ready for review @lukebp

tiagoalvesdulce (approved):
Looks good but it won't show on mobile since we aren't showing blocks left too.

vibros68:
@tiagoalvesdulce How do you think? Should we display blocks left on mobile?

src/components/Proposal/Proposal.jsx
```diff
@@ -397,7 +399,10 @@ const Proposal = React.memo(function Proposal({
                     />
                   )}
                   {isVoteActive && (
-                    <>
+                    <Tooltip
+                      placement="bottom"
+                      content={`Block heights: ${startblockheight} - ${endblockheight}`}
```
lukebp, line 404: Let's update this to just:

```suggestion
                      content={`Block ${startblockheight} to ${endblockheight}`}
```

src/components/Proposal/Proposal.jsx
```diff
@@ -408,7 +413,24 @@ const Proposal = React.memo(function Proposal({
...
+                      contentClassName={styles.quorumTooltip}>
+                      <Text
+                        className={classNames(
+                          "hide-on-mobile",
+                          styles.blocksLeft
+                        )}
+                        size="small">
+                        {`duration: ${votingBlocks} block${
```
lukebp, line 429: Let's replace this with the text `Block ${startblockheight} to ${endblockheight}` and remove the tooltip here since it's redundant.

lukebp (changes requested):

lukebp:
> Should we display blocks left on mobile?

No, I think it's fine as it.  It's ok if mobile doesn't have the tooltip info.

tiagoalvesdulce:
Yeah, I think it's fine as well.

lukebp (changes requested):
Two more UX tweeks.

1.  Let's not capitalize `Block` in `Block xxxx to xxxx`.   `block xxxx to xxxx` is more consistent with the other text.

2. After seeing this, I think it's too cluttered.

image

For proposals that have finished voting, can we display the `block xxxx to xxxx` as a second line in the exiting tooltip that shows the date.  i.e. the tooltip would display:

```
Mon, 10 Jan 2022 18:38:09 GMT
block xxxx to xxxx
```

Lets see how that looks.  I think it will be cleaner.

src/components/Proposal/Proposal.jsx
```diff
@@ -414,16 +418,16 @@ const Proposal = React.memo(function Proposal({
...
                       className={classNames(
                         "hide-on-mobile",
                         styles.blocksLeft
                       )}
                       size="small">
-                      {`Block ${startblockheight} to ${endblockheight}`}
+                      {`block ${startblockheight} to ${endblockheight}`}
                     </Text>
-                  )}
+                  )}*/}
```
tiagoalvesdulce, line 430: comment

vibros68: Sorry. My bad

lukebp (approved):
I do like it better this way.  What do you think @tiagoalvesdulce?

image
image

tiagoalvesdulce:
I agree.

tiagoalvesdulce (approved):
Code looks good and works as expected.

PR #2680, opened 2021-11-30
fix: show proposal published and edited dates

victorgcramos:
Closes #2675 

This commit displays the missing "published" timestamp event for
proposals that have been edited. Also, display timestamps for mobile
views and adds e2e tests to prevent this from happening further.

### UI Changes Screenshot

**Regular view**:

List:
Screen Shot 2021-11-29 at 9 03 10 PM

Details:
Screen Shot 2021-11-29 at 9 04 36 PM

**Mobile View**

List:
Screen Shot 2021-11-29 at 9 02 55 PM

Details:
Screen Shot 2021-11-29 at 9 03 21 PM

PR #2712, opened 2022-01-13
fix: Show published date for unvetted proposals.

victorgcramos:
Introduced by #2680 

This diff fixes a bug on unvetted proposals published dates.

Since Unvetted proposals have no edit metadata, we can't load the
`publishedat` timestamp, hence, the published timestamp was always 0,
which resulted on a Jan 1 1970 date.

#### Unvetted Proposals:

Before:
Screen Shot 2022-01-13 at 7 20 10 PM

 
After:
Screen Shot 2022-01-13 at 7 20 37 PM

#### Vetted Proposals

> Changes only affect unvetted proposals, as you can see below:

Before:
Screen Shot 2022-01-13 at 7 23 02 PM

After:
Screen Shot 2022-01-13 at 7 22 15 PM

### DIFF
--- a/src/components/Proposal/Proposal.jsx
+++ b/src/components/Proposal/Proposal.jsx
@@ -342,9 +342,17 @@ const Proposal = React.memo(function Proposal({
                       timestamp={linkby}
                     />
                   )}
-                  <span data-testid="proposal-published-timestamp">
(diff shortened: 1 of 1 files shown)


tiagoalvesdulce (approved):
ACK

PR #2737, opened 2022-03-08
core: Get correct params from route.

### LINKED ISSUE #2736: core: Router match bug on nested routes
Currently our router does not support nested routes with variables followed by static routes. For example, if we declare a route using the `/record/:token/raw`, and we navigate to `/record/abcdefg/raw`, the `token` param will be `abcdefg/raw`, which is not correct.

The expected behavior is to return `abcdefg` as the `token` param.

victorgcramos:
Closes #2736

### DIFF
--- a/plugins-structure/packages/core/src/router/router.test.js
+++ b/plugins-structure/packages/core/src/router/router.test.js
@@ -181,17 +181,20 @@ describe("Given the router", () => {
   it("should find correct matches for nested routes", () => {
     let matchNested = findMatch(mockNestedRoutes, "/test-url/5/test/1");
     let [route, arg1, arg2] = matchNested.result;
+    expect(matchNested.route.path).toEqual("/test-url/:id/test/:id2");
     expect(route).toEqual("/test-url/5/test/1");
     expect(arg1).toEqual("5");
     expect(arg2).toEqual("1");
 
     matchNested = findMatch(mockNestedRoutes, "/test-url/4");
     [route, arg1] = matchNested.result;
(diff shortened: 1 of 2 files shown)


plugins-structure/packages/core/src/router/router.test.js
```diff
@@ -157,6 +170,14 @@ describe("Given the router", () => {
     expect(router.matchPath("/testrl/5/test/1")).toBeFalsy();
     expect(router.matchPath("/test-url/5/tes/1")).toBeFalsy();
   });
+
+  it("should return correct params", () => {
```
tiagoalvesdulce, line 174: I don't think this is the best approach to test this. Can we make `verifyMatch()` return the `match` and test against the returned object?

tiagoalvesdulce (changes requested):
Left a comment

plugins-structure/packages/core/src/router/router.test.js
tiagoalvesdulce: Even better, I think we can make a pure function that receives two parameters, `routes` and `route`. This function will be pure and return a match object like the one we are using. This function can be exported and will be really easy to test. then we will have another function responsible to navigate to the matched route and will call our new pure function to get where it should navigate to:

```
goToMatch() {
   const match = verifyMatch(settings.routes, window. location.pathname);
   // Set title, call view cb and set cleanup
}
```

tiagoalvesdulce (approved):
LGTM

PR #2738, opened 2022-03-11
feat: Add API error handlers.

victorgcramos:
This diff adds api error handlers for ticketvote and records.

Previously, our error messages were retrieved from response's status
text, which isn't useful in terms of UX.

Now, each package has its own custom error handlers that transforms
response's error codes into human readable messages.


plugins-structure/packages/core/src/client/client.js
```diff
@@ -66,8 +73,8 @@ export async function getCsrf(state) {
 export async function parseResponse(response) {
   const { status, statusText } = response;
   const json = await response.json();
-  if (status === 200) return json;
-  throw Error(statusText);
+  if (status >= 200 && status <= 299) return json;
```
tiagoalvesdulce, line 76: I think we shouldn't expect successful statuses other than 200 and 201. 202, for example, can make the API down. I think if politeia ever returns a different status code it will be an edge case to be handled.

tiagoalvesdulce (approved):
LGTM, left an inline comment

tiagoalvesdulce (approved):
ACK

tiagoalvesdulce (approved):

PR #2741, opened 2022-03-17
feat(common-ui): Add Markdown component.

victorgcramos:
This diff adds the scope for Politeiagui's Markdown component.

⚠️ Depends on https://github.com/decred/pi-ui/pull/429 ⚠️ 

Go to `localhost:3007` for a better example.

Includes:
- Markdown Renderer
- Markdown Diff
- Markdown Editor and Commands
- Unit tests for created components and commands

Pros:
- Remove deprecated React-mde package
- Improve markdown diff for lines


tiagoalvesdulce (changes requested):
Looking pretty good. I think some stuff can be improved:

1. We can make our blockquotes look like github's. Example:

 
Screen Shot 2022-03-22 at 16 07 42

Will render:

Screen Shot 2022-03-22 at 16 07 47

But with our markdown editor the same input will render:

Screen Shot 2022-03-22 at 16 08 41

2. We can make our ordered list automatically jump to the next item, like github does:

numbered_list

3. Any plans to add keyboard shortcuts? Currently on politeia we can make text bold with Cmd+B and italic with Cmd+I. Would be nice to have key bindings for all the itens in the toolbar.
4.  You forgot to add the ordered list to the toolbar.
5. I think we can add the task list to the toolbar as well. I think it would be pretty useful for proposal updates.

plugins-structure/packages/common-ui/src/components/Markdown/MarkdownEditor/commands.js
```diff
@@ -0,0 +1,73 @@
+import BoldSVG from "./assets/bold.svg";
+import ItalicSVG from "./assets/italic.svg";
+import QuoteSVG from "./assets/quote.svg";
+import CodeSVG from "./assets/code.svg";
+import BulletListSVG from "./assets/bulletList.svg";
+// import { ReactComponent as ImageSVG } from "./assets/image.svg";
+// import { ReactComponent as NumberedListSVG } from "./assets/numberedList.svg";
```
tiagoalvesdulce, line 7: comments

plugins-structure/packages/common-ui/src/components/Markdown/MarkdownEditor/commands.js
```diff
@@ -0,0 +1,73 @@
+import BoldSVG from "./assets/bold.svg";
+import ItalicSVG from "./assets/italic.svg";
+import QuoteSVG from "./assets/quote.svg";
+import CodeSVG from "./assets/code.svg";
+import BulletListSVG from "./assets/bulletList.svg";
```
tiagoalvesdulce, line 5: I think we can make those pi-ui icons, right?

victorgcramos: That would be great!

plugins-structure/packages/common-ui/src/components/Markdown/MarkdownEditor/MarkdownEditor.js
```diff
@@ -25,16 +30,33 @@ export function MarkdownEditor({
...
-  };
+
+  const handleCommand =
+    (command, { offset = 0 } = {}) =>
+    () => {
+      const content = editorRef.current.value;
+      const startPos = editorRef.current.selectionStart;
+      const endPos = editorRef.current.selectionEnd;
+      const newString = executeCommand(command, content, startPos, endPos);
+      // prevent sync issues between textarea updates.
```
tiagoalvesdulce, line 41: We don't need this setTimeout to achieve the desired behavior here.

First, let's understand why this code actually works. The simple explanation is:

**setState is only async batched when called inside a React event handler.** As `setTimeout` is not a react event handler the `setValue` will be sync here. Crazy, huh? But that's how it works.

A more detailed explanation can be found here: https://github.com/facebook/react/issues/11527#issuecomment-360199710

Now we already know that it's a _hacky_ way to make things work but how do we do it idiomatically? I suggest we use a **useEffect** hook instead. See:

Create a new state:

```
  const [lastCommand, setLastCommand] = useState({});
```

Inside the `handleCommand` function, remove the `setTimeout` hack and call `setLastCommand` with `setValue` and `onChange`. The function will look something like this:

```
const handleCommand =
    (command, { offset = 0, label } = {}) =>
    () => {
      const content = editorRef.current.value;
      const startPos = editorRef.current.selectionStart;
      const endPos = editorRef.current.selectionEnd;
      const newString = executeCommand(command, content, startPos, endPos);
      setLastCommand({ label, offset, startPos, endPos });
      setValue(newString);
      onChange(newString);
    };
```

Now we only have to write a `useEffect` to manipulate the focus like we want and reset the `lastCommand` state:

```
  useEffect(() => {
    if (!isEmpty(lastCommand)) {
      setLastCommand({});
      const { offset, startPos, endPos } = lastCommand;
      editorRef.current.focus();
      editorRef.current.selectionStart = startPos + offset;
      editorRef.current.selectionEnd = endPos + offset;
    }
  }, [lastCommand]);
  ```

tiagoalvesdulce (changes requested):
Left a comment

tiagoalvesdulce (approved):
LGTM. Great job!! 👏

PR #2742, opened 2022-03-22
multi: Add Themes on Pi app.

victorgcramos:
Our current politeia app depends on the pi-ui theme provider to format
fonts, colors and styles.

In order to persist theme configuration, all pages should get values from
the same theme. However, each route has its own ReactDOM renderer, hence,
all routes would have its own theme config if we use the `ThemeProvider` from pi-ui.
That's because `ThemeProvider` uses Context API for state management.

The core idea behind this diff is to use the same theme config among the whole
app, therefore, use values from the redux-store instead of the context state.

Screen Shot 2022-03-22 at 2 28 59 PM


tiagoalvesdulce (commented):
I agree with moving the theme control to redux so we can access the theme name and switch themes from anywhere in the app. But is there a reason to save all the theme variables in the redux store? That seems unnecessary to me. Reasons:

1. It will never change at runtime. Themes are defined by the developer before building the app. We don't allow the user to choose a new primary color, for example.
2. Too much info to be saved in the store while we could just save the theme name and access variables in a theme object (like pi-ui does).
3. What I would do: store only current theme and theme options in the store.

Unless we are planning to allow users to customize their apps looks at runtime, I think saving theme variables in the redux store is a bad idea.

tiagoalvesdulce (approved):
LGTM

PR #2748, opened 2022-04-13
multi: Add Default Pi Layout on Home Page

victorgcramos:
This PR implements the new pi-ui layout system on our politeia
app-shell home page.

## Preview

Screen Shot 2022-04-13 at 2 24 04 PM


tiagoalvesdulce (changes requested):
Looking great! My only recommendation is about the splash screen/loading indicators. See:

https://user-images.githubusercontent.com/13955303/164750378-3aa460e9-52f5-4b76-aec8-1b011936f2c0.mp4

I think we should have a splash screen + skeleton for proposals cards and no flickering. What do you think?

victorgcramos:
@tiagoalvesdulce good suggestion, mate! Thanks!

victorgcramos:
After discussing with @tiagoalvesdulce on Matrix, we agreed that we should handle the loading behavior on another PR. Hence, I removed the loading screen component from page layouts.

tiagoalvesdulce (approved):
LGTM

PR #2754, opened 2022-04-25
feat(cli): add plugins flag to newapp command

tiagoalvesdulce:
This PR aims to extend the `newapp` CLI to accept a `plugins` flag with a list of comma separated plugins to add to the app.

Todo:

- [x] Add plugins flag to CLI
- [x] Add plugins to the `package.json` of the newly created app
- [x] Create basic config template

Decided to create a basic api for plugins config file and implement template after we have a better idea how all core packages will be integrated.

### DIFF
--- a/plugins-structure/packages/cli/README.md
+++ b/plugins-structure/packages/cli/README.md
@@ -6,9 +6,11 @@ line structure and arguments.
 
 ## Overview
 
-1. [Setup](#setup)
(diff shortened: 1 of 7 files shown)


victorgcramos (commented):
Looking really good! That's such a great improvement for app creation!

I think it would be nice if the `newapp` command logs the config if you loaded it from a file using the `--config` flag.

lukebp (approved):

PR #2755, opened 2022-04-26
multi: Add Proposal Loader + Store Listeners

victorgcramos:
This PR adds loading placeholders for loading proposals. Fixes https://github.com/decred/politeiagui/pull/2748#issuecomment-1106785947

Screen Shot 2022-04-26 at 8 38 37 PM

It also introduces the new `useStoreEffect` hook, which uses the also implemented `addStoreEffect` listener to our Store.

#### Usage

You can use the `addStoreEffect` outside the React scope, and trigger a side effect when some action is dispatched. In the following example, we will add a listener, using the regarded method, to log a "Hello" message on the console after `api` is fetched:

```javascript
addStoreEffect(() => {
  console.log("Hello!");
}, [api.fetch.fulfilled]);
```

Easy, huh?

If you're using it on some react component, and don't want it to be triggered on every render, you can use the `useStoreEffect` hook. In the following example, we will fetch records and ticketvote policies after api is fetched:

```javascript
 useStoreEffect(() => {
    dispatch(ticketvotePolicy.fetch());
    dispatch(recordsPolicy.fetch());
  }, [api.fetch.fulfilled]);
```

### DIFF
--- a/plugins-structure/apps/politeia/src/components/Proposal/ProposalLoader.js
+++ b/plugins-structure/apps/politeia/src/components/Proposal/ProposalLoader.js
@@ -1,0 +1,21 @@
+import React from "react";
+import { RecordCard } from "@politeiagui/common-ui";
+import styles from "./styles.module.css";
+
+const Block = () => <div className={styles.block} />;
+
+function ProposalLoader() {
+  return (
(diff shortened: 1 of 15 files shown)


tiagoalvesdulce (approved):
LGTM

PR #2758, opened 2022-04-28
comments: Markdown + Error Handlers

### LINKED ISSUE #2747: [plugin-architecture] Add Markdown Renderer on Comments
The comment below contains markdown escape characters, which creates a bold text.

Screen Shot 2022-04-06 at 3 52 28 PM

Currently, the text is being rendered as a raw string, but we want it to be properly parsed and highlighted using the `MarkdownRenderer` from `@politeiagui/common-ui`.

### LINKED ISSUE #2756: [plugin-architecture] Handle comments error messages
Comments error messages should be handled properly so we can provide the user a better feedback about what caused the error. 

Currently, `ticketvote` plugin and `records` layer both have their own error handlers, which converts API error codes into a human readable message. Comments should also have those handlers.

victorgcramos:
Closes #2756 
Closes #2747


tiagoalvesdulce (approved):
LGTM

PR #2751, opened 2022-04-19
[WIP] multi: Add Proposal New page

victorgcramos:
⏳ Work in Progress... ⏳ 

This PR implements the Proposal New page, using the also implemented
RecordForm structure.

#### Pi
- [x] Add Proposal New page (`/record/new`)
- [ ] Use single page layout (#2748)
- [ ] Add "Create Proposal" Banner
- [ ] Get domains from pi policy (https://github.com/decred/politeiagui/pull/2723/commits/e8ede2991ff4f6c5984f9656440c62a20e9f439d)

#### Common-ui
- [x] Use react-hook-form instead of Formik.
- [x] Text Input
- [x] Markdown Input
- [x] Select Input
- [ ] Date Picker input

### Preview
Screen Shot 2022-04-19 at 7 14 49 PM

PR #2760, opened 2022-04-29
comments: Remove formik and use react-hook-form

victorgcramos:
This diff fixes the currently broken CommentForm component by 
removing the unused formik package and using react-hook-form.


tiagoalvesdulce (changes requested):
You're adding the `react-hook-form` here as a Comments dep and on https://github.com/decred/politeiagui/pull/2751 as a Common-ui dep. We can lift it up and add it only as a workspace app.

tiagoalvesdulce (approved):
LGTM

PR #2763, opened 2022-05-10
multi: Add Navbar

victorgcramos:
This PR adds the Navbar component implementation on common-ui,
and adds its on our politeia app-shell.

common-ui package:
- Add Navbar component
- Add ThemeToggle component

politeia app:
- Add Navbar on app
- add politeia logo

### UI Preview

Light Theme:
Screen Shot 2022-05-10 at 6 27 17 PM

Dark Theme:
Screen Shot 2022-05-10 at 6 27 33 PM


tiagoalvesdulce (changes requested):
1. The Politeia Logo should be clickable and redirect to the `Home`.
2. Where is the `.splash` class used?
3. How would I add a hamburger menu to the nav when the children on the right are wrapped?
4. Your `React.Children.maps` on `Navbar` are a bit confusing. Can you add comments or  think about alternative implementations?
5. A little observation: If we are always importing `Column as Col` from `pi-ui`, why not exporting `Col` instead of `Column`?
6. The navbar is looking a little bit different than the actual prod app. See:

Prod:
Screen Shot 2022-05-11 at 17 35 25

This PR:
Screen Shot 2022-05-11 at 17 35 45

Background color, font color and spacing are different. Spacing is pretty minimal and if it looks good I'm OK but we should match background color and font color.

7. Responsive version can be improved a lot comparing to the prod version.

victorgcramos:
@tiagoalvesdulce sorry, PR still WIP. Forgot to make it a draft. My bad. Thanks for the feedback, though! I'll keep that in mind!

victorgcramos:
Navbar for small screens:

Screen Shot 2022-05-13 at 6 55 44 PM

tiagoalvesdulce (approved):
LGTM

PR #2775, opened 2022-05-31
multi: Plug and Play structure.

victorgcramos:
### Plug and Play

Some of our plugins tools require some prior information loaded in order to
execute some actions. For example, the ticketvote inventory plugin needs
the ticketvote policy previously loaded in order to properly paginate requests.

When we talk about something that is "plug and play", we should assert that some
plugin will have a default behavior without needing to configure anything. Just
plug and play :).

#### Plugin Setup

The core package also provides good utils for plugins creation, which helps us
managing the reducers and plugin intializers.

> An `initializer` is an `{ id, action }` object, which will setup pugins custom
> setup for some given use case described by the `id`.

Let's create a new custom plugin using the `pluginSetup` util from core package:

```javascript
// myplugin/plugin.js
import { pluginSetup, store } from "@politeiagui/core";
import { fetchApi, myReducer } from "./myPlugin";

const MyPlugin = pluginSetup({
  initializers: [
    { id: "myplugin/fetch", action: () => store.dispatch(fetchApi()) },
  ],
  reducers: [{ key: "myPlugin", reducer: myReducer }],
  name: "my-plugin",
});

export default MyPlugin;
```

You can initialize some plugin using the `initialize` method. Example:

```javascript
import React from "react";
import MyPlugin from "@politeiagui/my-plugin";

MyPlugin.initialize("myplugin/fetch");

function Component(props) {
  return <div>My Component</div>;
}
```

Easy, huh? Just use this interface and then use your politeiagui plugin on your
application shell.

#### App Setup

An application is just used to connect all plugins into an integrated interface.
In order to provide a good experience for developers who want to create a new
politeiagui application, we recommend using the `appSetup` util from the core
package.

Example:

```javascript
// app.js

import { appSetup } from "@politeiagui/core";
// Plugins
import MyPlugin from "@politeiagui/my-plugin";
import ExternalPlugin from "external-plugin";

const App = appSetup({
  plugins: [MyPlugin, ExternalPlugin],
  config: {
    name: "My App",
    foo: "bar",
  },
});

export default App;
```

Let's initialize the application with a `routes` array defined using the
`createRoute` method from our configured app.

The `createRoute` method receives an objecto containing a route `path`, a `view`
some `initializerIds` to **initialize plugins** for given path, and a `cleanup`
to be executed when route becomes inactive.

```javascript
// index.js
// Let's use React, but our setup is framework agnostic and can be used on any
// javascript application
import React from "react";
import ReactDOM from "react-dom";
import App from "./app";

const root = document.querySelector("#app-root");

const routes = [
  App.createRoute({
    path: "/",
    view: () => ReactDOM.render(<div>My App</div>, root),
    initializerIds: ["myplugin/fetch"],
    cleanup: () => ReactDOM.unmountComponentAtNode(root);
  })
]

App.init({ routes });
```

So, as described above, once the user hits `/`, before rendering the `view`, it
will execute the `myplugin/fetch` initializer so our view can be ready to use
without any additional setup.

### DIFF
--- a/plugins-structure/packages/core/README.md
+++ b/plugins-structure/packages/core/README.md
@@ -14,7 +14,6 @@ package management tool on any JavaScript application or plugin.
 - [Client](#core-client)
 - [Records](#core-records)
 - [Router](#core-router)
-- [Plugins Router](#core-router-plugins)
 - [Plugin Setup](#core-plugin-setup)
 - [App Setup](#core-app-setup)
 - [References](#references)
@@ -33,6 +32,9 @@ package management tool on any JavaScript application or plugin.
 │   │   ├── /policy
 │   │   ├── /records
 │   │   ├── /ui
+│   │   ├── constants.js
+│   │   ├── errors.js
+│   │   ├── initializers.js
 │   │   ├── utils.js
 │   │   ├── utils.test.js
 │   │   ├── validation.js
@@ -1338,16 +1340,10 @@ Example:
 ```javascript
 // app.js
 
-// Let's use React, but our setup is framework agnostic and can be used on any
-// javascript application
-import React from "react";
-import ReactDOM from "react-dom";
 import { appSetup } from "@politeiagui/core";
 // Plugins
 import MyPlugin from "@politeiagui/my-plugin";
 import ExternalPlugin from "external-plugin";
-
-const root = document.querySelector("#app-root");
 
 const App = appSetup({
   plugins: [MyPlugin, ExternalPlugin],
@@ -1369,7 +1365,13 @@ to be executed when route becomes inactive.
 
 ```javascript
 // index.js
+// Let's use React, but our setup is framework agnostic and can be used on any
+// javascript application
+import React from "react";
+import ReactDOM from "react-dom";
 import App from "./app";
+
+const root = document.querySelector("#app-root");
 
 const routes = [
   App.createRoute({


plugins-structure/packages/core/src/router/pluginsRouter.js
```diff
@@ -0,0 +1,142 @@
...
+    },
+  };
+}
+
+/**
+ * configurePluginsRouter returns a configured router to run on app background.
+ * This router won't render views, ans instead, it's used as a listener for
+ * routes changes.
+ */
+function configurePluginsRouter() {
```
tiagoalvesdulce, line 58: I could not understand this either. Are this "calls" running in the background? That means no spinner in the views?

tiagoalvesdulce (commented):
Code looks good. Thanks for this PR!

A couple observations:

1. Why is UI a plugin?
2. Why you named `Plugin` capitalized on `pluginSetup`? It makes me think it's a class that I should create an instance with the `new` keyword but it's in fact a function.
3. One doubt: apps can have `viewRoutes` and plugins don't, right? Why?

tiagoalvesdulce (commented):
When I am using a proxy map I'm just saying: "Look, on route `/` I am planning on using `/records/batch` and `ticketvote/inventory`, please set it up for me so I can start working". Is that correct? If so, this should be documented.

victorgcramos:
Changed the proxy approach to a more readable one. Thanks for the feedback, @tiagoalvesdulce!

Due to this change, there are still some work to be done:
- [ ] fix plugins dev app example
- [ ] improve docs

tiagoalvesdulce:
1. I still think `Plugin` should be renamed to `setupPlugin` or something like that because it's not a Class.
2. I don't think UI should be a plugin
3. I can still see references to `pluginsProxyMap`. Wouldn't that be renamed?

PR #2780, opened 2022-06-03
Minor visual enhancements for proposals

### LINKED ISSUE #2779: [plugin-architecture] Handle censored proposals.
Here's how a censored proposal looks like on politeia app-shell from the new plugin architecture:

Screen Shot 2022-06-02 at 4 08 13 PM

The `censored` status should be displayed, and the background should be changed as well.

victorgcramos:
Closes #2779 

This PR fixes and adds some visual enhancements for proposals view:

- Handle censored and abandoned proposals properly;
- Make politeia navbar logo clickable, and navigate to home on click;

### DIFF
--- a/plugins-structure/apps/politeia/src/components/Header/Header.js
+++ b/plugins-structure/apps/politeia/src/components/Header/Header.js
@@ -1,17 +1,21 @@
 import React from "react";
+import { router } from "@politeiagui/core/router";
 import { Navbar, ThemeToggle, theme } from "@politeiagui/common-ui/layout";
 import { useSelector } from "react-redux";
 import { DEFAULT_DARK_THEME_NAME, Text } from "pi-ui";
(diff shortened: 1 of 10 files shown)


tiagoalvesdulce (approved):
Looks good. Only one thing:

I think you should reconsider using multiple calls to useSelect instead of destructuring. From the [react-redux docs](https://react-redux.js.org/api/hooks):

> When an action is dispatched, useSelector() will do a reference comparison of the previous selector result value and the current result value. If they are different, the component will be forced to re-render. If they are the same, the component will not re-render.

> You may call useSelector() multiple times within a single function component. Each call to useSelector() creates an individual subscription to the Redux store. Because of the React update batching behavior used in React Redux v7, a dispatched action that causes multiple useSelector()s in the same component to return new values should only result in a single re-render.

tiagoalvesdulce (approved):
ACK

PR #2786, opened 2022-06-14
fix: Avoid duplicate inventories tokens

### LINKED ISSUE #2772: [plugin-architecture] Avoid duplicate tokens on ticketvote inventory
Currently, our ticketvote inventory slice does not recognize duplicate tokens, and instead, it populates the array for every returned tokens from `fetchTicketvoteInventory`, even if they are already defined on state.

We should check that and only populate the inventory with different tokens.

### How to Reproduce

If you fetch the same inventory status twice, the state will be populated with the same tokens batch twice.

victorgcramos:
Closes #2772 

This PR fixes the duplicate tokens issue on both records and ticketvote
inventories with unit tests to ensure inventory will be only composed by
uniq tokens

### DIFF
--- a/plugins-structure/packages/core/src/records/inventory/recordsInventory.test.js
+++ b/plugins-structure/packages/core/src/records/inventory/recordsInventory.test.js
@@ -209,7 +209,7 @@ describe("Given the recordsInventorySlice", () => {
             vetted: { public: publicTokens },
             unvetted: {},
           };
-          fetchRecordsInventorySpy.mockResolvedValueOnce(resValue);
+          fetchRecordsInventorySpy.mockResolvedValue(resValue);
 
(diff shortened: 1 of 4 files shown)


tiagoalvesdulce:
In 2772 you said:

> If you fetch the same inventory status twice, the state will be populated with the same tokens batch twice.

In which situation we do that? Isn't better to prevent this instead of filtering out duplicates? I don't see any situation that we should be going to `fetchTicketvoteInventory.fulfilled` with duplicate tokens. Maybe the solution is one step before?

victorgcramos:
We handle that on our politeia app, but I think we should also avoid this behavior on ticketvote plugin. Why not? If our inventory, by design, is an array of unique tokens, it makes no sense to "rely" on an inventory that can break this rule.

Makes sense? That's what I thought when I opened this issue...

PR #2793, opened 2022-06-22
fix(core/router): allow anchor to wrap other elements and fix external links

### LINKED ISSUE #2746: [plugin-architecture] Router link event target bug.
The default click handler for our Router is described below:

```javascript
function onClickHandler(e) {
  if (e.target.matches(settings.selector)) {
    e.preventDefault();
    this.navigateTo(e.target.href);
  }
}
```

The problem is that the event target (`e.target`) is not necessarily the `<a>` wrapper. If we have some element inside the
hyperlink tag, the event target will be the element, hence, if it has the `data-link` flag selector, and doesn't have an `href` prop, the navigation won't happen.

Example:

```javascript
// This WON'T prevent the default behavior
<a href="/my-route" data-link>
  <h1>Header</h1>
</a>
```

```javascript
// This will prevent the default behavior but won't navigate to /my-route
<a href="/my-route">
  <h1 data-link>Header</h1>
</a>
```

```javascript
// It works.
<h1>
  <a href="/my-route" data-link>Header</a>
</h1>
```

### LINKED ISSUE #2767: [plugin-architecture] External URL error on router navigation
`window.history.pushState` does not allow external domain urls, and it breaks the app logging the following error when you pass an external domain to `router.navigateTo` method.

```log
router.js:82 Uncaught DOMException: Failed to execute 'pushState' on 'History': A history state object with URL 'https://google.com/' cannot be created in a document with origin 'https://localhost:3000' and URL 'https://localhost:3000/'
```

Example:
```javascript
<button onClick={() => router.navigateTo("https://google.com")}>
  Navigate!
</button>
```

tiagoalvesdulce:
Closes #2746 
Closes #2767 

Current external links behavior should be changed. We should add a modal to alert the user he is going outside Politeia like we do on prod. Issue to track https://github.com/decred/politeiagui/issues/2794.

### DIFF
--- a/plugins-structure/packages/core/src/router/helpers.js
+++ b/plugins-structure/packages/core/src/router/helpers.js
@@ -67,3 +67,23 @@ export function getMatches(routes, targetRoute) {
     });
   return matches;
 }
+
+/**
+ * searchSelectorElement will look recursevely to the parents of the target of
+ * an event looking if it has a provided selector
+ * @param {HTMLElement} el
+ * @param {string} selector
+ * @returns false if it hits the root or the element if an element that matches
+ * the selector is found
+ */
+export function searchSelectorElement(el, selector) {
+  if (!el || !selector) {
+    return false;
+  }
+  if ((el.matches && el.matches(selector)) || el.selector) {
+    return el;
+  } else {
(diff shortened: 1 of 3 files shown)


victorgcramos (approved):
LGTM! Working as expected. Thanks!

PR #2799, opened 2022-06-27
multi: Plugins Setup fixes

### LINKED ISSUE #2797: [plugins-structure] Plugin setup
1. I think `initializerIds` isn't a good name. It's not clear what this is referring to for a dev creating a brand new app for the first time. I'd rather change it to `pluginInitializerRouteIds` ow something like that.
2. Would be nice to show an error (in the console or in the build process, I'd prefere the second) when a user defines a `initializer` without adding the plugin to the `appSetup` plugins array.
3. I think this function can be improved: https://github.com/decred/politeiagui/blob/bb0ea21cca6af41de761f176ad314776998f6ecf/plugins-structure/packages/core/src/appSetup.js#L65
    - I think the filter is unnecessary if we go with the error approach. _The plugin initializerId is not in initializers? Break and show an error_
    - Can fire all actions in parallel. `await Promise.all([...routeInitializerActions])`
4. When 2 plugins have the same `initializerId` the warning in the `appSetup` works nice. But if the dev tries to add an id to the `createRoute` `initializerIds` array and that id is not unique we should fire an error. Firing an error instead of a warning in the `appSetup` is an option too. It's very confusing to execute multiple actions from the same id.
5. It's not clear how do I use a view route from a plugin in the app. For example, if I have a `/tesplugin` route in my `testplugin` plugin, how do I tell my `testapp` to use `tesplugins's` `/testplugin` route when we hit the `/testplugin` route within the app?
6. We should document that, due to the nature of yarn monorepos, when the dev creates a plugin it should run `yarn` in the root before using it in an app.

victorgcramos:
Closes #2797

- Rename `initializerIds` to `pluginInitializerIds`
- Break the app setup if `pluginInitializerIds` aren't unique or if some id does not exist.
- Plugins initializers executed in parallel
- Add plugins routes integration example using `/statistics` route from `statistics` plugin.
- Move `mergeRoutes` util to core router
- Add docs to inform `yarn` is required to execute the plugin for the first time.


tiagoalvesdulce (approved):
LGTM. Only one thing: I think we should move the `statistics` example to a `newapp` and document it instead of adding it to the `politeia` app. That's because in the final version of the `plugins-structure` the statistics plugin will not be part of `politeia` since it's just a showcase.

PR #2800, opened 2022-06-28
[WIP] Decouple listeners effects

tiagoalvesdulce:
With this PR plugins can export pluggable `effects` that can be used inside a `listener` `effect` to do whatever the plugin want. This way we can reduce the overhead to create integrations. Example:

A common integration is to fetch records, comments counts and ticketvote summaries when the scroll hits the end of the page. Currently we have a listener listening to a `fetchNextBatch` action that is fired by the `InfiniteScroller` when it hits the end of the records loaded. This works great but we are doing all our fetching logic in the listener callback effect. That's not ideal since we want the developer to be able to easily compose plugins. With this PR I decoupled the fetching logic from the listener effect. Now plugins can export effects and the developer can just import and use them inside a listener accordingly. This is a work in progress and I'm planning on creating an util to allow the developer to register listeners and a "chain of effects" easily.

image

### DIFF
--- a/plugins-structure/apps/politeia/src/pages/Home/listeners.js
+++ b/plugins-structure/apps/politeia/src/pages/Home/listeners.js
@@ -1,13 +1,86 @@
 import { listener } from "@politeiagui/core/listeners";
-import { records } from "@politeiagui/core/records";
-import { ticketvoteSummaries } from "@politeiagui/ticketvote/summaries";
 import { commentsCount } from "@politeiagui/comments/count";
+import { ticketvoteSummaries } from "@politeiagui/ticketvote/summaries";
+import { records } from "@politeiagui/core/records";
 import { getTokensToFetch } from "@politeiagui/core/records/utils";
 import { getHumanReadableTicketvoteStatus } from "@politeiagui/ticketvote/utils";
 import isEmpty from "lodash/isEmpty";
 import { fetchNextBatch } from "./actions";
 
 const piFilenames = ["proposalmetadata.json", "votemetadata.json"];
+
+function fetchNextBatchCommentsEffect(state, dispatch, { inventoryList }) {
+  const {
+    commentsCount: { byToken },
+    commentsPolicy: {
+      policy: { countpagesize },
+    },
+  } = state;
+
+  const commentsCountToFetch = getTokensToFetch({
+    inventoryList,
+    lookupTable: byToken,
+    pageSize: countpagesize,
+  });
+
+  return (
+    !isEmpty(commentsCountToFetch) &&
+    dispatch(
+      commentsCount.fetch({
+        tokens: commentsCountToFetch,
+      })
+    )
+  );
+}
+
+function fetchNextBatchRecordsEffect(state, dispatch, { inventoryList }) {
+  const {
+    records: { records: recordsObj },
+    recordsPolicy: {
+      policy: { recordspagesize },
+    },
+  } = state;
+
+  const recordsToFetch = getTokensToFetch({
+    inventoryList,
+    lookupTable: recordsObj,
+    pageSize: recordspagesize,
+  });
+
+  return (
+    !isEmpty(recordsToFetch) &&
+    dispatch(
+      records.fetch({
+        tokens: recordsToFetch,
+        filenames: piFilenames,
+      })
+    )
+  );
+}
+
+function fetchNextBatchTicketvoteEffect(state, dispatch, { inventoryList }) {
+  const {
+    ticketvoteSummaries: { byToken },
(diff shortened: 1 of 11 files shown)


plugins-structure/apps/politeia/src/pages/Home/listeners.js
```diff
@@ -1,92 +1,56 @@
...
+function injectEffect(effect) {
+  return async (
+    { payload },
+    { getState, dispatch, subscribe, unsubscribe }
+  ) => {
+    unsubscribe();
+    const state = getState();
+    const inventoryList = getInventoryList(payload, state);
+    await effect(state, dispatch, { inventoryList });
+    subscribe();
```
victorgcramos, line 20: Noticed that `unsubscribe(); ... subscribe()` pattern on inject methods. If this is required, I think we should provide docs explaining that they need to be controlled, or add this pattern as default for inject methods. Maybe create a HoF that wraps the `injectEffect` prop on initializers and does that by default...

plugins-structure/packages/core/src/appSetup.js
```diff
@@ -55,23 +69,51 @@ export function appSetup({ plugins, config }) {
...
+          return [
+            ...prev,
+            {
+              ...cur,
+              ...el,
+            },
+          ];
+        }
+        return prev;
+      }, []);
```
victorgcramos, line 93: that could be improved, right?

victorgcramos (commented):
Looking good, mate! Left some minor inline comments from this first look.

You might have to change the `validatePlugin` method on `pluginSetup` as well. The first design only supported an `{ id, action }` object, but since you added `effects`, I guess `actions` are not required for an initializer anymore, right?

plugins-structure/packages/core/src/appSetup.js
```diff
@@ -16,24 +17,81 @@ function mergeInitializers(initializers, targetInitializers) {
...
+      idListeners.push({
+        actionCreator: listener.actionCreator,
+        effect: listener.injectEffect(effect),
+      });
+    }
+  }
+  return [...idListeners, ...listeners];
+}
+
+function mergeRouteAndAppInitializersPropoerties(
```
victorgcramos, line 33: typo

plugins-structure/packages/core/src/pluginSetup.js
```diff
@@ -1,35 +1,32 @@
...
 
 /**
- * Initializer is an `action` that will setup the plugin for usage in some given
+ * Service is an `action` that will setup the plugin for usage in some given
  * `id` case.
  * @typedef {{
  *  id: string,
- *  action: InitializerAction
- * }} Initializer
+ *  action: ServiceAction
```
victorgcramos, line 14: Missing `effect`, right?

plugins-structure/packages/core/src/appSetup.js
```diff
@@ -55,23 +105,39 @@ export function appSetup({ plugins, config }) {
...
      */
-    createRoute({ path, view, initializerIds = [], cleanup } = {}) {
-      const routeInitializerActions = initializers
-        .filter((init) => initializerIds.includes(init.id))
-        .map((init) => init.action);
+    createRoute({
+      path,
+      view,
+      setupServices = [],
+      listeners = [],
```
victorgcramos, line 120: I guess we should improve the `listeners` naming here. `setupServices` elements receive a `listener` field (proposals app on home page). Not sure if it's intuitive enough to know what each one mean. One listener accepts `{ type, effect }` while the other accepts `{ actionCreator, injectEffect }` and they are both `listeners`.

plugins-structure/packages/ticketvote/src/ticketvote/utils.js
```diff
@@ -0,0 +1,8 @@
+import { ticketvote } from "./";
+import { store } from "@politeiagui/core";
+
+export function fetchPolicyIfIdle() {
+  if (ticketvote.policy.selectStatus(store.getState()) === "idle") {
+    return store.dispatch(ticketvote.policy.fetch());
+  }
+}
```
victorgcramos, line 8: nice!

victorgcramos (changes requested):
Nice! code looking really good. I like the way you decoupled plugins dependencies. Very clean and readable.

One thing I noticed is that after disabling comments, we should also remove the fields from "Available Downloads" dropdown on Proposal Details. Although it does not break the application when clicked, it shouldn't be there. See:

Screen Shot 2022-07-06 at 4 31 21 PM

Left some minor inline comments.

tiagoalvesdulce:
Can't repro the bug. Can you send me a snapshot of your store?

PR #2805, opened 2022-07-05
fix: Correct grammar on a couple of user pages.

davecgh:
I didn't include screenshots since they are simple text changes.  If the maintainers feel it's really necessary, I can add them.


tiagoalvesdulce (approved):
ACK

PR #2817, opened 2022-07-19
fix(legacy): RFP submissions ctrl+click bug

### LINKED ISSUE #2814: bug: Ctrl+click rfp submission.
When you ctrl + click one of the RFP submissions, the submission proposal is opened in a new tab, but the current tab also navigates to the same submission proposal.  The expected behavior is for the RFP submission to open in a new tab while the existing tab stays on the current page.

The behavior is present on both firefox and chromium.

image

victorgcramos:
Closes #2814 

2814-ctrl-link-rfp-submissions


tiagoalvesdulce (approved):
ACK

PR #2819, opened 2022-07-20
fix: Avoid loop on proposal details page

### LINKED ISSUE #2810: [plugin-architecture] Infinite loop on Details page when request fails
When you navigate to `/record/<invalid-token>` page, infinite `records/v1/details` requests are dispatched. We should handle that, and instead, show a "Record not Found" message.

details-infinite-failed-requests

victorgcramos:
Closes #2810

Screen Shot 2022-07-20 at 4 09 23 PM


tiagoalvesdulce (approved):
ACK

Issue #2812, opened 2022-07-19
[plugin-architecture] Create toolkit package


tiagoalvesdulce:
We currently share a lot of configuration code between plugins (webpack, jest config, babel config, scripts on package.json).

We can create a toolkit package e.g. `politeiagui-scripts` to centralize and manage all scripts and config stuff. This way we can remove a lot of duplicated code from plugins and apps. We will also add an option to `eject` from `politeiagui-scripts` so the plugin author can customize the plugin as they please if necessary.

I'm still considering if we should merge this with the CLI package or not. One thing I might do is to move the template folder from CLI to a `politeiagui-template` package so it can be referenced by `politeiagui-scripts` and `cli`.

victorgcramos:
Good one! We should also pay attention to webpack polyfill so we don't need to manually install deps for each package.

PR #2827, opened 2022-07-22
feat: add politeiagui-scripts CLI toolbox

tiagoalvesdulce:
Closes https://github.com/decred/politeiagui/issues/2812

This PR adds the `politeiagui-scripts` CLI package. It adds 7 scripts used through the politeiagui plugins architecture packages and apps: `start-plugin`, `start-app`, `build-plugin`, `build-app`, `lint`, `prettify` and `test`.

With this we can remove all webpack config from plugins and apps.

## Usage

This is a CLI and exposes a bin called `politeiagui-scripts`.

## Overriding Config

`politeiagui-scripts` allows you to specify your own configuration for things. To have your own config, just add the configuration and `politeiagui-scripts` will use that instead of the internal config.

---

Inspired by `kcd-scripts` and `react-scripts`.


plugins-structure/packages/politeiagui-scripts/utils.js
```diff
@@ -0,0 +1,158 @@
...
+      } else {
+        console.log(chalk.green("Success!"));
+        const { errors, warnings } = stats.toJson({
+          all: false,
+          warnings: true,
+          errors: true
+        });
+        let msg = "";
+        if (errors.length !== 0) {
+          for (error of errors) {
```
victorgcramos, line 119: lint: undeclared var

victorgcramos (changes requested):
Nice refactor, mate! Putting together all config files and packages using a single script for each task is a good move.

One thing I noticed is that we should have a better `help` command interface. It's kinda hard to figure out how to run an app for the first time, and manipulate the commands without docs or code comments... Could you improve it?

Also, we should pay attention to `prettier` default config (https://github.com/prettier/prettier-vscode#prettierconfigpath), which means many devs will need to setup a different config path on vscode. Same is valid for eslint.

Left some inline comments.

victorgcramos (commented):
Docs ✅  Good job!

While testing the CLI package, I noticed that both `newapp` and `newplugin` commands are throwing an error, and are not creating correct plugin/app files. Would you mind take a look into that?

victorgcramos (approved):
LGTM! Good job!

PR #2832, opened 2022-07-24
multi: legacy politeiagui fixes

### LINKED ISSUE #2815: UX suggestions
Here are a couple of UX suggestions that stuck out to me when testing the release.

image

- The button to search the votes looks a bit out of place there.  Seems like it would look better down in line with the other buttons.
- It might be helpful to have a tooltip over the `quorum reached!` that details the quorum requirements.  Ex: `quorum of xx% (xxxx votes) has been reached!`

image

- The total vote count `18/5156 votes` is not shown on the proposals that have finished voting.  I think it's helpful to have that on the proposal card because it shows the particiation rate.

### LINKED ISSUE #2828: [legacy] process is not defined
Screen Shot 2022-07-24 at 17 14 01

Every time an error is fired in `dev` the app becomes unresponsive and the `process is not defined` error shows up on console. This error is caused by `react-error-overlay`. See: https://github.com/facebook/create-react-app/issues/11773

### LINKED ISSUE #2829: [legacy] Cannot read properties of null
`Cannot read properties of null` error happens when we click on proposals imported by `importlegacy`.

Screen Shot 2022-07-24 at 17 31 22

### LINKED ISSUE #2830: [legacy] LaxyList hasMore property warning
Screen Shot 2022-07-24 at 17 14 23

Invalid prop hasMore of type number supplied to LazyList, expected boolean

### LINKED ISSUE #2831: [legacy] Invalid argument supplied to oneOfType, expected an instance of array warning
Screen Shot 2022-07-24 at 17 54 16

tiagoalvesdulce:
Closes #2828 
Closes #2829 
Closes #2830 
Closes #2831 
Closes #2815

### DIFF
--- a/src/components/Proposal/Proposal.jsx
+++ b/src/components/Proposal/Proposal.jsx
@@ -548,7 +548,7 @@ const Proposal = React.memo(function Proposal({
                   {(isVoteActive || isVotingFinished) && !legacytoken && (
                     <ChartsLink token={fullToken} />
                   )}
-                  {(isVoteActive && getVotesReceived(voteSummary) > 0) && (
+                  {isVoteActive && getVotesReceived(voteSummary) > 0 && (
                     <div>
                       <ButtonIcon
                         type="search"
--- a/src/components/Proposal/Proposal.jsx
+++ b/src/components/Proposal/Proposal.jsx
@@ -7,6 +7,7 @@ import {
   useMediaQuery,
   useTheme,
   Tooltip,
+  ButtonIcon,
   DEFAULT_DARK_THEME_NAME
 } from "pi-ui";
 import React, { useMemo } from "react";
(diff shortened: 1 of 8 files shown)


src/components/Proposal/Proposal.jsx
```diff
@@ -547,6 +548,15 @@ const Proposal = React.memo(function Proposal({
                   {(isVoteActive || isVotingFinished) && !legacytoken && (
                     <ChartsLink token={fullToken} />
                   )}
+                  {isVoteActive && getVotesReceived(voteSummary) > 0 && (
```
victorgcramos, line 551:
```suggestion
                  {getVotesReceived(voteSummary) > 0 && (
```
Here, IMO, we should display it for every proposal that received votes.

victorgcramos (changes requested):
Looking good. Noticed that proposal's footer buttons are misaligned:

Screen Shot 2022-07-25 at 3 18 48 PM

What do you think of adding a tooltip to it (using the `tooltipText` prop) to indicate that this button means "search for proposals votes" or something like that?

Although the old code only used to display the "search votes" button for active voting proposals, I think we should display the button for also approved and rejected proposals. _(see my inline comment)_.

Screen Shot 2022-07-25 at 3 18 48 PM

Other issues (#2828 #2829 #2830 #2831) have been fixed by this PR. Good job!

victorgcramos (approved):
tACK!

PR #2838, opened 2022-08-03
feat(politeiagui-scripts): add jest/babel configuration

tiagoalvesdulce:
This PR adds babel configuration for jest test env to `politeiagui-scripts`. This way, `babel.config` and `jest.config` are no longer needed on plugins. If the plugin has a `babel.config` or `jest.config` the default configs will be ignored and the plugin file will be the one used by `politeiagui-scripts`.


plugins-structure/packages/cli/package.json
```diff
@@ -21,8 +21,7 @@
     "prettify:write": "politeiagui-scripts prettify --write",
     "prettify:check": "politeiagui-scripts prettify --check",
     "lint:test": "politeiagui-scripts lint",
-    "lint:fix": "politeiagui-scripts lint --fix",
-    "test": "politeiagui-scripts test"
```
victorgcramos, line 25: Any reasons for removing that?

victorgcramos (commented):
I'm not sure if I'm doing the correct setup, but when I run the `politeiagui-scripts test` command on `packages/core` path, I get the following error:
```bash
SyntaxError: Cannot use import statement outside a module
```

In order to avoid that, I think it would be nice if you added some examples for that command on politeiagui-scripts readme. What do you think?

plugins-structure/packages/cli/package.json
tiagoalvesdulce: `lint:fix` was a mistake. Sorry. `test` I'm removing because there are no test files on `cli` and if you want to run tests for all packages jest complains about that.

tiagoalvesdulce:
> I'm not sure if I'm doing the correct setup, but when I run the politeiagui-scripts test command on packages/core path, I get the following error

Yeah, you have to run a fresh yarn install inside `politeiagui-scripts`.

plugins-structure/packages/cli/package.json
victorgcramos: that makes sense. my bad. Thanks!

victorgcramos:
@tiagoalvesdulce after a fresh install, it works as expected. Thanks!

victorgcramos (approved):
LGTM!

PR #2854, opened 2022-08-23
multi: plugins hooks and files cleanup

### LINKED ISSUE #2774: [plugins-architecture] Investigate ways to simplify plugins hooks
We have some very good and simple hooks but I feel some of them are bloated. IMO the `useFetch` hooks should:

1. Trigger a fetch if the status of that request is idle.
2. Return response, new status and error.

victorgcramos:
Closes #2774 

This PR removed unused stuff from plugins. It organizes the architecture to
use the new services + listeners approach, instead of using them on the
react way.


tiagoalvesdulce (approved):
tACK. Cool cleanup!

tiagoalvesdulce (approved):

PR #2866, opened 2022-10-18
feat: navigation improvements

victorgcramos:
This PR adds a new navigation service on core, as well as the structure
for navigating between pages using the `GoBackLink` component.

- [x] Add core navigation global service
- [x] dynamically update page titles according to proposal's name
- [x] add "Go Back" link and fix legacy errors
- [x] e2e tests for page title changes

### Preview
2866-navigation-improvements


tiagoalvesdulce (changes requested):
Initially I thought everything was too verbose but @victorgcramos pointed out the need to have a navigation history on Matrix because we want some custom navigation different from `history.back()`. See https://github.com/decred/politeiagui/issues/2655.

We agreed that there's no need for a service to update the page title since it's a fairly simple task and this was indeed verbose.

tiagoalvesdulce (approved):
tACK

PR #2867, opened 2022-10-19
feat(multi): better handle politeiawww error + fix client and core alias paths

tiagoalvesdulce:
- Handle 504 error when politeiawww is not running.
- Fix `politeiagui/core` and `politeiagui/core/client` paths on `politeiagui-scripts`.
- Use modern `?.` operator on validation files.


victorgcramos (approved):
LGTM

PR #2868, opened 2022-10-24
build: improve code-splitting and chunks cache

tiagoalvesdulce:
Depends on https://github.com/decred/pi-ui/pull/466

- Lazy load routes (this separate routes in different chunks)
- Split pi-ui from vendors chunk (since pi-ui is a UI lib and can change frequently so we cache it separately)
- Cache vendors and pi-ui
- Gzip

Bundle size before (1.5Mb):

Screen Shot 2022-10-22 at 17 49 32

Bundle size after (482Kb gziped):

Screen Shot 2022-10-24 at 14 11 36


victorgcramos:
Nice!

victorgcramos (approved):
LGTM! Good job!

PR #2870, opened 2022-10-27
fix(app/pi): app startup error + lint fixes

victorgcramos:
This PR fixes the following error caused by the absence of the 
`#root` element when app routes are initialized:

image


tiagoalvesdulce (approved):
ACK

PR #2872, opened 2022-11-01
fix(multi): core API error fixes + E2E tests fixes

victorgcramos:
This PR fixes some errors on politeia app e2e tests. While fixing those
errors, I noticed that some API errors were being handled incorrectly,
due to an incorrect variable usage, which were causing tests to fail.

Before:
Screen Shot 2022-11-01 at 8 42 32 PM

After:
Screen Shot 2022-11-01 at 8 46 33 PM

I also added some e2e tests for app startup errors, which asserts that
`/api` errors are displayed correctly. Once server returns an error, we
should also display the error codes. However, if no response body is
returned, we should display the generic `Cannot fetch "${url}". Is politeiawww running?`
message.

Screen Shot 2022-11-01 at 8 51 48 PM
Screen Shot 2022-11-01 at 8 52 09 PM

### DIFF
--- a/plugins-structure/apps/politeia/cypress/e2e/app.cy.js
+++ b/plugins-structure/apps/politeia/cypress/e2e/app.cy.js
@@ -1,0 +1,23 @@
+function errorMock() {
+  return {
+    errorcode: 1658261424,
+  };
+}
+describe("Given failed app startup", () => {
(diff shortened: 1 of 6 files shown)


tiagoalvesdulce (approved):
LGTM

PR #2889, opened 2023-01-18
feat: Add Form validation

victorgcramos:
This PR adds validation to all forms on politeia, including error messages,
and form submission control.

- Login Form validation
- Signup form validation
- User key update form validation
- user password change/reset form validation
- username change form validation
- email verification form validation


tiagoalvesdulce (approved):
Code looks good.

Password min length validation is missing in the login page.

victorgcramos:
@tiagoalvesdulce This will be added once the user layer gets implemented. Sounds good? We need that to get the password min and max lengths, so I can replace `MIN_PASSWORD_LENGTH` with the policy values as well

Issue #2875, opened 2022-11-11
[plugin-architecture] Pages


tiagoalvesdulce:
Missing pages: 
- User
    - [ ] Log in
    - [ ] Sign up
    - [ ] Verify Key
    - [ ] Verify Email
    - [ ] Details
    - [ ] Privacy Policy
    - [ ] Request Resend Verification Email
    - [ ] Request Reset Password
    - [ ] Reset Password
    - [ ] Search
- Admin
    - [ ] List
- Proposals
    - [ ] Edit
    - [ ] Draft
    - [ ] Raw

User and Admin pages depend on the backend user plugin to be completely functional but we can start moving the markup from the old codebase.

victorgcramos:
Working on User Proposals.

victorgcramos:
Working on remaining User Details tabs: 
- #2884

victorgcramos:
User Log in, Sign up, Verify Key, Verify Email, Privacy Policy, Request Resend Verification Email, Request Reset Password and Reset Password are coming up with:
- #2887

victorgcramos:
Admin user search and unvetted proposals list are coming up with:
- #2891

victorgcramos:
Proposals Edit and New are coming up with:
- #2892

PR #2892, opened 2023-01-23
feat: Add Proposal Edit and Raw pages

victorgcramos:
This PR adds the Proposal Edit and Proposal Raw pages.

Part of #2875 

While implementing this, I noticed that the logic used to fetch proposal
details, on Details (`record/:token`) page,  was being affected by the
proposal details fetch on Edit (`record/:token/edit`) page. Since we have
already fetched it on another page, other listeners such as voteSummaries,
comments, piSummaries, weren't triggered.

One way to quick fix that was to force a request on every page load on
both details and edit pages. This, of course isn't a correct design in terms
of cache and navigation delays, but since it's out of this PR scope, I 
decided to go for this quick fix.

- Proposal Edit page: `/record/:token/edit`,
- Proposal Raw page: `/record/:token/raw`.


tiagoalvesdulce (changes requested):
Raw page shouldn't be stylized.

How it is:
Screen Shot 2023-01-26 at 17 06 18

How it should be (matching production):
Screen Shot 2023-01-26 at 17 06 35
:

tiagoalvesdulce (approved):
LGTM

